Issue #295, opened 2015-03-15
Revisit Resolvers


GoogleCodeExporter:
```
For v2, I think we should reconsider which resolvers we include and the 
behavior of the LenientResolver.

Personally, I can't think of a use case for ReturnStartOfIntervalAfter or 
ReturnEndOfIntervalBefore.  These do more than compensate for the gap, as they 
shift the time by an amount other than the DST saving.

For example, if you had a job scheduling system with a recurrence pattern like 
"Daily at 2:30 AM Pacific Time", and you used the ReturnStartOfIntervalAfter 
resolver, the job would run at 3:00 AM on the day of the spring-forward 
transition.  One would commonly expect the job to run at 3:30 AM - the original 
time plus the DST shifted amount.  We need to include a resolver that has this 
behavior.  Perhaps "ReturnForwardShifted" or something like that.

The LenientResolver combines ReturnStartOfIntervalAfter and ReturnLater. 
Thinking of the same use case, it would be strange if a "Daily at 1:00 AM" job 
didn't fire at 1:00 on the day of the fall-back transition just because it was 
waiting for the second instance of 1:00 AM.  We should consider changing the 
composition of the LenientResolver, to ReturnForwardShifter and ReternEarlier.  
This would be a breaking change, but that's something that we could document as 
a v2 feature.
```

Original issue reported on code.google.com by `mj1856` on 25 Jun 2014 at 8:04

PR #365, opened 2015-05-05
Revisit Resolvers per #295

mj1856:
Discussed in #295, this PR does the following:

- Adds a `ReturnForwardShifted` resolver

- Changes the behavior of the `LenientResolver` and associated methods:
  - Ambiguous values are now resolved with `ReturnEarlier` instead of `ReturnLater`
  - Skipped values are now resolved with `ReturnForwardShifted` instead of `ReturnStartOfInterval`

- Updates associated xml docs and unit tests.

Note that this is a breaking change, which should be described and highlighted in the documentation and release notes when 2.0 is released.  I'm convinced that this is a better default behavior than the existing lenient behavior, based on numerous real-world examples.

My thinking goes like this:
- If you have an ambiguous value, it's probably because you are scheduling a job to run at the same time every day.  In the vast majority of cases, it would be considered a bug if a job did not run at the first time the local time was encountered on the clock.

- If you have a skipped value, you probably have it due to either some external clock source not advancing for DST, or because you are scheduling a job to run at the same time every day, and that time falls into the gap.  Thus, the value should shift forward, applying the DST gap to the value.  The current behavior of advancing to the first instant after the gap doesn't really make sense, because a clock would never tick forward in that manner.

The only exception case that comes to mind is when scheduling the close of business when the business elects to stay open the extra hour of the DST overlap.  For example, a bar might have to stop serving alcohol at 1:00AM, but once a year they get the option to serve until the *second* 1:00AM.  I think this should be thought of as an edge case rather than the default, and can thus be handled explicitly with the `ReturnLater` resolver.

### DIFF
--- a/src/NodaTime/TimeZones/Resolvers.cs
+++ b/src/NodaTime/TimeZones/Resolvers.cs
@@ -67,6 +67,19 @@ namespace NodaTime.TimeZones
         };
 
         /// <summary>
+        /// A <see cref="SkippedTimeResolver"/> which shifts values in the "gap" forward by the duration
+        /// of the gap (which is usually 1 hour).
+        /// </summary>
+        public static readonly SkippedTimeResolver ReturnForwardShifted = (local, zone, before, after) =>
+        {
+            Preconditions.CheckNotNull(zone, nameof(zone));
+            Preconditions.CheckNotNull(before, nameof(before));
+            Preconditions.CheckNotNull(after, nameof(after));
+            long gap = after.WallOffset.Ticks - before.WallOffset.Ticks;
+            return new ZonedDateTime(local.PlusTicks(gap), zone, after.WallOffset);
+        };
+
+        /// <summary>
         /// A <see cref="SkippedTimeResolver"/> which simply throws a <see cref="SkippedTimeException"/>.
         /// </summary>
         public static readonly SkippedTimeResolver ThrowWhenSkipped = (local, zone, before, after) =>
--- a/src/NodaTime/TimeZones/Resolvers.cs
+++ b/src/NodaTime/TimeZones/Resolvers.cs
@@ -68,7 +68,8 @@ namespace NodaTime.TimeZones
 
         /// <summary>
         /// A <see cref="SkippedTimeResolver"/> which shifts values in the "gap" forward by the duration
-        /// of the gap (which is usually 1 hour).
+        /// of the gap (which is usually 1 hour).  This corresponds to the instant that would have occured,
+        /// had there not been a transition.
         /// </summary>
         public static SkippedTimeResolver ReturnForwardShifted { get; } = (local, zone, before, after) =>
         {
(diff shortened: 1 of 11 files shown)


malcolmr:
fwiw, this looks fairly reasonable to me. Let's see what Jon thinks.

mj1856:
Also worth noting, I just discovered that this is the default behavior in java.time in Java 8.  At least, that's how it's described in the javadocs for [`LocalDateTime.atZone`](http://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#atZone-java.time.ZoneId-)

> In most cases, there is only one valid offset for a local date-time. In the case of an overlap, where clocks are set back, there are two valid offsets. This method uses the earlier offset typically corresponding to "summer".

> In the case of a gap, where clocks jump forward, there is no valid offset. Instead, the local date-time is adjusted to be later by the length of the gap. For a typical one hour daylight savings change, the local date-time will be moved one hour later into the offset typically corresponding to "summer".

jskeet:
Sorry, for some reason wasn't watching the repo. Will try to take a look tonight.

mj1856:
(rebased to fix a typo)

src/NodaTime/TimeZones/Resolvers.cs
```diff
@@ -67,6 +67,19 @@ public static class Resolvers
         };
 
         /// <summary>
+        /// A <see cref="SkippedTimeResolver"/> which shifts values in the "gap" forward by the duration
+        /// of the gap (which is usually 1 hour).
+        /// </summary>
+        public static readonly SkippedTimeResolver ReturnForwardShifted = (local, zone, before, after) =>
```
jskeet: Could you make this a read-only property (like ThrowWhenAmbiguous) and fix up the other fields while you're in the file? I'm trying to get rid of public fields from the API almost everywhere.

src/NodaTime/TimeZones/Resolvers.cs
```diff
@@ -67,6 +67,19 @@ public static class Resolvers
         };
 
         /// <summary>
+        /// A <see cref="SkippedTimeResolver"/> which shifts values in the "gap" forward by the duration
+        /// of the gap (which is usually 1 hour).
```
jskeet: Perhaps add a "remarks" that this means it returns a value corresponding to the instant that *would* have occurred using the earlier zone interval, if there hadn't been a transition? That might lead to a simpler implementation, too - not sure. (It could be that I'm mistaken about the effect...)

jskeet:
Basically fine - see a couple of comments. Could you add something into the version history and the 1->2 migration documentation? (Or create an issue for me to do so...)

jskeet:
(Whoops - close was accidental.)

mj1856:
Ok, how's it look now?

jskeet:
Thanks Matt - merged, and will pull to my local clone right now before I forget :)

PR #381, opened 2015-09-13
Implement IEquatable for DateInterval.

### LINKED ISSUE #380: Make DateInterval IEquatable
This is a feature request to make DateInterval implement IEquatable

malcolmr:
Also implement `GetHashCode()` and a basic `ToString()`.

Fixes #380.

### DIFF
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -60,6 +60,117 @@ namespace NodaTime.Test
         }
 
         [Test]
+        public void Equals_EqualValues()
+        {
+            LocalDate start = new LocalDate(2000, 1, 1);
+            LocalDate end = new LocalDate(2001, 6, 19);
+            var interval1 = new DateInterval(start, end, false);
+            var interval2 = new DateInterval(start, end, false);
+            Assert.AreEqual(interval1, interval2);
+            Assert.AreEqual(interval1.GetHashCode(), interval2.GetHashCode());
+            Assert.IsTrue(interval1 == interval2);
+            Assert.IsFalse(interval1 != interval2);
+            Assert.IsTrue(interval1.Equals(interval2)); // IEquatable implementation
+        }
+
+        [Test]
+        public void Equals_DifferentStart()
+        {
+            LocalDate start1 = new LocalDate(2000, 1, 1);
+            LocalDate start2 = new LocalDate(2000, 1, 2);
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -60,6 +60,117 @@ public void Construction_Properties()
...
+        [Test]
+        public void Equals_DifferentStart()
+        {
+            LocalDate start1 = new LocalDate(2000, 1, 1);
+            LocalDate start2 = new LocalDate(2000, 1, 2);
+            LocalDate end = new LocalDate(2001, 6, 19);
+            var interval1 = new DateInterval(start1, end, false);
+            var interval2 = new DateInterval(start2, end, false);
+            Assert.AreNotEqual(interval1, interval2);
+            Assert.AreNotEqual(interval1.GetHashCode(), interval2.GetHashCode());
```
malcolmr: This doesn't seem valid, by the way: there's generally no guarantee that two hash codes will be non-equal in the event that the instances are non-equal. However, we seem to do this elsewhere as well, so perhaps I'm missing something?

src/NodaTime/DateInterval.cs
```diff
@@ -87,6 +92,73 @@ public DateInterval(LocalDate start, LocalDate end)
...
+        /// <param name="lhs">The first value to compare</param>
+        /// <param name="rhs">The second value to compare</param>
+        /// <returns>True if the two date intervals have the same properties; false otherwise.</returns>
+        public static bool operator ==(DateInterval lhs, DateInterval rhs)
+        {
+            if (ReferenceEquals(lhs, rhs))
+            {
+                return true;
+            }
+            if (ReferenceEquals(lhs, null) || ReferenceEquals(rhs, null))
```
malcolmr: I assume that this is correct, but I note that we don't implement `operator==` for any other classes, so I don't have anything to check against.

src/NodaTime.Test/DateIntervalTest.cs
jskeet: There's no guarantee, but if the test fails, the chances of it being due to a bug are extremely high.

src/NodaTime/DateInterval.cs
jskeet: Yup, that's fine. An alternative approach is to make `==` call `Equals`, leading to just one nullity check. But this feels nicer.

jskeet:
Right, will merge this in and then address my own comments :)

PR #389, opened 2015-10-12
Test more calculators in YearMonthCalculatorTest... 

jskeet:
… and fix them when they break.

HebrewYearMonthCalculator was broken for year 0 and earlier - which isn't a problem for
most code, but it's possible that in a few cases our internals need to be able to handle values
outside the normal range. The cache format was assuming it could tweak the two high bits
arbitrarily - which doesn't work so well when the original value is negative. We now use the
bottom two bits for the lengths of Heshvan/Kislev.

### DIFF
--- a/src/NodaTime.Test/Calendars/YearMonthDayCalculatorTest.cs
+++ b/src/NodaTime.Test/Calendars/YearMonthDayCalculatorTest.cs
@@ -13,37 +13,52 @@ namespace NodaTime.Test.Calendars
     [TestFixture]
     public class YearMonthDayCalculatorTest
     {
+        // Here the term "Islamic" only refers to whether the implementation is IslamicYearMonthDayCalculator,
+        // not whether the calendar itself is based on Islamic scripture.
+        // TODO(code review): Make sure I've reported a bug for the CodeRush Roslyn preview test runner - it
+        // seems to only have one test per *type* of calculator. (So only one IslamicCalculator per test...)
         private static readonly TestCaseData[] NonIslamicCalculators = {
             new TestCaseData(new GregorianYearMonthDayCalculator()).SetName("Gregorian"),
             new TestCaseData(new CopticYearMonthDayCalculator()).SetName("Coptic"),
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/Calendars/YearMonthDayCalculatorTest.cs
```diff
@@ -13,37 +13,52 @@ namespace NodaTime.Test.Calendars
     [TestFixture]
     public class YearMonthDayCalculatorTest
     {
+        // Here the term "Islamic" only refers to whether the implementation is IslamicYearMonthDayCalculator,
+        // not whether the calendar itself is based on Islamic scripture.
+        // TODO(code review): Make sure I've reported a bug for the CodeRush Roslyn preview test runner - it
```
malcolmr: Well, have you? :-)

src/NodaTime/Calendars/HebrewScripturalCalculator.cs
```diff
@@ -211,7 +213,7 @@ private static bool IsKislevShort(int year)
         internal static int ElapsedDays(int year)
         {
             int cache = GetOrPopulateCache(year);
-            return cache & ElapsedDaysCacheMask;
+            return cache >> 2;
```
malcolmr: Not sure if it's worth pulling out that `2` into an `ElapsedDaysShift` or similar?

malcolmr:
LGTM.

src/NodaTime/Calendars/HebrewScripturalCalculator.cs
jskeet: Wondered about it, but had thought it was a bit trivial. Given the comment, happy to do so :)

src/NodaTime.Test/Calendars/YearMonthDayCalculatorTest.cs
jskeet: Not yet. Going to write up a small test example. Now that I have an internet connection, I can email myself to keep it in my inbox until I've done it though - will remove the comment :)

Issue #317, opened 2015-03-15
Consider MonthDay and YearMonth


GoogleCodeExporter:
```
java.time has types MonthDay and YearMonth.  Consider if we want these for Noda 
Time vNext.
```

Original issue reported on code.google.com by `mj1856` on 4 Aug 2014 at 5:44

jskeet:
I've been approached about these types again, and I think it would be worth looking at them more closely. Will create a new branch to see what they could look like, in terms of operations.

idg10:
I'd question whether MonthDay is the right name. It feels inconsistent with other names in NodaTime. (E.g., this feels akin to renaming LocalTime HoursMinutesSecondsMilliseconds.) It's just a description of what's inside, whereas most type names give a clue as to what they're for. So I'm wondering if RecurringDate, or AnnualDate, or even Anniversary might be more informative names.

jskeet:
Starting to play with this a little... MonthDay feels useful (and that probably *is* what I'd name it - Anniversary feels like what someone might use it for, rather than what it represents; it's also consistent with Joda Time and Java 8).

YearMonth feels harder to pin down in terms of a use case. There's the argument that it's appropriate for (say) a credit card expiry date, but it feels like in that case it would be cleaner to represent it as a LocalDate anyway, to avoid the confusion about whether it expires at the end of the month or the start. (There's still the question about whether it expires at the end of the day or the start, admittedly.) It's easy enough to write the code - just harder to really justify its presence.

MenoData:
There has been an interesting debate about this stuff among Java folks (Oracle and Threeten-team).

https://github.com/ThreeTen/threeten/issues/209

Naming `MonthDay` as `Anniversary` seems to be a good idea IMHO. About the other types like `YearMonth`, `YearQuarter` etc., I would mainly model such classes as date range intervals and source for iterating over full calendar dates (my plan for my own Java library Time4J in the future).

jskeet:
@MenoData Thanks for that.

Anniversary feels like an odd name for a type used to store (say) the start of the UK tax year (April 6th). That's not an anniversary in that it's not remembering a particular event in *one* year, but several years later. MonthDay could certainly be used *for* anniversaries, but it's not uniquely for that. I think "MonthDay" gives a clearer indication of what it *is*, and then it can be used for whatever people want. But I'm still happy to be persuaded.

[...]

PR #394, opened 2015-10-15
First pass at an ISO-only annual date type.

jskeet:
(Looks a bit simple at the moment... not sure what else is needed.)

### DIFF
--- a/src/NodaTime.Test/AnnualDateTest.cs
+++ b/src/NodaTime.Test/AnnualDateTest.cs
@@ -1,0 +1,64 @@
+﻿// Copyright 2015 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using NUnit.Framework;
+using System;
+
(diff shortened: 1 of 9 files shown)


malcolmr:
I assume this is the MonthDay suggestion from #317?

Any reason not to go for `MonthDay` as the type name, especially given that's what `java.time` uses?

jskeet:
Reason for not using MonthDay was user feedback from idg10 and MenoData. Personally I'd still like MonthDay, but I'm okay with AnnualDate too... although I wonder whether we could have MonthDay implement an IAnnualDate interface, where other implementions could be for things like "Easter".

I'd be happy to work out the functionality and get it submitted with *any* name, and then revisit before release.

jskeet:
I *think* I've fixed everything other than the name and the text parsing, both of which need further discussion. I'll squash the commits before merging, after they've been reviewed.

malcolmr:
LGTM as-is. Let's discuss the name separately.

Parsing we can always add later, so I don't think is a blocker for anything.

Issue #399, opened 2015-11-01
LocalDate.InZone missing?


jskeet:
I could have sworn there was a way of finding out when a particular date started in a time zone, but I can't find it now. Hopefully I'm just failing to find the relevant documentation, but if not, we should at it. (Quite possibly as LocalDate.InZone and DateTimeZone.AtStartOfDay.)

malcolmr:
How about `localDate.AtMidnight().InZoneLeniently(zone)`? Or did you want something else?

malcolmr:
Actually I'm not sure that works in 2.0 due to the way that `InZoneLeniently()` now works; you'd need `InZone(zone, resolver)` using a resolver with the old 1.x behaviour.

mj1856:
We already have a [`DateTimeZone.AtStartOfDay`](http://nodatime.org/1.3.x/api/html/M_NodaTime_DateTimeZone_AtStartOfDay.htm) instance method.

```csharp
LocalDate ld = new LocalDate(2015, 10, 18);
DateTimeZone tz = DateTimeZoneProviders.Tzdb["America/Sao_Paulo"];
ZonedDateTime zdt = tz.AtStartOfDay(ld); // "2015-10-18T01:00:00 America/Sao_Paulo (-02)"
```

It might be nice to invert it and put it on `LocalDate` also.  I don't think resolvers are useful here, as we're not starting with any local time.

jskeet:
Thanks for finding it... I think I was looking in ZonedDateTime for no
obvious reason. But yes, we should put it in LocalDate too.

PR #400, opened 2015-11-04
Implement LocalDate.InZone.

jskeet:
This is a convenience method to call DateTimeZone.AtStartOfDay.

This fixes issue #399.

(I'd be open to other naming suggestions, btw.)

// cc @malcolmr

### DIFF
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -581,6 +581,24 @@ namespace NodaTime
         public bool Equals(LocalDate other) => this == other;
 
         /// <summary>
+        /// Resolves this local date into a <see cref="ZonedDateTime"/> in the given time zone representing the
+        /// start of this date in the given zone.
+        /// </summary>
+        /// <remarks>
+        /// This is a convenience method for calling <see cref="DateTimeZone.AtStartOfDay(LocalDate)"/>.
+        /// </remarks>
+        /// <param name="zone">The time zone to map this local date into</param>
+        /// <exception cref="SkippedTimeException">The entire day was skipped due to a very large time zone transition.
+        /// (This is extremely rare.)</exception>
+        /// <returns>The <see cref="ZonedDateTime"/> representing the earliest time on this date, in the given time zone.</returns>
+        [Pure]
+        public ZonedDateTime AtStartOfDayInZone([NotNull] DateTimeZone zone)
+        {
+            Preconditions.CheckNotNull(zone, nameof(zone));
+            return zone.AtStartOfDay(this);
+        }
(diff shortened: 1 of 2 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -581,6 +581,24 @@ int IComparable.CompareTo(object obj)
...
+        /// </summary>
+        /// <remarks>
+        /// This is a convenience method for calling <see cref="DateTimeZone.AtStartOfDay(LocalDate)"/>.
+        /// </remarks>
+        /// <param name="zone">The time zone to map this local date into</param>
+        /// <exception cref="SkippedTimeException">The entire day was skipped due to a very large time zone transition.
+        /// (This is extremely rare.)</exception>
+        /// <returns>The <see cref="ZonedDateTime"/> representing the earliest time on this date, in the given time zone.</returns>
+        [Pure]
+        public ZonedDateTime InZone([NotNull] DateTimeZone zone)
```
malcolmr: I'm not convinced that `InZone` is a great name for a `LocalDate` ↦ `ZonedDateTime` function that only returns a small portion of the latter domain.

How about `GetStartOfDay(DateTimeZone)`? (though I'm not too keen on "Get".)

jskeet: I'd like something with "zone" in the name - that's what I think I'd be looking for if I were hunting for methods.
StartOfDayInZone? InZoneStartOfDay? InZoneAtStartOfDay? All somewhat clumsy...

malcolmr: Since we already have `zone.AtStartOfDay(localDate)`, I'm actually thinking that `localDate.AtStartOfDay(zone)` is fairly sensible.

(No, it doesn't have 'zone' in the name.)

The only problem is that `LocalDate.AtMidnight()` exists, returning a `LocalDateTime`, and it's not obvious how this differs.

mj1856: I like either `AtStartOfDay` or `AtStartOfDayInZone`.  The later more explicitly clarifies the intent, addressing Malcolm's point about differing it from `AtMidnight`.

I don't like `InZone`, as it's not clear at all what time will be chosen.  At least with `AtStartOfDay`, you still have to pass a time zone as a parameter.

mj1856:
Other than the name, LGTM

jskeet:
Righto. Will go with AtStartOfDayInZone, favouring clarity over brevity. Will rename and merge. (Need to investigate the CI problems at some point - looks like it's nothing to do with us, but maybe we can change which Mono version we use...)

PR #411, opened 2015-12-03
Handle completely empty fractional seconds with "F" format specifier.

scottstephens:
I would have expected the LocalTime pattern "HHmmssFF" to handle either of these two cases:
(1) "11223344" -> 11:22:33.44
(2) "112233" -> 11:22:33.00

Case 1 works, but case 2 does not.

Some similar cases do work:
"1122334" / "HHmmssFF" -> 11:22:33.4
"112233" / "HHmmss.FF" -> 11:22:33.0

The fix turns out to be pretty easy, as you can see by the diff. The nature of the change (removing an explicit error condition check) made me think that this would cause problems, but all the unit tests still pass, and I haven't been able to think of any cases where I would actually want the old behavior.

If somebody else can think of some cases where the check actually gives good behavior, we should at a minimum add some tests to demonstrate it. If not, you can merge in my change!

### DIFF
--- a/src/NodaTime/Text/ValueCursor.cs
+++ b/src/NodaTime/Text/ValueCursor.cs
@@ -291,7 +291,7 @@ namespace NodaTime.Text
         /// <param name="allRequired">If true, <paramref name="maximumDigits"/> digits must be present in the
         /// input sequence. If false, there must be just at least one digit.</param>
         /// <returns><c>true</c> if the digits were parsed.</returns>
-        internal bool ParseFraction(int maximumDigits, int scale, out int result, bool allRequired)
+        internal bool ParseFraction(int maximumDigits, int scale, out int result, int minimumDigits)
         {
             unchecked
             {
@@ -302,16 +302,13 @@ namespace NodaTime.Text
 
                 result = 0;
                 int localIndex = Index;
-                int maxIndex = localIndex + maximumDigits;
-                if (maxIndex > Length)
+                int minIndex = localIndex + minimumDigits;
+                if (minIndex > Length)
                 {
                     // If we don't have all the digits we're meant to have, we can't possibly succeed.
-                    if (allRequired)
-                    {
-                        return false;
-                    }
-                    maxIndex = Length;
+                    return false;
                 }
+                int maxIndex = Math.Min(localIndex + maximumDigits, Length);
                 for (; localIndex < maxIndex; localIndex++)
                 {
                     // Optimized digit handling: rather than checking for the range, returning -1
@@ -324,19 +321,14 @@ namespace NodaTime.Text
                     result = result * 10 + digit;
                 }
                 int count = localIndex - Index;
-                // Couldn't parse any digits?
-                if (count == 0)
+                // Couldn't parse the minimum number of digits required?
+                if (count  < minimumDigits)
                 {
                     return false;
                 }
                 result = (int) (result * Math.Pow(10.0, scale - count));
-                bool ret = !allRequired || localIndex == maxIndex;
-                // Only move the cursor on success.
-                if (ret)
-                {
-                    Move(localIndex);
-                }
-                return ret;
+                Move(localIndex);
+                return true;
             }
         }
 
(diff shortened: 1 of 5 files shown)


jskeet:
To be clear, were you really expecting the pattern `HHmmssFF` to work, or `HH:mm:ss.FF`? I would only fully expect it to work with the dot between the `s` and the `F`. The following works fine for me (in 2.0):

```
using System;
using NodaTime;
using NodaTime.Text;

class Program
{
    static void Main(string[] args)
    {
        var pattern = LocalTimePattern.CreateWithInvariantCulture("HH:mm:ss.FF");
        var text = "11:22:33";
        Console.WriteLine(pattern.Parse(text).Value);
    }
}
```

I'll have a look at the change and consider it, but the idea of having a variable-width fractional part without a decimal point sounds like a bad idea to start with, IMO. (i.e. I'd very strongly urge anyone to change the pattern anyway.)

jskeet:
Possible example of a pattern/value this would allow to be parsed which I don't think we'd want - a more "normal" pattern of `HH:mm:ss.FFF`, with a value of `13:00:00.` (note the trailing decimal point). Personally I would want that to fail; I don't view a trailing decimal point with nothing after it as "good input".

But aside from anything else, if we agree that that's an example of "bad input" I ought to add a test case... we should either adopt your change or have a test case that fails when it's made, so thanks massively for putting the effort into finding this to start with :)

scottstephens:
I really was expecting pattern `HHmmssFF` to work on both inputs like "11223344" and "112233".  That said, I definitely agree it's a bad pattern choice. Unfortunately I'm stuck with it; it's in a file format created by somebody else.

Yesterday I was okay with "11:22:33." as valid input, but on second thought, I think I agree with you that it makes more sense for that to fail. I've pushed a revised version to this request that keeps the failure behavior in the period case, but works in the non-period case. I've squashed the commits for this request, but the non-squashed version is also here for reference: https://github.com/scottstephens/nodatime/tree/empty_fractional_seconds_history

Viewed in isolation (without considering the existence of the period as a possible pattern element) I think there are two obvious interpretations of `F` 
(1) a fractional second digit that is optional if it and all following digits are zero, or
(2) a fractional second digit that is not included in the text if it and all following digits are zero.
    With the current implementation (excluding cases where the first digit is zero) the formatting code is compatible with either definition, but the  parsing code is only compatible with definition 1: it doesn't fail if a trailing zero is present. I think this is sensible behavior, and I think getting rid of the exception to this rule makes the behavior more consistent.

It's only in the context of a period that allowing a second or later zero to be left off but not allowing the first zero to be left off makes sense.

jskeet:
I haven't had time to review the code yet, but hope to some time this week. The result sounds okay - if somewhat difficult to document clearly! Would you like to have a go at amending the documentation in this PR, or do you want me to fix that later? (I really don't mind much either way.)

Do you have a need for this fix urgently, btw? I'm just thinking that it's only currently part of 2.0, which I suspect still won't be out for a few months. I think I'd suggest some temporary string manipulation in your code to fix it until the next release, but if that's impractical we could look at alternatives.

scottstephens:
Personally I think the new behavior matches the existing documentation. If you'd prefer further clarification I think it would be best for you to write it.

There's no need to modify release schedules on my account. I solved my actual application's problem with a little string manipulation. I'm just a fan of the library and I like to contribute back to good projects when I find something.

jskeet:
Great - that all sounds good to me. Will try to review the change this week.

src/NodaTime/Text/ValueCursor.cs
```diff
@@ -302,16 +302,13 @@ internal bool ParseFraction(int maximumDigits, int scale, out int result, bool a
...
                 {
                     // If we don't have all the digits we're meant to have, we can't possibly succeed.
-                    if (allRequired)
-                    {
-                        return false;
-                    }
-                    maxIndex = Length;
+                    return false;
                 }
+                int maxIndex = Math.Min(localIndex + maximumDigits,Length);
```
jskeet: Nit: space after maximumDigits.

src/NodaTime/Text/ValueCursor.cs
```diff
@@ -325,18 +322,13 @@ internal bool ParseFraction(int maximumDigits, int scale, out int result, bool a
                 }
                 int count = localIndex - Index;
                 // Couldn't parse any digits?
```
jskeet: Nit: update comment.

jskeet:
Changes look good, thanks. If you could fix the couple of nits (and squash the commit), I'll merge then elaborate the docs just a little.

Thanks!

jskeet:
Lovely - thanks very much! Merging.

Issue #262, opened 2015-03-15
Implementation from Julian Date (JD)


GoogleCodeExporter:
```
Hello,
have you already thought to implement the Julian Date (JD)? Here I do not mean 
the Julian calendar. In the field of astronomy, this date is often used and 
made ​​possible by the use of a simpler handling of the date.

regards

Ralf La Porte
```

Original issue reported on code.google.com by `jplapo....@gmail.com` on 29 Dec 2013 at 12:17

jskeet:
Closing as this is now implemented.

PR #416, opened 2015-12-26
Added functions to convert Instant to and from Julian Day Numbers

GoodOlClint:
Julian Day Numbers are widely used in Astronomical calculations. I added two functions to convert the NodaTime.Instant classes back and forth to and from their JDN.

### DIFF
--- a/src/NodaTime/Instant.cs
+++ b/src/NodaTime/Instant.cs
@@ -502,6 +502,13 @@ namespace NodaTime
         #endregion
 
         /// <summary>
+        /// Constructs a <see cref="double"/> representing the same instant of time as this value.
+        /// </summary>
+        /// <returns>Returns a <see cref="double"/> containing the number of whole and fractional days since the Julian Epoch (Noon January 1st, 4713 BCE)</returns>
+        [Pure]
+        public double ToJulianDayNumber() => this.duration.TotalSeconds / 86400d + 2440587.5;
+
+        /// <summary>
         /// Constructs a <see cref="DateTime"/> from this Instant which has a <see cref="DateTime.Kind" />
         /// of <see cref="DateTimeKind.Utc"/> and represents the same instant of time as this value.
         /// </summary>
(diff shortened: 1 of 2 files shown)


mj1856:
I like the idea.  Needs unit tests. Also, why is one marked `[Pure]` while the other is not?

GoodOlClint:
Added unit tests.

Honestly, I marked `ToJulianDayNumber` as `[Pure]` because most of the other To functions were marked the same way. I didn't mark `FromJulianDayNumber` because I didn't see any From functions marked with it.

src/NodaTime/Instant.cs
```diff
@@ -502,6 +502,13 @@ public static Instant Max(Instant x, Instant y)
         #endregion
 
         /// <summary>
+        /// Constructs a <see cref="double"/> representing the same instant of time as this value.
+        /// </summary>
+        /// <returns>Returns a <see cref="double"/> containing the number of whole and fractional days since the Julian Epoch (Noon January 1st, 4713 BCE)</returns>
+        [Pure]
+        public double ToJulianDayNumber() => this.duration.TotalSeconds / 86400d + 2440587.5;
```
jskeet: Why not use TotalDays here?

src/NodaTime/Instant.cs
```diff
@@ -526,6 +533,13 @@ public static Instant Max(Instant x, Instant y)
             BclEpoch.PlusTicks(dateTimeOffset.Ticks - dateTimeOffset.Offset.Ticks);
 
         /// <summary>
+        /// Converts a Julian Day Number into a new Instant representing the same instant in time
+        /// </summary>
+        /// <param name="JDN">a <see cref="double"/> containing the number of whole and fractional days since the Julian Epoch (Noon January 1st, 4713 BCE)</param>
+        /// <returns>An <see cref="Instant"/> value representing the same instant in time as the given <see cref="double"/>.</returns>
+        public static Instant FromJulianDayNumber(double JDN) => UnixEpoch.PlusTicks(Convert.ToInt64((JDN - 2440587.5) * 86400) * TicksPerSecond);
```
jskeet: I don't think I'd use the Unix epoch here at all. I'd create a property of JulianEpoch representing noon on January 1st 4713 BCE (in UTC) and use that for both calculations.

jskeet:
Thanks for this. It's fine for the instance method to be marked Pure but not the static method - I believe that's how the R# annotations are expected to be used.

I have a few issues around the implementation, as noted in comments - would it be simpler if I merged and then changed those myself, leaving the unit tests as they are?

Note that this was an enhancement requested in #262 but without suitable detail provided by the requestor.

src/NodaTime/Instant.cs
GoodOlClint: That would work, I used TotalSeconds since I was referencing the UnixEpoch in the other function and wanted to keep them consistent.

src/NodaTime/Instant.cs
GoodOlClint: I originally did just than, but I didn't want to add more code than absolutely required so I ended up doing it this way.

GoodOlClint:
I made the suggested changes in a [new branch](https://github.com/GoodOlClint/nodatime/commits/JdnEpoch) I can merge those into my master if they look good

jskeet:
I think I'll still want to tinker a bit, e.g. with naming and using properties instead of public readonly fields. I'm happy to do that with a follow-up commit myself after merging your changes in, or I could suggest changes via code review. I suspect the former approach would be simpler if you're happy with that.

GoodOlClint:
Either way works, I'm still learning :)

jskeet:
Right, will merge and edit :)

jskeet:
See #418 for a few more changes.

PR #417, opened 2015-12-30
Fix Unix time conversions by using long arithmetic

jskeet:


jskeet:
Will self-merge when green. Noticed this bug while refactoring the Julian Date code.

malcolmr:
LGTM, nice.

PR #418, opened 2015-12-30
Refactor Julian Date arithmetic

jskeet:
This uses Julian Date rather than Julian Day Number as terminology, following
https://en.wikipedia.org/wiki/Julian_day - a Julian Day Number is an integer instead.

Annoying that we don't have a good way of constructing a duration from a double at the moment...

### DIFF
--- a/src/NodaTime.Test/InstantTest.cs
+++ b/src/NodaTime.Test/InstantTest.cs
@@ -17,19 +17,20 @@ namespace NodaTime.Test
         private static readonly Instant negativeFiftyMillion = Instant.FromUntrustedDuration(Duration.FromNanoseconds(-50000000L));
 
         [Test]
-        public void FromJDN()
+        // Gregorian calendar: 1957-10-04
+        [TestCase(2436116.31, 1957, 9, 21, 19, 26, 24, Description = "Sample from Astronomical Algorithms 2nd Edition, chapter 7")]
+        // Gregorian calendar: 2013-01-01
+        [TestCase(2456293.520833, 2012, 12, 19, 0, 30, 0, Description = "Sample from Wikipedia")]
+        [TestCase(1842713.0, 333, 1, 27, 12, 0, 0, Description = "Another sample from Astronomical Algorithms 2nd Edition, chapter 7")]
+        [TestCase(0.0, -4712, 1, 1, 12, 0, 0, Description = "Julian epoch")]
+        public void JulianDateConversions(double julianDate, int year, int month, int day, int hour, int minute, int second)
         {
-            Instant viaJDN = Instant.FromJulianDayNumber(2436116.31);
-            Instant expected = Instant.FromUtc(1957, 10, 4, 19, 26, 24);
-            Assert.AreEqual(viaJDN, expected);
-        }
-
-        [Test]
-        public void ToJDN()
-        {
-            Instant toJDN = new NodaTime.LocalDateTime(333, 1, 27, 12, 0, CalendarSystem.Julian).InUtc().ToInstant();
-            double expected = 1842713.0;
-            Assert.AreEqual(expected, toJDN.ToJulianDayNumber());
+            // When dealing with floating point binary data, if we're accurate to 50 milliseconds, that's fine...
+            // (0.000001 days = ~86ms, as a guide to the scale involved...)
+            Instant actual = Instant.FromJulianDate(julianDate);
+            Instant expected = new LocalDateTime(year, month, day, hour, minute, second, CalendarSystem.Julian).InUtc().ToInstant();
+            Assert.AreEqual(expected.ToUnixTimeMilliseconds(), actual.ToUnixTimeMilliseconds(), 50, "Expected {0}, was {1}", expected, actual);
+            Assert.AreEqual(julianDate, expected.ToJulianDate(), 0.000001);
         }
 
         [Test]
--- a/src/NodaTime.Test/NodaConstantsTest.cs
+++ b/src/NodaTime.Test/NodaConstantsTest.cs
@@ -1,0 +1,22 @@
+﻿// Copyright 2015 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using NUnit.Framework;
+
+namespace NodaTime.Test
+{
+    [TestFixture]
+    public class NodaConstantsTest
+    {
+        [Test]
+        public void JulianEpoch()
+        {
+            // Compute the Julian epoch using the Julian calendar, instead of the
+            // Gregorian version.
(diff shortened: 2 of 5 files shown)


src/NodaTime.Test/InstantTest.cs
```diff
@@ -16,20 +16,39 @@ public partial class InstantTest
         private static readonly Instant threeMillion = Instant.FromUntrustedDuration(Duration.FromNanoseconds(3000000L));
         private static readonly Instant negativeFiftyMillion = Instant.FromUntrustedDuration(Duration.FromNanoseconds(-50000000L));
 
+        // When dealing with floating point binary data, if we're accurate to 50 milliseconds, that's fine...
+        // (0.000001 days = ~86ms, as a guide to the scale involved...)
+
         [Test]
-        public void FromJDN()
+        public void FromJulianDate_OriginalTestData()
```
malcolmr: "Original" in what sense?

Perhaps just FromJulianDate() here and similarly below, if these are just test vectors we've come up with from scratch?

src/NodaTime.Test/NodaConstantsTest.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace NodaTime.Test
+{
+    [TestFixture]
+    public class NodaConstantsTest
+    {
+        [Test]
+        public void JulianEpoch()
+        {
+            // Compute the Julian epoch using the Julian calendar, instead of the
+            // Gregorian version
```
malcolmr: nit: .

src/NodaTime/Instant.cs
```diff
@@ -502,11 +502,12 @@ public static Instant Max(Instant x, Instant y)
         #endregion
 
         /// <summary>
-        /// Constructs a <see cref="double"/> representing the same instant of time as this value.
+        /// Returns the Julian Date of this instance - the number of days since
+        /// <see cref="NodaConstants.JulianEpoch"/> (noon on January 1st, 4713 BCE in the Julian calendar).
         /// </summary>
-        /// <returns>Returns a <see cref="double"/> containing the number of whole and fractional days since the Julian Epoch (Noon January 1st, 4713 BCE)</returns>
+        /// <returns>The number of days since the Julian Epoch.</returns>
```
malcolmr: Possibly "The (fractional) number of days" or similar?

src/NodaTime/Instant.cs
```diff
@@ -533,11 +534,20 @@ public static Instant Max(Instant x, Instant y)
             BclEpoch.PlusTicks(dateTimeOffset.Ticks - dateTimeOffset.Offset.Ticks);
 
         /// <summary>
-        /// Converts a Julian Day Number into a new Instant representing the same instant in time
+        /// Converts a Julian Date (JD) into a new <see cref="Instant"/> representing the given number of days
```
malcolmr: Arguably it's the Julian Date that represents the given number of days, not the Instant.

Perhaps "Converts a Julian Date represented as the given number of days since [...] into a new Instant"?

malcolmr: Not sure the (JD) adds anything useful here?

src/NodaTime/NodaConstants.cs
```diff
@@ -159,12 +159,18 @@ public static class NodaConstants
         /// <summary>
         /// The instant at the Unix epoch of midnight 1st January 1970 UTC.
         /// </summary>
-        public static readonly Instant UnixEpoch = Instant.FromUnixTimeTicks(0);
+        public static Instant UnixEpoch { get; } = Instant.FromUnixTimeTicks(0);
 
         /// <summary>
         /// The instant at the BCL epoch of midnight 1st January 0001 UTC.
```
malcolmr: I note that these two don't mention calendars; should they?

src/NodaTime/NodaConstants.cs
```diff
@@ -159,12 +159,18 @@ public static class NodaConstants
...
         /// </summary>
-        public static readonly Instant BclEpoch = Instant.FromUtc(1, 1, 1, 0, 0);
+        public static Instant BclEpoch { get; } = Instant.FromUtc(1, 1, 1, 0, 0);
+
+        /// <summary>
+        /// The epoch used for Julian dates. This is noon (UTC) on January 1st 4713 BCE in the proleptic
```
malcolmr: "The instant at the Julian epoch of ...", for consistency with {Unix,Bcl}Epoch, maybe?

malcolmr:
LGTM overall though.

src/NodaTime/NodaConstants.cs
jskeet: Done.

src/NodaTime/NodaConstants.cs
jskeet: I think it's reasonable to assume Gregorian unless otherwise mentioned.

src/NodaTime/Instant.cs
jskeet: That's the abbreviation used by Wikipedia, compared with JDN which is the integer Julian Date Number. Happy to remove it if it doesn't add anything though.

jskeet: Fixed the comment as per your other note too :)

src/NodaTime/Instant.cs
jskeet: Fractional sounds like it would be in the range (-1, 1) to me. Have changed to "The number of days (including fractional days)..."

src/NodaTime.Test/NodaConstantsTest.cs
jskeet: Done.

src/NodaTime.Test/InstantTest.cs
jskeet: I don't know where the original test data came from, unfortunately. @GoodOlClint, can you cast any light on this? I agree that it would be nice to rename the test appropriately. (Would be even nicer to parameterize, but it wouldn't be terribly clean in terms of parameters.)

jskeet:
Pushed changes so far. Will squash when we're happy to merge.

src/NodaTime.Test/InstantTest.cs
GoodOlClint: The original tests were from Chapter 7 of Jean Meeus' [Astronomical Algorithms 2nd Edition](http://www.willbell.com/math/mc1.htm). Example 7.a gives a date and calculates a JD and example 7.b gives a JD and calculates a date. Example 7.c uses the same dates and JDs but reverses them.

There shouldn't be a problem with changing one or both, but personally I would keep one test far enough in the past that the Julian Calendar differs from the Gregorian by a few days.

jskeet:
Right - have refactored the tests. It's slightly irritating that some were originally expressed in Julian dates and some in Gregorian, but it's at least better than before.

malcolmr:
All still LGTM,

jskeet:
Thanks. Squashed and force-pushed - will merge when green.

PR #420, opened 2016-01-05
"Partial patterns" for local date/time/date-time in ZonedDateTime and OffsetDateTime

### LINKED ISSUE #375: Support for culture sensitive date/time in ZoneDateTimePattern for formatting
http://stackoverflow.com/questions/31792508/with-nodatime-how-do-i-format-a-zoneddatetime-in-current-culture

I have a `ZonedDateTime` and I want to display it such that the datetime is formatted with the short date and short time configured on the workstation followed by the offset (something like... 05/01/2005 02:30 PM -05:00). I expected something like this would work...

```csharp
var patternDateTimeOffset = ZonedDateTimePattern.CreateWithCurrentCulture("g o<m>", DateTimeZoneProviders.Tzdb);
lblOriginalDateTimeAndOffsetVal.Text = patternDateTimeOffset.Format(zonedDateTime);
```

BUT, it appears that the "g" is not supported in `ZonedDateTimePattern` the way it is in `LocalDateTimePattern`. The code above throws a `NodaTime.Text.InvalidPatternException`.

I could replace the "g" with "MM/dd/yyyy hh:mm", but then it's not using the current culture.

I could use a `LocalDateTimePattern` for the datetime and then concatenate the offset using the `ZonedDateTimePattern`. This works, but seems ugly.

This seems like a pretty common thing. I'm new to NodaTime, so I'm certain I'm missing something. I'm using NodaTime 1.3.1 and targeting .net 4.0. Any help is appreciated.

jskeet:
Allow the local date, local time, or local date/time parts of ZonedDateTimePattern and OffsetDateTimePattern to be specified as whole patterns, so they can use standard patterns.

It would be logical to allow this within LocalDateTime as well, but it would rarely be as useful, and then we'd need to worry about nested patterns.

This fixes #375 (or at least works around it).

Documentation will be written when we're happy about the behaviour.


src/NodaTime.Test/Text/OffsetDateTimePatternTest.cs
```diff
@@ -78,6 +91,20 @@ public class OffsetDateTimePatternTest : PatternTestBase<OffsetDateTime>
...
             new Data(MsdnStandardExample) { Pattern = "r", Text = "2009-06-15T13:45:30.09+01 (ISO)", Culture = Cultures.FrFr },
 
+            // Custom embedded patterns (or mixture of custom and standard)
+            new Data(2015, 10, 24, 11, 55, 30, AthensOffset) { Pattern = "ld<yyyy*MM*dd>'X'lt<HH_mm_ss> o<g>", Text = "2015*10*24X11_55_30 +03" },
+            new Data(2015, 10, 24, 11, 55, 30, AthensOffset) { Pattern = "lt<HH_mm_ss>'Y'ld<yyyy*MM*dd> o<g>", Text = "11_55_30Y2015*10*24 +03" },
+            new Data(2015, 10, 24, 11, 55, 30, AthensOffset) { Pattern = "l<HH_mm_ss'Y'yyyy*MM*dd> o<g>", Text = "11_55_30Y2015*10*24 +03" },
+            new Data(2015, 10, 24, 11, 55, 30, AthensOffset) { Pattern = "ld<d>'X'lt<HH_mm_ss> o<g>", Text = "10/24/2015X11_55_30 +03" },
+            new Data(2015, 10, 24, 11, 55, 30, AthensOffset) { Pattern = "ld<yyyy*MM*dd>'X'lt<T> o<g>", Text = "2015*10*24X11:55:30 +03" },
+
+            // Standard embedded patterns. Short time versions have a second value of 0 so they can round-trip.
```
malcolmr: "seconds values", perhaps?

malcolmr:
per discussion:
- this seems sensible.
- support for `LocalDateTime` would probably be a good idea too, even though it implies support for e.g. `l<lt<g>>` as a `ZonedDateTime` pattern.

jskeet:
Okay, I think this is now ready for review. I'm not keen on so much code being in SteppedPatternBuilder, but at the moment I can't see anywhere better. Fortunately it's internal.

Docs still to come, of course.

jskeet:
Merging - still need to remember to write docs :)

PR #430, opened 2016-01-19
Use plain reflection instead of Roslyn for annotation tests

jskeet:
At some point we may want to revisit Roslyn, but it's not worth it right now - only one test really needed it.
NCrunch, testing under Mono, and testing under DNX all suffered due to this.


malcolmr:
Nice. Could you remove the Roslyn* test exclusion from the .travis.yml file as well? That was previously required, but should be no longer.

LGTM assuming it passes in that state.

jskeet:
Ah yes, good point. Will fix that, squash, force push, and merge when green. Ta!

PR #433, opened 2016-01-25
DNX support

jskeet:
This is still early work, but merging now in an attempt to get us moving forward.


jskeet:
(Not sure why Travis isn't running for this, but will merge and we'll see what happens...)

Issue #437, opened 2016-02-12
Get number of weeks in Week Year


maggiepint:
While coding yesterday, it would have been useful to me to be able to ask how many weeks are in the week year. As far as I can tell, there is no support for just asking NodaTime this, and no super simple way to get it. 
Since 12/31 can fall in week 52,53, or 1, that shortcut is out.

I coded around this, but it would be nice to have the functionality built in. I would make a pull request, but I struggle with finding a place in the API that I think it 'fits'. Maybe CalendarSystem.Iso - but then Iso's properties differ from the rest which is bad.

I can see the calculation in WeekYearCalculator. It would be great if it got exposed. Again, happy to do the work if advised on a place to put it.

jskeet:
Happy to look into it - it's likely to just mean a few methods proxying through the layers. We'd expose it in CalendarSystem, basically. (So you'd use `CalendarSystem.Iso.GetWeekYears(2015)` or whatever... something like that.

I'd initially only implement this in v2.0, which means you're unlikely to be able to use it for a while (unless you're using head from github, which is pretty stable, admittedly...). Let me know if that's an issue.

maggiepint:
No need for it right now. I can always manually compute it.

PR #438, opened 2016-02-12
Added CalendarSystem.GetWeeksInWeekYear

jskeet:
This suffers the problem we have elsewhere of not really having a min/max week-year... will file an issue.

This fixes issue #437.

### DIFF
--- a/src/NodaTime/CalendarSystem.cs
+++ b/src/NodaTime/CalendarSystem.cs
@@ -615,7 +615,7 @@ namespace NodaTime
         }
 
         /// <summary>
-        /// The maximum valid month (inclusive) within this calendar in the given year.
+        /// Returns the maximum valid month (inclusive) within this calendar in the given year.
         /// </summary>
         /// <remarks>
         /// It is assumed that in all calendars, every month between 1 and this month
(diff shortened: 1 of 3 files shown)


jskeet:
// cc @maggiepint

src/NodaTime/CalendarSystem.cs
```diff
@@ -634,6 +634,21 @@ public int GetMonthsInYear(int year)
             return YearMonthDayCalculator.GetMonthsInYear(year);
         }
 
+        /// <summary>
+        /// Returns the number of weeks in the given week-year.
+        /// </summary>
+        /// <param name="weekYear">The week-year to determine the number of weeks for.</param>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="weekYear"/> is outside
+        /// the range of this calendar. </exception>
```
malcolmr: nit: extra space before <

src/NodaTime/CalendarSystem.cs
```diff
@@ -634,6 +634,21 @@ public int GetMonthsInYear(int year)
...
+        /// <returns>The number of weeks in the given week-year.</returns>
+        public int GetWeeksInWeekYear(int weekYear)
+        {
+            // TODO: Is this really valid? It means we can't necessarily round-trip. Maybe be more lenient, but then check the resulting days-since-epoch.
```
malcolmr: I'm not sure I follow the reason we can't round-trip?

Perhaps add a test that demonstrates the limitation we have here?

malcolmr:
question above, but LGTM either way.

src/NodaTime/CalendarSystem.cs
jskeet: Will add a test when I start addressing the issue. Basically, a date can be in a different year to its week-year... so while it would be nice to have:

```csharp
LocalDate date = ...;
LocalDate date2 = LocalDate.FromWeekYearWeekAndDay(date.WeekYear, date.WeekofWeekYear, date.IsoDayOfWeek);
```

... work on all dates, I suspect it may fail for extremes where the week-year may be beyond the maximum year or below the minimum year.

(We also need an overload accepting a CalendarSystem, as currently it always assumes the ISO calendar...)

src/NodaTime/CalendarSystem.cs
jskeet: Done, thanks.

PR #441, opened 2016-02-14
Integration with BenchmarkDotNet

AndreyAkinshin:


jskeet:
Thanks! Will merge as-is, then work out exactly how we're going to integrate it into regular builds etc.

Issue #443, opened 2016-02-17
Add private SNK files to repo


jskeet:
As per https://msdn.microsoft.com/en-us/library/wd40t7ad.aspx:

> Do not rely on strong names for security. They provide a unique identity only.

And https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md:

>  Firstly, we would recommend that these open source projects check-in their private key

This also simplifies things as it means we can always sign assemblies regardless of project configuratoin, thus removing the ReleaseSigned configuration and the conditional bits of the InternalsVisibleTo.

PR #444, opened 2016-02-17
Add the snk "private" signing key to the public repo

jskeet:
This follows the guidance from Microsoft in
https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md

> Firstly, we would recommend that these open source projects check-in their private key
> (remember, strong-names are used for identity, and not for security).

Changes to project configurations can follow (but aren't urgent):

- Remove ReleaseSigned etc
- Sign all builds (including debug)
- Get rid of conditional keys in AssemblyInfo for InternalsVisibleTo

This fixes issue #443.


mj1856:
Sure. I guess.  I don't really like strong naming and the whole checking in of private key seems to defeat the purpose - at least from the security perspective.  It does enable the identity part of the story though - so sure.  (meh).

jskeet:
I wouldn't strong name it at all if it weren't for the viral nature of it - that without strong naming, we're stopping strongly-named assemblies from using us. But basically, I'm just following the guidance :)

At some point we could consider Authenticode-signing the assemblies if we wanted to. I'm far from expert enough on security to know the ins and outs, but I gather that's the way forward...

mj1856:
ok.  :+1:

Issue #440, opened 2016-02-12
GuessZoneIdByTransitionsUncached failing for Baja California


jskeet:
This may well be due to new time zone data - needs investigating.

mj1856:
'America/Santa_Isabel` turned out to be a mistake, was moved to a link.

jskeet:
@mj1856: The issue is that it can't find *any* appropriate time zones - would that really be explained by Santa Isabel? (I haven't investigated yet...)

jskeet:
Currently, Santa Isabel ends up being the best candidate, with a score of 31/105 - well below the 70% we require.

I note that CLDR has:

```xml
<!-- (UTC-08:00) Baja California -->
<mapZone other="Pacific Standard Time (Mexico)" territory="001" type="America/Santa_Isabel"/>
<mapZone other="Pacific Standard Time (Mexico)" territory="MX" type="America/Santa_Isabel"/>
```

... and that *is* the zone which is failing. (i.e. if we mapped by ID, we'd get the right result from CLDR.)

I've had a look at the mail you (Matt) sent to the IANA list a while ago, and that makes sense... but I'm not sure why it's scoring so badly in the current tests. I'd need to look at the exact BCL transitions... time to dig up my explorer project.

mj1856:
The mapping to `America/Santa_Isabel` *was* correct, until 2016a removed the zone entry and replaced it with a link to `America/Tijuana`.

Windows has not yet deprecated `Pacific Standard Time (Mexico) | (UTC-08:00) Baja California`, so the [CLDR trunk](http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml) currently lists it as "Unmappable".  I expect a future Windows update will set `IsObsolete = 1` on the registry entry, and then a future CLDR will drop it entirely.

Ultimately it comes down to the fact that it was a red-herring to begin with.  Either nobody used it, or if they did use it they did so incorrectly - perhaps without knowing.

mj1856:
Note that it's undetermined as of yet whether `"Pacific Standard Time (Mexico)"` will be deprecated or not.  It may just be that we update it to match the data of `"Pacific Standard Time"`.  We're discussing this now internally, and I'll update you when I know more.

jskeet:
Okay, so can I just check I understand...

[...]

PR #445, opened 2016-02-18
Ignore test failure described in issue #440

jskeet:
I hate just ignoring test failures, but in this case it makes sense to do so.


mj1856:
:+1:

PR #446, opened 2016-02-19
Remove uses of NumberFormatInfo and CompareInfo in NodaTimeFormatInfo

jskeet:
I'd like to get these two commits in before changing NodaFormatInfo.GetInstance, as the latter change is likely to require more discussion.


malcolmr:
Per discussion, let's drop df59220.

PR #448, opened 2016-02-23
Use CoreCLR on Travis

jskeet:
(Initial commit is just for diagnostics, to see what's been installed...)


jskeet:
Killed this branch; have fixed the travis build in a different way. Will still need to work out exactly what test environments we want at some point...

PR #449, opened 2016-02-26
Remove version and metadata redundancy

jskeet:
Most projects now don't need AssemblyInfo, and those that have it need less of it.
project.json keeps metadata, basically.

Microsoft projects appear not to have the equivalent of -alpha in their project.json files,
presumably specifying that only in environment variables. I propose we keep -alpha/-beta
in the master branch, to avoid finger failures accidentally creating a full release. We could
fairly easily create a myget feed to publish a build after each commit; we probably don't
want to do that on nuget.org.


jskeet:
(Assigned to Malcolm for a review more for general principles of our build process than anything else. Comments welcome.)

malcolmr:
Sounds reasonable (though I might have picked "-dev" as a suffix rather than "-alpha", but it doesn't really matter).

How does this change the generated AssemblyInfo in the built DLLs, if at all?

We'll need to update the release documentation, and presumably remove/change the tool we have for bumping these version numbers.

jskeet:
alpha comes before beta, whereas dev doesn't - the joys of precedence in package names :)

The version numbers in the assembly will be affected by DNX_BUILD_VERSION and another environment variable whose name I'm having trouble rediscovering.

And yes, the release docs and tools will need a lot of changes :) (We should probably consider using https://github.com/GitTools/GitVersion too.)

Issue #434, opened 2016-01-27
Reconsider BCL TimeZoneInfo support in light of .NET Core


jskeet:
Previously we've had:
- Pretty decent support for TimeZoneInfo on Windows desktop
- Some support for TimeZoneInfo on Windows PCL build (but with some guesswork)
- Ropey support for TimeZoneInfo on Mono, particularly TimeZoneInfo.Local

There are some assumptions that we'll always see a Windows time zone name in `TimeZoneInfo.Id`, and we use that as a map. That's already broken in some cases with odd localization, but leaving that aside, it ignores non-Windows platforms, whether running on Mono or DNX.

The two contentious aspects are:
- How well can we wrap TimeZoneInfo in a DateTimeZone?
- How well can we guess an IANA ID from a TimeZoneInfo?

jskeet:
Related to #543. Need to untangle these two.

malcolmr:
ftr, the resolution here was:

- require netstandard1.3, which supports `TimeZoneInfo.Id`.
- use that property, not `StandardName`, to map BCL zones (in practice now, just the local one) to TZDB zones.
- handle BCL zones that have TZDB IDs or unknown IDs (the latter, by guessing, as we did on PCL).

PR #450, opened 2016-02-26
Fix our Travis build, by skipping tests failing on Mono

jskeet:
Basically Mono doesn't play well with TimeZoneInfo in general. The NUnitLite
test runner fails if there's a test case source with no tests, so here we insert
a dummy one and explicitly use Assert.Ignore so we know the test is ignored.
Core CLR adopting IANA on Linux makes everything that bit trickier - this is
already covered in issue #434.

### DIFF
--- a/src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
+++ b/src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
@@ -17,13 +17,17 @@ namespace NodaTime.Test.TimeZones
     {
         // This test is effectively disabled on Mono as its time zone support is broken in the current
         // stable release - see https://github.com/nodatime/nodatime/issues/97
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
```diff
@@ -134,13 +146,13 @@ public void DateTimeMinValueStartRuleExtendsToBeginningOfTime()
         [Test]
         public void Equality()
         {
-            if (BclZonesOrEmptyOnMono.Count < 2)
+            if (BclZonesOrJustNullOnMono.Count < 2)
```
malcolmr: This doesn't seem likely to be sensible.

jskeet: Why not? Basically if we don't have two time zones to compare, we don't compare them. Fortunately, a single null reference still satisfies the "we don't have enough to compare" constraint in the same way that an empty collection does.

Issue #447, opened 2016-02-23
Switch decimal use in Duration's API to BigInteger


malcolmr:
We added `Duration.FromNanoseconds(decimal)` and `Duration.ToDecimalNanoseconds()` for 2.0.

However, since we're now (definitely?) going to depend upon .NET 4.0, we should consider replacing those with `Duration.FromNanoseconds(BigInteger)` and `Duration.ToBigIntegerNanoseconds()` instead, at least partly because `BigInteger` documents that we won't have a fractional part.

fubar-coder:
I'm not sure if this is a good change, because it prevents compilation for .NET 2.0/3.5, NETStandard 1.0, and .NET Portable (Profiles 328 and 259)

jskeet:
Noda Time 2.0 already doesn't support .NET 2.0 or 3.5. netstandard 1.0 doesn't work due to other issues - we have to use netstandard 1.3, and that's fine with BigInteger. I don't have a firm plan in terms of supporting old-school PCLs rather than netstandard... I can investigate that further though.

jskeet:
Looks like a dependency on `System.Runtime.Numerics` shouldn't block much:

https://www.nuget.org/packages/System.Runtime.Numerics/

fubar-coder:
I'll create a pull request that includes the following changes:

- Support for netstandard1.1 (first version supporting BigInteger)
- Support for net40
- Simplified dependencies

Issue #404, opened 2015-11-10
Duration should be divisible by Duration


filmor:
`Duration` objects have all the same dimension so dividing them would lead to a unit-less `float`. This makes it easy to find out "duration in hours" or "duration in days" by just writing `duration / Duration.FroumHours(1)` and `duration / Duration.FromDays(1)` respectively. If this is deemed useful I can prepare a PR for this change.

jskeet:
For 2.0, there are already TotalHours, TotalDays etc. That doesn't mean we *can't* implement division (returing a double) but it removes that particular use case. Do you find it useful for other reasons?

filmor:
It's much more generic. For example we have to count quarter hours and half hours quite often. Sure, this is possible with something like `d.TotalMinutes() / 15.0` but I think `d / Duration.FromMinutes(15)` is nicer "dimensionally". Also, check the `Align` function that I attached to #379. I don't think you would need to expose `Ticks` on `Duration` anymore.

jskeet:
Well, we still want to expose Ticks anyway - or something similar, although now possibly TotalTicks and SubsecondTicks. But yes, that sounds basically reasonable. The simplest implementation would be to just `return (double) d1.ToDecimalNanoseconds() / (double) d2.ToDecimalNanoseconds()` but we may want to do better than that...

jskeet:
Out of interest, would you be looking for a corresponding `Duration op*(Duration, double)` operator? Would kinda make sense...

filmor:
Sure, if just for symmetry. Regarding the implementation, what do you have in mind for a better implementation?

jskeet:
Well, there are various common situations which don't require going via ToDecimal to do the arithmetic. I've started looking at this now.

PR #453, opened 2016-02-26
Convert Duration to use BigInteger instead of decimal

jskeet:
Fixes issue #447.
(Still need to check performance, although relatively few operations will be affected.)

### DIFF
--- a/src/NodaTime.Benchmarks/NodaTimeTests/DurationBenchmarks.cs
+++ b/src/NodaTime.Benchmarks/NodaTimeTests/DurationBenchmarks.cs
@@ -3,6 +3,7 @@
 // as found in the LICENSE.txt file.
 using System;
 using BenchmarkDotNet.Attributes;
+using System.Numerics;
 
 namespace NodaTime.Benchmarks.NodaTimeTests
 {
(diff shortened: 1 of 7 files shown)


jskeet:
Minor data point: calling Duration.To{BigInteger/Decimal}Nanoseconds(), 10 million times, on a duration of 100 days and 5 hours:

- Decimal: ~0.6s
- BigInteger: ~1.9s

(For reference, ToInt64Nanoseconds takes about 0.05s. So whenever it's reasonable to use that - if the duration would never be more than +/- 292 years unless there are other problems - that's a good thing.)

So it's a significant (3x) but not ridiculous performance regression.
If we don't need to do this too often, that's probably okay.

Next step: working out how often we do this internally.

jskeet:
Okay, so, analysis:

- ToBigIntegerNanoseconds is called by:
  - Multiplication where either the duration operand is negative, or the result ends up being larger than +/-292 years
  - Division where the duration operand is larger than +/- 292 years

- FromBigIntegerNanoseconds is called by:
  - The above operations (BigInteger is just an intermediate format for computation)
  - Duration text parsing (relatively rare)

I think we should optimize multiplication for negative durations as well, but at that point this performance sacrifice is probably good enough.

jskeet:
Unfortunately with the state of our benchmarks at the moment I can't *verify* that this degrade performance significantly, but I don't believe it will.

This should be a simpler basis for implementing issue #404.

malcolmr:
LGTM

PR #455, opened 2016-02-27
First part of supporting double in the Duration API

jskeet:
This builds on PR #453 - it should be a single commit once that's in.


malcolmr:
LGTM, though I don't follow why we should remove operator*(long, Duration).

jskeet:
We can chat about that another time. It just feels weird to multiply a number by a duration... whereas multiplying a duration by a number makes more sense to me. I don't have precedent for it at the moment though.

jskeet:
Thanks for the review - will merge when the squashed commit has been through CI.

PR #456, opened 2016-02-28
Implement Duration division by double or another Duration

### LINKED ISSUE #404: Duration should be divisible by Duration

jskeet:
This fixes #404.


malcolmr:
LGTM, thanks!

PR #464, opened 2016-03-18
Simplify Instant.FromJulianDate using new Duration code for doubles

jskeet:
(I knew there was something this would massively simplify...)


malcolmr:
LGTM. Nice :)

Issue #360, opened 2015-03-28
NuGet package for latest TZDB


jskeet:
As discussed in mailing list thread "Point Releases".

Probably a separate NuGet package for 1.x and 2.x, released at the same time as the new tzdb is produced.

filmor:
This would be relevant right now, since Turkey surprised us with another nice DST change. How can I help with this?

jskeet:
Well if you want the latest database, you definitely don't need a new nuget package of anything - you can download the .nzd file as described in the docs. If you'd like to think about what you'd *like* the API to look like for "most recent TZDB" I'd be happy to review a pull request - but I don't have a lot of time right now, so I doubt that it'll actually be released imminently.

mj1856:
@filmor Keep in mind the Turkey change hasn't yet been confirmed.  Only if it gets published in a tzdb release would it end up in an nzd file.

jskeet:
Okay, I'd like to get to this pretty soon. A few thoughts/questions:

- If we have this package, do we need any TZDB packaged in NodaTime itself?
- What do we want the package name to be?
- Do we want to take this time to change from "TZDB" to "IANA" everywhere? (I suspect not.)
- Is it possible to target .NET 3.5 and the existing PCL profile within DNX projects? (I would really like to, as we'll need to release NodaTime.Tzdb-1.x regularly.)

malcolmr:
* I think it'd be reasonable to ship the NodaTime 2.0 package without an embedded TZDB, yes. We'd probably need to remove `DateTimeZoneProviders.Tzdb` to do that sensibly(?).
* NodaTime.Tzdb seems like a reasonable package name? (I'm not sure whether we'd want to keep the code in `NodaTime.TimeZones`, or move it to a different namespace, though. Maybe `NodaTime.TimeZones.Tzdb`?)
* No, I think TZDB is a good specific name; it would just be churn to change it. (Searching for [tzdb] brings up the IANA site as the first hit, for example.)

jskeet:
Yes, we'd want to remove `DateTimeZoneProviders.Tzdb`, at which point we'd probably want to consider doing the same for the BCL provider... which only leaves serialization, basically. Hmm.

I'm quite _tempted_ to just make this a tiny package, leaving all the rest of the code in place... moving things around feels like it may be making breaking changes for no real benefit. Hmm again.

I think I'll try to get suitable code in place for the "it's just a single file" solution first, then see how much could be moved.

[...]

PR #469, opened 2016-03-24
First pass at NodaTime.Tzdb targeting 1.3.1.

jskeet:
Once this is in and we're happy, I'll push a beta Nuget release for each of 2016a, 2016b and 2016c so folks can try it out, then a non-beta release.

I'd quite like to try to keep this same API in 2.0, so please bear that in mind when reviewing :)

// cc @mj1856


1.x-src/NodaTime.Tzdb/PackagedTzdb.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.IO;
+using System.Reflection;
+
+namespace NodaTime.Tzdb
+{
+    /// <summary>
+    /// Holder class for the TZDB data in the <c>NodaTime.Tzdb</c>
+    /// package, which is updated when IANA ships a new data release.
+    /// </summary>
+    public class PackagedTzdb
```
malcolmr: static?

malcolmr: I'm not particularly fond of `PackagedTzdb` as a name, but I'm not sure what's better; `Tzdb` alone?

1.x-src/NodaTime.Tzdb/project.json
```diff
@@ -0,0 +1,40 @@
...
+
+  "compile": [
+    "../NodaTime.Tzdb/*.cs"
+  ],
+  "resource": [
+    "../../src/NodaTime/TimeZones/Tzdb.nzd"
+  ],
+  
+  "dependencies": {
+    "NodaTime": { "version": "1.3.1", "target": "package" }
```
malcolmr: Why not 1.1.0? (needed for `TzdbDateTimeZoneSource.FromStream()`)

malcolmr:
LGTM in general, though I'm not sure why we need a separate version for 1.x vs. 2.x? From what I can see, this should work for any version ≥ 1.1.0.

1.x-src/NodaTime.Tzdb/PackagedTzdb.cs
jskeet: Should definitely be static - will fix that now.

Tzdb would be fine if it weren't in a namespace called Tzdb. We could put it in NodaTime.TimeZones instead, although it would be slightly unconventional to have a package name of NodaTime.Tzdb without that as a namespace anywhere.

Definitely up for more brainstorming on this.

1.x-src/NodaTime.Tzdb/project.json
jskeet: I did wonder about that. I haven't tried, but I suspect 1.1.0 doesn't support all the same PCL targets that we have in 1.3.1. Will give it a go.

jskeet:
2.x will have different target frameworks again. I wouldn't be able to depend on 1.1 from a netstandard 1.0 build, for example. At least, I don't think I could. It might also introduce assembly binding redirects which having two separate versions doesn't. Would be worth a try though...

jskeet:
See http://codeblog.jonskeet.uk/2016/03/28/versioning-conundrum-for-noda-time-help-requested/ for discussion on versioning in general.

malcolmr:
It sounds like (#360) we've decided not to do this in the end?

jskeet:
Yup, I think so. Will close and delete.

PR #476, opened 2016-04-08
Prepare for 1.3.2

jskeet:
Suggest you review one commit at a time. They're not in any particular order.


malcolmr:
all LGTM

jskeet:
(Merged - we should add a travis file to the branch at some point, although it won't be the same as the one in the master branch.)

Issue #374, opened 2015-08-01
Reconsider year / year-of-era in format patterns


jskeet:
It seems that the BCL Calendar.GetYear reports the year of era - and this is also what's used for "y" in a year format. We may want to swap the meaning of "y" and "Y" for Noda Time 2.0 to be in line with this - although I can see it being quite confusing.

There are relatively few calendars with multiple eras, mind you - in the BCL it's just Japanese as far as I'm aware (which we don't support) but Noda Time supports BC/AD in Gregorian/Julian calendars.

(Oh, and for some reason "yyyy" doesn't always mean "4 digits" in the BCL - it sometimes formats to only 2, which is weird. More investigation required.)

jskeet:
Note that there'd probably be a slight performance hit in making "y" mean "year-of-era" - we may need to optimize that a bit more heavily than we currently do.

jskeet:
Looking at http://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns that has:

- y: year - looks like it means "year of era"
- Y: week-year (which we don't support at all in text formatting, but possibly should)
- u: "extended year" this looks like our absolute year
- U: cyclic year name - we don't support, nor are likely to
- r: related Gregorian year - unsupported, but we could support if we wanted to

If we're going to break this for 2.0, I'm tempted to break it to be more in line with TR-35 (which is still BCL-compatible)

jskeet:
Urgh, this also affects our ISO formats etc - should we use "uuuu-MM-dd"?

mj1856:
For clarity, could you give an example of where `yyyy` and `uuuu` would be different?

jskeet:
The simplest example would be for BC dates.

June 21st, 5BC:

yyyy-MM-dd => 0005-06-21
uuuu-MM-dd => -0004-06-21 (note the sign)

None of our other calendars have multiple eras, so even though in theory this is a massively-breaking change, I don't think it will actually affect most people.

mj1856:
I'm good with that.  Thanks for the example.

jskeet:
Looking at TR-35 again, we're actually divergent from it in a number of other ways, such as ddd meaning abbreviated day name (instead of E) so possibly we *should* just be switching Y and y, effectively. I've written most of the code for this now - it's just a matter of chasing a couple more tests, and deciding for sure what we want to do (and writing the docs).

[...]

PR #480, opened 2016-04-12
Clarify year-of-era issue in test comment

jskeet:
Just to check that the new comment is rather more reasonable than the old one :)

### DIFF
--- a/src/NodaTime.Test/Text/LocalDatePatternTest.cs
+++ b/src/NodaTime.Test/Text/LocalDatePatternTest.cs
@@ -236,12 +236,15 @@ namespace NodaTime.Test.Text
                 return;
             }
 
-            // We got year-of-era and year the wrong way round in 1.x.
+            // We got year-of-era and year wrong in 1.x.
             // See https://github.com/nodatime/nodatime/issues/374
-            // It would be a breaking change to make the 2.0 change in 1.x; we could
-            // potentially change the behaviour to allow the given cultures to work
-            // by interpreting YYYY as yyyy if there's no era specifier, but for the moment
-            // these are just expected failures :(
+            // Basically, in 1.x we treat "y" as absolute year and "Y" as year-of-era.
+            // In 2.0 we treat "y" as year-of-era and "u" as absolute year.
+            // In the "known broken cultures" the format contains an era specifier (g)
+            // which we deem to be invalid due to what 1.x thinks is a lack of a year-of-era specifier.
+            // We could potentially fix this for 1.x by only complaining if we're asked to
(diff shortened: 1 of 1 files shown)


src/NodaTime.Test/Text/LocalDatePatternTest.cs
```diff
@@ -236,12 +236,15 @@ public void BclLongDatePatternGivesSameResultsInNoda(CultureInfo culture)
...
+            // We got year-of-era and year wrong in 1.x.
             // See https://github.com/nodatime/nodatime/issues/374
-            // It would be a breaking change to make the 2.0 change in 1.x; we could
-            // potentially change the behaviour to allow the given cultures to work
-            // by interpreting YYYY as yyyy if there's no era specifier, but for the moment
-            // these are just expected failures :(
+            // Basically, in 1.x we treat "y" as absolute year and "Y" as year-of-era.
+            // In 2.0 we treat "y" as year-of-era and "u" as absolute year.
+            // In the "known broken cultures" the format contains an era specifier (g)
+            // which we deem to be invalid in 1.x due to the lack of a year-of-era specifier.
```
malcolmr: Slightly confusing: it _does_ have a year-of-era specifier (`y`), just that we think that's an absolute year specifier.

src/NodaTime.Test/Text/LocalDatePatternTest.cs
```diff
@@ -236,12 +236,15 @@ public void BclLongDatePatternGivesSameResultsInNoda(CultureInfo culture)
...
+            // We could potentially fix this for 1.x by only complaining if we're asked to
```
malcolmr: Where "this" is issue #374, I assume (which would then make this workaround unnecessary).

malcolmr:
LGTM in general though.

src/NodaTime.Test/Text/LocalDatePatternTest.cs
jskeet: Right. How about "due to what 1.x thinks is a lack of a year-of-era specifier"?

src/NodaTime.Test/Text/LocalDatePatternTest.cs
jskeet: Well, sort of. We could potentially fix some symptoms without fixing the "proper" bug (as we have done in 2.0).

src/NodaTime.Test/Text/LocalDatePatternTest.cs
malcolmr: sgtm.

PR #481, opened 2016-04-13
Remove traces of code.google.com from 1.3 user guide

jskeet:
(We won't be updating the online user guide, but we ship an offline copy.)
These are all I could find easily, but I'll have another look when I build the package.

(Anything else you can think of that I might have forgotten?)


malcolmr:
Have you already diffed the `www/` part of 1.3.x against master? That's all I'd suggest.

(lgtm)

jskeet:
There are going to be quite a few changes there due to API changes etc. Will commit this, generate the package and we'll see what's been missed.

Issue #439, opened 2016-02-12
Week year boundaries are dodgy


jskeet:
Currently we use min/max year when validating week-year values. We should know what the actual min/max week years are, even though not all values within them will be valid.

jskeet:
Example: min local date in `CalendarSystem.GetGregorianCalendar(7)` has a WeekYear of -9999.
Max local date in `CalendarSystem.GetGregorianCalendar(1)` has a WeekYear of 10000.

We can't pass these to `LocalDate.FromWeekYearWeekAndDay` - although we can't pass a calendar system to that either. If text patterns supported week years, we could try it that way - but we don't at the moment.

Will try to fix this as best I can as part of #408. I may temporarily introduce an overload of `FromWeekYearWeekAndDay` just to start with broken tests...

Issue #408, opened 2015-11-16
Reconsider week calendars


jskeet:
Currently we have a situation where only the ISO calendar can be configured in terms of "minimum number of days in the first week of the week-year" and that's still fairly restrictive. The BCL allows you to specify a first-day-of-week and a week rule - although it doesn't have anything which exactly meets ISO-8601.

Perhaps instead of WeekYear and WeekOfWeekYear being properties, they should be methods which accept either a `WeekYearCalculator` or options as parameters? Then we can just have _one_ ISO calendar, and provide flexibility for _all_ calendars. Note that this would also potentially affect text parsing/formatting, and construction from week/week-year.

(We need to decide this before 2.0, as it's a breaking change.)

mj1856:
I agree that it should be a method with a parameter instead of a property.  However, I think instead of an enum like BCL's `CalendarWeekRule`, it should accept a delegate.  This would be similar to how we pass resolvers to `InZone`.

jskeet:
What would the delegate accept? If it's just a `Func<LocalDate, int>` (or similar) then having the delegate means we don't need the method... whereas the resolvers are picked and invoked appropriately only when necessary. Maybe that really *is* all we need - expose appropriate delegate implementations and be done. I was considering an interface though, which effectively exposed some bits of the `WeekYearCalculator`. Maybe I should noodle on that idea and see what it looks like (as a PR straw man).

jskeet:
This is effectively done by removing the properties from LocalDate entirely; instead we have dates on WeekYearRule.

PR #496, opened 2016-04-21
Change how we handle week-years, proposed in issue #408

jskeet:
// cc @mj1856 

We no longer have any members in most of Noda Time which deal with week-years,
other than LocalDate.FromWeekYearWeekAndDay (maintained for discoverability).
The normal "entry point" for this functionality is now WeekYearRule, an
abstract class with static factory methods. Unlike CalendarSystem, we *do*
expect that users may create their own implementations of this.

Most of the implementation came from WeekYearCalculator, but has been refactored
a bit and fixed in terms of validation. (We now round-trip all dates. This
fixes issue #439.)

This reduces the number of calendar systems we have - there's now just
Gregorian, rather than Gregorian-1 etc. ISO and Gregorian are now functionally
equivalent, but have separate IDs - we may want to reconsider that at some point
(pre-2.0).

Work still to do:
- Implement the BCL functionality (CalendarWeekRule)
- Update non-API documentation
- Update version history


jskeet:
In terms of review, I wouldn't worry too much about the implementation - I'm reasonably confident in that, and there are quite a few tests. If you could check that the descriptions in WeekYearRule make sense, along with the API removals, that would be great.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -76,10 +76,10 @@ public void Equals_DifferentCalendars()
         {
             LocalDate start1 = new LocalDate(2000, 1, 1);
             LocalDate end1 = new LocalDate(2001, 6, 19);
-            // This is a really, really similar calendar to ISO - the dates could differ by week of year,
-            // but that's all.
-            LocalDate start2 = start1.WithCalendar(CalendarSystem.GetGregorianCalendar(1));
-            LocalDate end2 = end1.WithCalendar(CalendarSystem.GetGregorianCalendar(1));
+            // This is a really, really similar calendar to ISO, but we do distinguish.
+            // TODO: Should we?
```
malcolmr: If we keep a distinction between Gregorian and Iso calendars at all, then I think yes.

src/NodaTime/CalendarOrdinal.cs
```diff
@@ -11,30 +11,24 @@ namespace NodaTime
...
-        IslamicAstronomicalBase16 = 16,
-        IslamicAstronomicalIndian = 17,
-        IslamicAstronomicalHabashAlHasib = 18,
-        IslamicCivilBase15 = 19,
-        IslamicCivilBase16 = 20,
-        IslamicCivilIndian = 21,
-        IslamicCivilHabashAlHasib = 22,
-        UmAlQura = 23,
+        Gregorian = 1,
+        Julian = 2,
```
malcolmr: Any compatibility issues in changes these values? (I assume not, given that we'd be talking about 1.3->2.0 anyway?)

src/NodaTime.Test/DateIntervalTest.cs
jskeet: Absolutely. Options are:

- Have `CalendarSystem.Iso` and `CalendarySystem.Gregorian`, but return the same thing
- Have `CalendarSystem.Iso` and `CalendarySystem.Gregorian`, but return different calendars which are deemed unequal
- Have just `CalendarSystem.Iso`
- Have just `CalendarSystem.Gregorian`

src/NodaTime/CalendarOrdinal.cs
jskeet: Exactly. It changes the serialized form, but not in a way that impacts any currently-serialized data. (And I think we're going to break serialization between 1.3 and 2.0 anyway, for the sake of sanity.)

src/NodaTime/CalendarSystem.cs
```diff
@@ -938,22 +865,16 @@ private static class MiscellaneousCalendars
...
-            internal static readonly CalendarSystem[] GregorianByMinWeekLength;
+            internal static readonly CalendarSystem Gregorian;
             internal static readonly CalendarSystem Julian;
 
             static GregorianJulianCalendars()
             {
                 var julianCalculator = new JulianYearMonthDayCalculator();
-                Julian = new CalendarSystem(CalendarOrdinal.Julian, JulianId, JulianName, new JulianYearMonthDayCalculator(), 4, new GJEraCalculator(julianCalculator));
+                Julian = new CalendarSystem(CalendarOrdinal.Julian, JulianId, JulianName, new JulianYearMonthDayCalculator(), new GJEraCalculator(julianCalculator));
                 // Variations for the calendar systems which have different objects for different "minimum first day of week"
```
malcolmr: Remove this comment?

src/NodaTime/CalendarSystem.cs
```diff
@@ -938,22 +865,16 @@ private static class MiscellaneousCalendars
...
         {
-            internal static readonly CalendarSystem[] GregorianByMinWeekLength;
+            internal static readonly CalendarSystem Gregorian;
             internal static readonly CalendarSystem Julian;
 
             static GregorianJulianCalendars()
             {
                 var julianCalculator = new JulianYearMonthDayCalculator();
-                Julian = new CalendarSystem(CalendarOrdinal.Julian, JulianId, JulianName, new JulianYearMonthDayCalculator(), 4, new GJEraCalculator(julianCalculator));
+                Julian = new CalendarSystem(CalendarOrdinal.Julian, JulianId, JulianName, new JulianYearMonthDayCalculator(), new GJEraCalculator(julianCalculator));
```
malcolmr: nit: appears to be unnecessarily creating a second `JulianYearMonthDayCalculator`?

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// of time.
+        /// </para>
+        /// </remarks>
+        /// <param name="weekYear">The week-year of the new date.</param>
+        /// <param name="weekOfWeekYear">The week of week-year of the new date.</param>
+        /// <param name="dayOfWeek">The day-of-week of the new date.</param>
+        /// <param name="calendar">The calendar system for the date.</param>
+        /// <exception cref="ArgumentOutOfRangeException">The parameters do not combine to form a valid date.</exception>
+        /// <returns>A <see cref="LocalDate"/> corresponding to the specified values.</returns>
+        public abstract LocalDate CreateLocalDate(
```
malcolmr: Seems a little odd to call this "CreateXxx" when LocalDate is a value type. `GetLocalDate`, perhaps?

src/NodaTime/CalendarSystem.cs
jskeet: Done. (All of these "done" will mean "locally" for a while...)

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// A rule determining how "week years" are arranged, including the weeks within the week year.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Dates are usually identified within a calendar system by a calendar year, a month within that
+    /// calendar year, and a day within that month. For example, the date of birth of Ada Lovelace can be identified
+    /// within the Gregorian calendar system as the year 1815, the month December (12), and the day 10. However,
+    /// dates can also be identified (again within a calendar system) by week-year, week and day-of-week. How
+    /// that identification occurs depends on which rule you use - but again as an example, within the Gregorian
+    /// calendar system, using the ISO-8601 week year rule, the date of the Unix epoch is week-year 1815,
```
malcolmr: Pretty sure it isn't (1815).

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// week 49, day-of-week Sunday.
```
malcolmr: (ditto)

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// </para>
+    /// <para>
+    /// The mapping of dates into week-year, week and day-of-week is always relative to a specific calendar system.
+    /// For example, years in the Hebrew calendar system vary very significantly in length due to leap months, and this
+    /// is reflected in the number of weeks within the week-years - as low as 50, and as high as 55.
+    /// </para>
+    /// <para>
+    /// This class allows conversions between the two schemes of identifying dates: <see cref="GetWeekYear(LocalDate)"/>
+    /// and <see cref="GetWeekOfWeekYear(LocalDate)"/> allow the week-year and week to be obtained, and
+    /// <see cref="CreateLocalDate(int, int, IsoDayOfWeek, CalendarSystem)"/> allows the reverse mapping. Note that
```
malcolmr: "obtained for a date", maybe.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// the calendar system does not need to be specified in the former methods as a <see cref="LocalDate"/> already
+    /// contains calendar information, and there is no method to obtain the day-of-week as that is not affected by the
+    /// week year rule being used.
+    /// </para>
+    /// <para>Specific rules are obtained using the static factory methods within this class, or with the
+    /// <see cref="Iso"/> property.</para>
+    /// <para>
+    /// All implementations within Noda Time are immutable, and it is advised that any external implementations
```
malcolmr: need to mention thread-safety at all? (in either case)

src/NodaTime/CalendarSystem.cs
jskeet: Fixed.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+    public abstract class WeekYearRule
+    {
+        private static readonly WeekYearRule[] isoBasedRules = Enumerable.Range(1, 7).Select(x => new IsoBasedWeekYearRule(x)).ToArray();
+
+        /// <summary>
+        /// Returns a <see cref="WeekYearRule"/> similar to ISO-8601, but allowing the minimum number of
+        /// days in the first week to vary from the value of 4 specified by ISO-8601.
+        /// </summary>
+        /// <remarks>
+        /// <paramref name="minDaysInFirstWeek"/> determines when the first week of the week-year starts.
```
malcolmr: Possibly note here or para above that ISO defines weeks to start on a Monday?

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Yup, that would probably be better. Done.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// deemed to be the first week of week-year X. Otherwise, the week is deemed to be the last week of
+        /// week-year X-1, and the first week of week-year X starts on the following Monday.
+        /// </remarks>
+        /// <param name="minDaysInFirstWeek">The minimum number of days in the first Monday-to-Sunday week
+        /// which have to be in the new calendar year for that week to count as being in that week-year.
+        /// Must be in the range 1 to 7 inclusive.
+        /// </param>
+        /// <returns>A <see cref="WeekYearRule"/> with the specified minimum number of days in the first
+        /// week.</returns>
+        public static WeekYearRule ForMinDaysInFirstWeek(int minDaysInFirstWeek)
```
malcolmr: Should this have "Iso" in the method name?

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// For any given calendar year X, consider the Monday-to-Sunday week that includes the first day of the
+        /// calendar year. Usually, some days of that week are in calendar year X, and some are in calendar year 
+        /// X-1. If <paramref name="minDaysInFirstWeek"/> or more of the days are in year X, then the week is
+        /// deemed to be the first week of week-year X. Otherwise, the week is deemed to be the last week of
+        /// week-year X-1, and the first week of week-year X starts on the following Monday.
+        /// </remarks>
+        /// <param name="minDaysInFirstWeek">The minimum number of days in the first Monday-to-Sunday week
+        /// which have to be in the new calendar year for that week to count as being in that week-year.
+        /// Must be in the range 1 to 7 inclusive.
+        /// </param>
```
malcolmr: possibly add a <see> or <seealso> to `Iso`?

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Done.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: I *think* we do that automatically when we build the docs due to it being marked immutable.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Done.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+            Preconditions.CheckArgumentRange(nameof(minDaysInFirstWeek), minDaysInFirstWeek, 1, 7);
+            return isoBasedRules[minDaysInFirstWeek - 1];
+        }
+
+        /// <summary>
+        /// Returns a <see cref="WeekYearRule"/> consistent with ISO-8601.
+        /// </summary>
+        /// <remarks></remarks>
+        /// <para>
+        /// In the standard ISO-8601 week algorithm, the first week of the year
```
malcolmr: Again, probably want to mention weeks starting on a Monday by (ISO's) definition here.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: I don't think so, because only the 4-day version is truly ISO. I couldn't come up with a better name though.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Not sure what you mean.... are you suggesting an extra sentence saying "If you want the ISO rules, just use the `Iso` property"?

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Done.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// is that in which at least 4 days are in the year. As a result of this
```
malcolmr: (Could perhaps say that this is equivalent to the first Thursday being in the first (Mon-Sun) week of the year.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// definition, day 1 of the first week may be in the previous year.
+        /// </para>
+        /// <para>
+        /// For example, January 1st 2011 was a Saturday, so only two days of that week
+        /// (Saturday and Sunday) were in 2011. Therefore January 1st is part of
+        /// week 52 of week-year 2010. Conversely, December 31st 2012 is a Monday,
+        /// so is part of week 1 of week-year 2013.
+        /// </para>
```
malcolmr: Probably worth mentioning here or above that ISO week numbers run from 1-53, since I'm not sure that's obvious.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        // TODO: Make IsoBasedWeekYearRule more flexible, allowing non-Monday-to-Sunday weeks?
+
+        /// <summary>
+        /// Creates a <see cref="LocalDate" /> from a given week-year, week within that week-year,
+        /// and day-of-week, for the specified calendar system.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Valid values for <paramref name="weekYear"/> and <paramref name="weekOfWeekYear"/> will
+        /// depend on the rule and calendar system. Rules will usually permit any week-year which is also a valid calendar year,
```
malcolmr: Perhaps "Usually, rules will permit...". I was expecting a sentence of the form "Rules will usually permit [...] _but_ [...]".

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// and sometimes the value one less than the minimum calendar year and one more than the maximum
+        /// calendar year. The valid weeks within that week-year may vary as some years are longer than others
+        /// in terms of weeks - and the first valid week-year and last valid week-year usually only contain a few
+        /// weeks each. In most applications this is unlikely to be an issue as the range of valid dates is far
```
malcolmr: Worth saying that the valid day-of-week may depend upon the week year/week? (e.g. the US system of "01/01 is in week 1, week starts on Sunday", where day-of-week 1 may not be valid.)

Also I wonder if it's worth breaking the paragraph into a bulleted list for week-year/week/day.

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        public abstract LocalDate CreateLocalDate(
+            int weekYear, int weekOfWeekYear, IsoDayOfWeek dayOfWeek, [NotNull] CalendarSystem calendar);
+
+        /// <summary>
+        /// Convenience overload to call <see cref="CreateLocalDate(int, int, IsoDayOfWeek, CalendarSystem)"/>
+        /// passing in the ISO calendar system.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Valid values for <paramref name="weekYear"/> and <paramref name="weekOfWeekYear"/> will
```
malcolmr: (as above, so below)

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <code>rule.CreateLocalDate(rule.GetWeekYear(date), rule.GetWeekOfWeekYear(date), date.IsoDayOfWeek, date.Calendar)</code>
+        /// should always return <code>date</code>. This is true for all rules within Noda Time, but third party
+        /// implementations may choose to simplify their implementations by restricting them to appropriate portions
+        /// of time.
+        /// </para>
+        /// </remarks>
+        /// <param name="weekYear">The week-year of the new date.</param>
+        /// <param name="weekOfWeekYear">The week of week-year of the new date.</param>
+        /// <param name="dayOfWeek">The day-of-week of the new date.</param>
+        /// <param name="calendar">The calendar system for the date.</param>
```
malcolmr: May implementations impose restrictions on applicable calendar systems?

(If not, does that imply that's there's some common functionality that should sit in this base class?)

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// so is part of week 1 of week-year 2013.
+        /// </para>
+        /// <value>A <see cref="WeekYearRule"/> consistent with ISO-8601.</value>
+        public static WeekYearRule Iso { get; } = isoBasedRules[3];
+
+        // TODO: public static WeekYearRule FromBclRule(CalendarWeekRule rule, IsoDayOfWeek firstDayOfWeek) => null;
+        // TODO: public static WeekYearRule FromBclRule(CalendarWeekRule rule, DayOfWeek firstDayOfWeek)
+        //    => FromBclRule(rule, BclConversions.ToIsoDayOfWeek(firstDayOfWeek));
+
+        // TODO: Make IsoBasedWeekYearRule more flexible, allowing non-Monday-to-Sunday weeks?
```
malcolmr: _Perhaps_ consider exposing it and allowing it to be extended externally, but I don't think there should be built-in support here for anything other than the standard ISO week-year-rule. (Even if we add one ourselves, I don't think we should mention it here.)

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// <para>
+        /// For example, January 1st 2011 was a Saturday, so only two days of that week
+        /// (Saturday and Sunday) were in 2011. Therefore January 1st is part of
+        /// week 52 of week-year 2010. Conversely, December 31st 2012 is a Monday,
+        /// so is part of week 1 of week-year 2013.
+        /// </para>
+        /// <value>A <see cref="WeekYearRule"/> consistent with ISO-8601.</value>
+        public static WeekYearRule Iso { get; } = isoBasedRules[3];
+
+        // TODO: public static WeekYearRule FromBclRule(CalendarWeekRule rule, IsoDayOfWeek firstDayOfWeek) => null;
```
malcolmr: Would prefer these on a separate `BclWeekYearRule` type (also see immediately below).

Basically: `CalendarSystem` is a closed set of types, so I think it'd be reasonable to enumerate the possibilities there. However, `WeekYearRule` is deliberately open, so I don't think we should make the API for accessing built-in implementations different to that for external ones (i.e. `new BclWeekYearRule()`, `new MySpecialWeekYearRule()`).

src/NodaTime/Calendars/WeekYearRule.cs
malcolmr: also if it were exposed, I'd suggest removing `ForMinDaysInFirstWeek(int)` from here in favour of `new IsoBasedWeekYearRule(int)`.

(also most of the documentation above ends up on the type then.)

src/NodaTime/Calendars/WeekYearRule.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// </para>
+    /// <para>Specific rules are obtained using the static factory methods within this class, or with the
+    /// <see cref="Iso"/> property.</para>
+    /// <para>
+    /// All implementations within Noda Time are immutable, and it is advised that any external implementations
+    /// should be immutable too.
+    /// </para>
+    /// </remarks>
+    [Immutable]
+    public abstract class WeekYearRule
```
malcolmr: This is coupling the user-facing API with our implementation, which is the same thing we did with `DateTimeZone`.

Would probably prefer an interface here (and extension method for `GetWeeksInWeekYear(int)`?), and then if we later decide it's useful to provide one or more abstract base classes for external implementations to make use of, we can do so without breaking existing implementations.

src/NodaTime.Test/DateIntervalTest.cs
malcolmr: Have just `CalendarSystem.Iso` sgtm.

malcolmr:
Looks mostly reasonable; open question as to whether `WeekYearRule` should be an abstract class or an interface is the biggest question, I think.

jskeet:
Will fix the last few doc bugs I haven't already replied to, and let's chat about the abstract class vs interface part in person. One benefit of the abstract class is that it gives us somewhere obvious to put the factory methods. (It also allows us to move common code into the abstract class at some point, when we've got more than one implementation...)

src/NodaTime.Test/DateIntervalTest.cs
jskeet: Okay. Let's discuss and make that change separately though. (I think it might be a bit confusing for users aware of the Gregorian calendar system, but not aware of ISO-8601.)

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Done for this whole example. (No mention of the unix epoch now.)

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Done.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Well, only in the ISO calendar system, of course - using the ISO numbering against the Hebrew calendar gives up to 55. Not sure where we can put that without being confusing.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: The difference being that I don't expect user A to need user B's implementation often, whereas the same user may want both the BCL and ISO implementations. I still think it's handy to have one place to get at all the Noda-provided implementations. Definitely happy to talk this over though.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Mmm. I see what you mean, but I'm not sure what to do about it. Let's brainstorm in person.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Let's try to fix that at the same time as the above.

src/NodaTime/Calendars/WeekYearRule.cs
jskeet: Yes, it seems reasonable that they might. Will mention that.

jskeet:
Pushed another commit with changes. Let's chat in person about how much more we want to do before merging, and how much we're happy to leave to later commits.

jskeet:
Documentation fixed up. Will merge when green, then create another change to see what it would look like as an interface.

Open question: how do we handle the methods which have overloads defaulting to ISO calendar? Options:

- All methods in interface, provide an abstract base class with defaulting
- Specific methods in interface, use extension methods for defaulting
- Use optional parameters

PR #501, opened 2016-04-25
Week year rule interface

jskeet:
Two-step change to move to WeekYearRule being an interface.

If we're happy with the code here, we can then brush up on the documentation (either as extra commits in this PR or as other PRs). I wanted to get the right shape of the API first though.

// cc @mj1856 for any feedback on abstract class vs interface, and extension methods vs optional parameters.

### DIFF
--- a/src/NodaTime.Test/Calendars/RegularWeekYearRuleTest.cs
+++ b/src/NodaTime.Test/Calendars/RegularWeekYearRuleTest.cs
@@ -18,7 +18,7 @@ namespace NodaTime.Test.Calendars
             // In the Gregorian calendar with a minimum of 7 days in the first
             // week, Tuesday January 1st -9998 is in week year -9999. We should be able to
             // round-trip.
-            var rule = WeekYearRule.ForMinDaysInFirstWeek(7);
+            var rule = RegularWeekYearRule.ForMinDaysInFirstWeek(7);
             var date = new LocalDate(-9998, 1, 1);
             Assert.AreEqual(date, rule.GetLocalDate(
                 rule.GetWeekYear(date),
@@ -32,7 +32,7 @@ namespace NodaTime.Test.Calendars
             // In the Gregorian calendar with a minimum of 1 day in the first
             // week, Friday December 31st 9999 is in week year 10000. We should be able to
             // round-trip.
-            var rule = WeekYearRule.ForMinDaysInFirstWeek(1);
+            var rule = RegularWeekYearRule.ForMinDaysInFirstWeek(1);
             var date = new LocalDate(9999, 12, 31);
             Assert.AreEqual(date, rule.GetLocalDate(
                 rule.GetWeekYear(date),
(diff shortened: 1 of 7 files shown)


src/NodaTime.Test/Calendars/RegularWeekYearRuleTest.cs
```diff
@@ -18,7 +18,7 @@ public void RoundtripFirstDay_Iso7()
             // In the Gregorian calendar with a minimum of 7 days in the first
             // week, Tuesday January 1st -9998 is in week year -9999. We should be able to
             // round-trip.
-            var rule = WeekYearRule.ForMinDaysInFirstWeek(7);
+            var rule = new RegularWeekYearRule(7);
```
malcolmr: I think the factory method is probably clearer..

src/NodaTime/Calendars/RegularWeekYearRule.cs
```diff
@@ -12,21 +12,63 @@ namespace NodaTime.Calendars
     // Possible future feature: allow the first day of the week to vary too.
 
     /// <summary>
-    /// WeekYearRule implementing the rules from Noda Time 1.x. Basically this is the ISO
-    /// rules, but with a variable "minimum number of days in the first week".
+    /// Implements <see cref="IWeekYearRule"/> for a rule where weeks are regular:
+    /// every week has exactly 7 days, which means that some overlap from one calendar
```
malcolmr: I'm not sure "overlap" is the right word here, since the weeks themselves don't overlap.

malcolmr:
Looks reasonable. I'd actually be okay with keeping the concrete class as something like a specific `IsoWeekYearRule`, though: do you think it's better with the generic name?

src/NodaTime.Test/Calendars/RegularWeekYearRuleTest.cs
jskeet: Okay, happy to shift that back.

src/NodaTime/Calendars/RegularWeekYearRule.cs
jskeet: You're right. Will have a think.

jskeet:
I'd rather not call it `IsoWeekYearRule` when it can end up being non-ISO... only the instance with minDaysInFirstWeek=4 is really an ISO rule. Will continue to ponder.

src/NodaTime/Calendars/RegularWeekYearRule.cs
jskeet: Straddle - that's the word I was thinking of.

> which means that some week years straddle the calendar year boundary. (So the
> start of a week can occur in one calendar year, and the end of the week in the following
> calendar year, but the whole week is in the same week-year.)

Is that clearer?

malcolmr: +1

src/NodaTime.Test/Calendars/RegularWeekYearRuleTest.cs
jskeet: Done. Made it return RegularWeekYearRule instead of just IWeekYearRule as we may want to expose MinDaysInFirstWeek as a property, for example.

jskeet:
Okay, addressed those comments. Would still like a better name than RegularWeekYearRule, but not IsoWeekYearRule, but let's keep thinking about that.

PR #520, opened 2016-05-24
Implement BCL week rules.

jskeet:
It turns out these are pretty simple:
- Either 1, 4 or 7 as the "min days in first week"
- Variable "first day of week"
- Unlike the ISO rules, the end of the calendar year is always in the same
  week-year, so the first week of the week-year may be shorter than 7 days.
  (The first days of the calendar year *may* be in the previous week-year
  though.)

There's a lot of common code between this and RegularWeekYearRule; it shouldn't
be too hard to refactor the code into a single class, providing the ability
for "regular" week year rules to have a variable first day of week too. That
can come later though, along with a whole user guide page with many, many
examples.

### DIFF
--- a/src/NodaTime.Test/Calendars/BclCalendars.cs
+++ b/src/NodaTime.Test/Calendars/BclCalendars.cs
@@ -33,7 +33,13 @@ namespace NodaTime.Test.Calendars
             }
             // Try to initialize by reflection instead...
             Type type = typeof(Calendar).GetTypeInfo().Assembly.GetType($"System.Globalization.{name}");
-            return type == null ? null : (Calendar) Activator.CreateInstance(type);
+            if (type == null)
+            {
+                // We can start being defensive if/when we try to test on a platform where
+                // this becomes a problem.
+                throw new Exception($"Unable to get calendar {name}");
+            }
+            return (Calendar) Activator.CreateInstance(type);
         }
 
         public static Calendar Hebrew => GetCalendar("HebrewCalendar");
@@ -43,6 +49,14 @@ namespace NodaTime.Test.Calendars
         public static Calendar Julian => GetCalendar("JulianCalendar");
         public static Calendar Hijri => GetCalendar("HijriCalendar");
 
+        /// <summary>
+        /// Returns a sequence of all the BCL calendar systems for which we have a
+        /// mapping in Noda Time. The first entry is Gregorian so that it's easy to
+        /// comment out the rest for initial testing of a new feature (where the
+        /// Gregorian calendar is typically easy to reason about).
+        /// </summary>
+        public static IEnumerable<Calendar> MappedCalendars =>
+            new[] { Gregorian, Hebrew, UmAlQura, Persian, Julian, Hijri };
 
         /// <summary>
         /// Tries to work out a roughly-matching calendar system for the given BCL calendar.
--- a/src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
+++ b/src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
@@ -1,0 +1,78 @@
+﻿// Copyright 2016 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using NUnit.Framework;
+using System;
+using System.Globalization;
+using static System.Globalization.CalendarWeekRule;
+using static System.DayOfWeek;
+using NodaTime.Calendars;
+
+namespace NodaTime.Test.Calendars
+{
+    public class BclWeekYearRuleTest
+    {
+        /// <summary>
+        /// For each calendar and rule combination, check everything we can about every date
+        /// from mid December to mid January around each year between 2016 and 2046.
+        /// (For non-Gregorian calendars, the rough equivalent is used...)
+        /// That should give us plenty of coverage.
+        /// </summary>
+        [Test]
+        [Combinatorial]
+        public void BclEquivalence(
+            [ValueSource(typeof(BclCalendars), nameof(BclCalendars.MappedCalendars))] Calendar calendar,
+            [Values(FirstDay, FirstFourDayWeek, FirstFullWeek)] CalendarWeekRule bclRule,
+            [Values(Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday)]  DayOfWeek firstDayOfWeek)
+        {
+            var nodaCalendar = BclCalendars.CalendarSystemForCalendar(calendar);
+            var nodaRule = BclWeekYearRule.FromCalendarWeekRule(bclRule, firstDayOfWeek);
+            var startYear = new LocalDate(2016, 1, 1).WithCalendar(nodaCalendar).Year;
+
+            for (int year = startYear; year < startYear + 30; year++)
+            {
+                var startDate = new LocalDate(year, 1, 1, nodaCalendar).PlusDays(-15);
+                for (int day = 0; day < 30; day++)
+                {
+                    var date = startDate.PlusDays(day);
+                    var bclDate = date.ToDateTimeUnspecified();
+                    var bclWeek = calendar.GetWeekOfYear(bclDate, bclRule, firstDayOfWeek);
+                    // Weird... the BCL doesn't have a way of finding out which week-year we're in.
+                    // We're starting at "start of year - 15 days", so a "small" week-of-year
+                    // value means we're in "year", whereas a "large" week-of-year value means
+                    // we're in the "year-1".
+                    var bclWeekYear = bclWeek < 10 ? year : year - 1;
+
+                    Assert.AreEqual(bclWeek, nodaRule.GetWeekOfWeekYear(date), "Date: {0}", date);
+                    Assert.AreEqual(bclWeekYear, nodaRule.GetWeekYear(date), "Date: {0}", date);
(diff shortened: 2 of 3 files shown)


jskeet:
(If you have any suggestions about what a single week-year-rule class with factory methods for ISO-like rules and BCL-like rules would be called, that would be great :) `StandardWeekYearRule`, `SimpleWeekYearRule`, `WeekYearRule`...?

src/NodaTime.Test/Calendars/BclCalendars.cs
```diff
@@ -43,6 +49,8 @@ private static Calendar GetCalendar(string name)
         public static Calendar Julian => GetCalendar("JulianCalendar");
         public static Calendar Hijri => GetCalendar("HijriCalendar");
 
+        public static IEnumerable<Calendar> MappedCalendars =>
```
malcolmr: Probably at least some trivial documentation would be nice: it's not obvious to me what this is.

src/NodaTime.Test/Calendars/BclCalendars.cs
```diff
@@ -43,6 +49,8 @@ private static Calendar GetCalendar(string name)
...
+            new[] { Gregorian, Hebrew, UmAlQura, Persian, Julian, Hijri };
```
malcolmr: Any reason this is out of order compared to the list above?

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        [Test]
+        [Combinatorial]
+        public void BclEquivalence(
+            [ValueSource(typeof(BclCalendars), nameof(BclCalendars.MappedCalendars))] Calendar calendar,
+            [Values(FirstDay, FirstFourDayWeek, FirstFullWeek)] CalendarWeekRule bclRule,
+            [Values(Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday)]  DayOfWeek firstDayOfWeek)
+        {
+            var nodaCalendar = BclCalendars.CalendarSystemForCalendar(calendar);
+            var nodaRule = BclWeekYearRule.GetRule(bclRule, firstDayOfWeek);
```
malcolmr: nit: it seems odd to be passing something described as a 'rule' (`bclRule`) into a method called `GetRule()`.

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
```diff
@@ -0,0 +1,75 @@
...
+            for (int year = startYear; year < startYear + 30; year++)
+            {
+                var startDate = new LocalDate(year, 1, 1, nodaCalendar).PlusDays(-15);
+                for (int day = 0; day < 30; day++)
+                {
+                    var date = startDate.PlusDays(day);
+                    var bclDate = date.ToDateTimeUnspecified();
+                    var bclWeek = calendar.GetWeekOfYear(bclDate, bclRule, firstDayOfWeek);
+                    // Weird... the BCL doesn't have a way of finding out which week-year we're in.
+                    var bclWeekYear = bclWeek < 10 ? year : year - 1;
```
malcolmr: Probably worth adding to the comment: we start at start-of-year-minus-15-days (which is easy to miss, btw), and so 'small' week-of-year values mean that we're in the current year, whereas larger ones mean that we're in the prior year.

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
```diff
@@ -0,0 +1,75 @@
...
+
+                    Assert.AreEqual(bclWeek, nodaRule.GetWeekOfWeekYear(date), "Date: {0}", date);
+                    Assert.AreEqual(bclWeekYear, nodaRule.GetWeekYear(date));
```
malcolmr: No Date: message for this assertion?

src/NodaTime/Calendars/BclWeekYearRule.cs
```diff
@@ -0,0 +1,215 @@
...
+    /// <see cref="Calendar.GetWeekOfYear(DateTime, CalendarWeekRule, DayOfWeek)"/>.
+    /// </summary>
+    [Immutable]
+    public sealed class BclWeekYearRule : IWeekYearRule
+    {
+        private readonly CalendarWeekRule rule;
+        private readonly IsoDayOfWeek firstDayOfWeek;
+        private readonly int minDaysInFirstWeek;
+
+        // TODO: Expose properties?
```
malcolmr: You don't have any properties.

src/NodaTime/Calendars/BclWeekYearRule.cs
```diff
@@ -0,0 +1,215 @@
...
+                case FirstDay:
+                    minDaysInFirstWeek = 1;
+                    break;
+                case FirstFourDayWeek:
+                    minDaysInFirstWeek = 4;
+                    break;
+                case FirstFullWeek:
+                    minDaysInFirstWeek = 7;
+                    break;
+                default: throw new ArgumentException($"No such CalendarWeekRule: {rule}", nameof(rule));
```
malcolmr: nit: $"unhandled case: {rule}" would make more sense: clearly there _is_ such a CalendarWeekRule if you're passing it in?

(or are .NET enum values unconstrained; I forget.)

malcolmr: (also, is this idiomatic formatting?)

src/NodaTime/Calendars/BclWeekYearRule.cs
```diff
@@ -0,0 +1,215 @@
...
+                return calendarYear;
+            }
+        }
+
+        /// <summary>
+        /// Validate that at least one day in the calendar falls in the given week year.
+        /// </summary>
+        private void ValidateWeekYear(int weekYear, CalendarSystem calendar)
+        {
+            // FIXME: Adjust for BCL truncation.
```
malcolmr: ?

malcolmr:
LGTM: I trust your tests to prove whether the maths is correct.

src/NodaTime/Calendars/BclWeekYearRule.cs
jskeet: Yeah, that was meant as "Expose the above fields as properties".

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
jskeet: Well, it's converting a rule from one form (the BCL enum) to the Noda equivalent. CreateRule?

src/NodaTime.Test/Calendars/BclCalendars.cs
jskeet: It makes it easier to comment out all but Gregorian when debugging ;)

src/NodaTime/Calendars/BclWeekYearRule.cs
jskeet: Will remove the FIXME - it's done now.

src/NodaTime/Calendars/BclWeekYearRule.cs
jskeet: We cover all the defined rules; this would be an unknown enum value like `(CalendarWeekRule) 100`.

Will change the formatting for the throw - at one point this was in a factory method and all the cases were returns.

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
malcolmr: I think that's just as problematic (which is to say, not really): you have a BCL rule and you want to make a Noda Time rule, but the two things aren't equivalent.

perhaps just s/bclRule/calendarWeekRule/ (per firstDayOfWeek, which isn't called bclDay).

src/NodaTime/Calendars/BclWeekYearRule.cs
malcolmr: Ah, I wasn't sure if `(CalendarWeekRule) 100` was possible, thanks.

(Also, you might be covering all the defined cases now, but someone might add one to the enum in the future, in which case "unhandled case" makes more sense. That's probably more sensible with Java-style constrained enum values, though, rather than C#/C++ aliases-for-int.)

jskeet: Changed message to "Unsupported CalendarWeekRule" and the value.

src/NodaTime/Calendars/BclWeekYearRule.cs
jskeet: Comment amended.

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
jskeet: Done.

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
jskeet: Done.

src/NodaTime.Test/Calendars/BclWeekYearRuleTest.cs
jskeet: Have renamed method to FromCalendarWeekRule, and also the parameter.

src/NodaTime.Test/Calendars/BclCalendars.cs
jskeet: Noted, and added comment for property.

jskeet:
Merging now, and will then start looking at how we can refactor to a single class.

PR #525, opened 2016-05-28
Refactor weekrule

jskeet:
It's probably worth reviewing each commit here separately.

This takes us as far as having all the week-year-rule code in one class, but that class basically has the wrong name. Suggestions welcome. Once we've renamed, I can write documentation - which will probably be a whole new page.


src/NodaTime/Calendars/RegularWeekYearRule.cs
```diff
@@ -70,7 +71,31 @@ private RegularWeekYearRule(int minDaysInFirstWeek)
...
+
+        /// <summary>
+        /// Creates a week year rule where the boundary between one week-year and the next
+        /// is parameterized in terms of how many days of the first week of the week
+        /// year have to be in the new calendar year, and also by which day is deemed
+        /// to be the first day of the week.
+        /// </summary>
+        /// <remarks>
+        /// <paramref name="minDaysInFirstWeek"/> determines when the first week of the week-year starts.
+        /// For any given calendar year X, consider the Monday-to-Sunday week that includes the first day of the
```
malcolmr: Not always Monday-to-Sunday.

src/NodaTime/Calendars/RegularWeekYearRule.cs
```diff
@@ -70,7 +71,31 @@ private RegularWeekYearRule(int minDaysInFirstWeek)
...
+        /// calendar year. Usually, some days of that week are in calendar year X, and some are in calendar year 
+        /// X-1. If <paramref name="minDaysInFirstWeek"/> or more of the days are in year X, then the week is
+        /// deemed to be the first week of week-year X. Otherwise, the week is deemed to be the last week of
+        /// week-year X-1, and the first week of week-year X starts on the following Monday.
```
malcolmr: likewise.

src/NodaTime/Calendars/BclWeekYearRule.cs
```diff
@@ -82,13 +82,25 @@ public LocalDate GetLocalDate(int weekYear, int weekOfWeekYear, IsoDayOfWeek day
...
+                int daysIntoWeekYear = (weekOfWeekYear - 1) * 7 + daysIntoWeek;
+                int days = startOfWeekYear + daysIntoWeekYear;
+                if (days < calendar.MinDays || days > calendar.MaxDays || daysIntoWeekYear < 0)
                 {
                     throw new ArgumentOutOfRangeException(nameof(weekYear),
                         $"The combination of {nameof(weekYear)}, {nameof(weekOfWeekYear)} and {nameof(dayOfWeek)} is invalid");
                 }
-                return new LocalDate(yearMonthDayCalculator.GetYearMonthDay(days).WithCalendar(calendar));
+                LocalDate ret = new LocalDate(yearMonthDayCalculator.GetYearMonthDay(days).WithCalendar(calendar));
+                // The requested week year and the actual calendar year don't match, it could be due to short weeks.
```
malcolmr: "If the requested..."?

src/NodaTime/Calendars/RegularWeekYearRule.cs
```diff
@@ -124,7 +175,19 @@ public LocalDate GetLocalDate(int weekYear, int weekOfWeekYear, IsoDayOfWeek day
                     throw new ArgumentOutOfRangeException(nameof(weekYear),
                         $"The combination of {nameof(weekYear)}, {nameof(weekOfWeekYear)} and {nameof(dayOfWeek)} is invalid");
                 }
-                return new LocalDate(yearMonthDayCalculator.GetYearMonthDay(days).WithCalendar(calendar));
+                LocalDate ret = new LocalDate(yearMonthDayCalculator.GetYearMonthDay(days).WithCalendar(calendar));
+                // The requested week year and the actual calendar year don't match, it could be due to short weeks
+                // in a non-regular rule. The simplest way to find out is just to check what the week year is...
+                // This problem doesn't apply to rules with regular weeks.
```
malcolmr: That's a bit confusing: it's not a 'problem' at all in the ISO case: it's a valid state.

Perhaps reword as "For rules with irregular weeks, the requested week year and actual calendar year should always match. If they don't, the caller has specified an invalid combination (like the Thursday in week 2016-01, which doesn't exist)."

(I also wonder if we should invert the sense of the boolean, or name it more specifically as e.g. `actuallyUseBclShortWeeks` - or something :).)

malcolmr:
lgtm.

and perhaps `WeekYearRules` or `CommonWeekYearRules` would make sense.

src/NodaTime/Calendars/RegularWeekYearRule.cs
jskeet: Fixed.

src/NodaTime/Calendars/RegularWeekYearRule.cs
jskeet: It is for this factory method - because you don't pass in a first day of week. I'll clarify that a bit.

src/NodaTime/Calendars/RegularWeekYearRule.cs
jskeet: Yes, given that we almost always use `!regular` that would make sense. Will update the comment first, then reverse the sense in a different commit.

jskeet:
Right, think I'm done now apart from the docs. Will merge this on green, and write the docs separately.

PR #528, opened 2016-06-06
Add detailed week-year documentation

jskeet:
Hope these make sense...

### DIFF
--- a/www/unstable/userguide/calendars.md
+++ b/www/unstable/userguide/calendars.md
@@ -25,6 +25,13 @@ Note that although the API version listed indicates when a calendar system was f
 the access to most calendars was simplified in 2.0 to use properties instead of methods. The API access listed
 is from 2.0.0 onwards; please refer to the relevant copy of this user guide if you are using an older version.
 
+As of Noda Time 2.0.0, calendar systems themselves only deal with
+concepts of eras, years, months and days - not the alternative
+mapping of a date to a "week year, week of week year, day of week".
+That is now represented by
+[`IWeekYearRule`](noda-type://NodaTime.Calendars.IWeekYearRule).
+Details are in the [week year page](weekyears.html).
+
 ISO
 ===
 
@@ -35,14 +42,11 @@ This is the default calendar system when constructing values without explicitly 
 It is designed to correspond to the calendar described in [ISO-8601](http://en.wikipedia.org/wiki/ISO_8601),
 and is equivalent to the Gregorian calendar in all respects other than the century and year-of-century values.
 
-As of Noda Time 2.0, century and year-of-century properties have been removed as they almost certainly didn't
-behave how everyone would have expected, at which point the ISO calendar system and the Gregorian calendar system
-(with a "minimum number of days in the first week" of 4) are equivalent. The separation between them in Noda Time
-is maintained for simplicity, compatibility and consistency.
-
-The ISO calendar system always has at least 4 days in the first week of the week-year, where a week always
-runs Monday to Sunday. The first week of the week-year is therefore always the one containing the first Thursday
-in the year.
+As of Noda Time 2.0, century and year-of-century properties have
+been removed, at which point the ISO calendar system and the
+Gregorian calendar system are equivalent. The separation between
+them in Noda Time is maintained for simplicity, compatibility and
+consistency.
 
 Gregorian
 ===
@@ -63,9 +67,6 @@ at different times. Noda Time's implementation is *proleptic*, in that it extend
 no support for a "cutover" calendar system which observes the Julian calendar until some point, at which point it
 cuts over to the Gregorian calendar system. Although such a calendar system would more accurately represent
 the calendar observed in many countries over time, it causes all sorts of API difficulties.
-
-The parameter to `GetGregorianCalendar()` indicates the minimum number of days in the first week of the week-year;
-the `Gregorian` properties returns the same calendar as `GetGregorianCalendar(4)`.
 
 Julian
 ===
--- a/www/unstable/userguide/weekyears.md
+++ b/www/unstable/userguide/weekyears.md
@@ -1,0 +1,303 @@
+---
+layout: userguide
+title: Week years
+category: advanced
+weight: 3037
+---
+
+Introducing week-years
+===
+
+Most of the time, dates are represented as a year, month and day in
+a particular calendar system - occasionally with an era as well. An
+alternative approach is to represent a date as a year, week,
+and day-of-week.
+
+For example, using the ISO-8601 rule, the year/month/day combination
+"2016, February, 4" is equivalently year 2016, week 5,
+Thursday. All of this is performed relative to a specific calendar
+system.
+
+However, under the same rule the year/month/day combination "2014,
+December, 29th" is equivalently year **2015**, week 1, Monday: the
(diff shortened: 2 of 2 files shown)


www/unstable/userguide/calendars.md
```diff
@@ -35,14 +42,12 @@ This is the default calendar system when constructing values without explicitly
...
-behave how everyone would have expected, at which point the ISO calendar system and the Gregorian calendar system
-(with a "minimum number of days in the first week" of 4) are equivalent. The separation between them in Noda Time
-is maintained for simplicity, compatibility and consistency.
-
-The ISO calendar system always has at least 4 days in the first week of the week-year, where a week always
-runs Monday to Sunday. The first week of the week-year is therefore always the one containing the first Thursday
-in the year.
+As of Noda Time 2.0, century and year-of-century properties have
+been removed as they almost certainly didn't behave how everyone
+would have expected, at which point the ISO calendar system and the
```
malcolmr: I know this is pre-existing, but perhaps simplify as "everyone expected".

(also, seems like some odd rewrapping going on here?)

www/unstable/userguide/weekyears.md
```diff
@@ -0,0 +1,291 @@
...
+category: advanced
+weight: 3037
+---
+
+Introducing week-years
+===
+
+Most of the time, dates are represented as a year, month and day in
+a particular calendar system - occasionally with an era as well. An
+alternative approach is to represent a date as a week-year, week,
```
malcolmr: perhaps "as a year, week, and day-of-week" here, then a following sentence to introduce the idea that we have to use a slightly different year (called a week-year) for some cases?

www/unstable/userguide/weekyears.md
```diff
@@ -0,0 +1,291 @@
...
+and day-of-week. This is performed relative to a specific calendar
+system, along with a rule used to work out when each week-year starts.
+
+For example, using the ISO-8601 rule, the year/month/day combination
+"2016, February, 4" is equivalently week-year 2016, week 5,
```
malcolmr: In fact, if you move the "This is performed relative" line above to below this para, you could use "year" here again, and then follow with "However, the year/month/day combination "2016, January, 1" falls into year 2015, week 53 [TODO: substitute actual valid example], and so we call this type of year a "week-year".

Or similar.

www/unstable/userguide/weekyears.md
```diff
@@ -0,0 +1,291 @@
...
+[`IWeekYearRule`](noda-type://NodaTime.Calendars.IWeekYearRule)
+interface, and built-in implementations are obtained via factory
+members in the
+[`WeekYearRules`](noda-type://NodaTime.Calendars.WeekYearRules)
+static class.
+
+Although the `GetWeeksInWeekYear` and `GetLocalDate` methods accept
+`CalendarSystem` parameters, extension methods in
+[`WeekYearRuleExtensions`](noda-type://NodaTime.Calendars.WeekYearRuleExtensions)
+are equivalent but without the parameters, simply delegating to the
```
malcolmr: Although the `GetWeeksInWeekYear` and `GetLocalDate` methods require a `CalendarSystem`, we also provide equivalent extension methods in `WeekYearRuleExtensions` that the default to the ISO calendar system, delegating to the existing `IWeekYearRule` methods. This reduces...

www/unstable/userguide/weekyears.md
```diff
@@ -0,0 +1,291 @@
...
+2014-12-29 to 2015-01-04 inclusive
+- January 1st 2016 is a Friday, so **2015 week 53** runs from
+2015-12-28 to 2016-01-03 inclusive, and 2016 week 1 runs from
+2016-01-04 to 2016-01-10 inclusive.
+- January 1st 2017 is a Sunday, so **2016 week 52** runs from
+2016-12-26 to 2017-01-01 inclusive, and 2017 week 1 runs from
+2017-01-02 to 2017-01-08 inclusive.
+
+Once you understand this, it's not terribly hard to add two other
+axes of flexibility:
```
malcolmr: I'd like a clearer bridge sentence between "this is _the_ ISO calendar system" and "here are some ISO-like calendars we can construct".

Perhaps replace this with something like "ISO-8601 only defines a single rule for calculating week years, but it is possible to vary some of the parameters to produce ISO-like rules:"

www/unstable/userguide/weekyears.md
```diff
@@ -0,0 +1,291 @@
...
+
+- First day of week: for example, if you want Sunday-to-Saturday
+weeks, but still "four days or more" then the first day of the
+calendar year is in week 1 if it's a Sunday, Monday, Tuesday or
+Wednesday.
+- The required number of days in the week containing the first day
+of the calendar year to make that day part of week 1. For example,
```
malcolmr: List item one is "Item: text". Item two is "Item. Text". Please parallelise :)

www/unstable/userguide/weekyears.md
```diff
@@ -0,0 +1,291 @@
...
+It's easiest to think of the BCL rules as being equivalent to the
+ISO-like rules in most respects, with a mapping from
+`CalendarWeekRule` to "minimum number of days in first week" of:
+
+- `FirstDay`: 1
+- `FirstFourDayWeek`: 4
+- `FirstFullWeek`: 7
+
+There's then one subtle difference between ISO-like rules and
+BCL-like rules:
```
malcolmr: Perhaps add a summary here:

"
There's then one subtle difference between ISO-like rules and BCL-like rules: BCL weeks do not always have seven days.

> (quote)

"

malcolmr:
lgtm

www/unstable/userguide/calendars.md
jskeet: Have just removed the motivation. Meh :)
Rewrapped as well.

www/unstable/userguide/weekyears.md
jskeet: Hmm. Have tried something, but it feels a bit clunky. Will merge it anyway, but let's pair to improve.

www/unstable/userguide/weekyears.md
jskeet: Done.

www/unstable/userguide/weekyears.md
jskeet: Done.

www/unstable/userguide/weekyears.md
jskeet: Failing to follow this right now, but will have another look when less sleepy.

jskeet: Okay, I *think* I've now got something useful. Will consult with you to check.

www/unstable/userguide/weekyears.md
jskeet: That removes the symmetry between "In an ISO-like rule, X. In a BCL-like rule, Y." I should probably explain that this isn't really quoting anything stated elsewhere - it's just to highlight it.

jskeet: That removes the symmetry between "In an ISO-like rule, X. In a BCL-like rule, Y." I should probably explain that this isn't really quoting anything stated elsewhere - it's just to highlight it.

PR #532, opened 2016-06-10
Add the ability to provide an "override" file for Windows zone mappings

### LINKED ISSUE #473: Consider including CLDR mapping overrides at time of NZD release
(Mentioned in #470, but worth its own discussion.)

When we update our NZD files for each TZDB release, we're including CLDR mappings for conversion to Windows time zones.  We use the current shipping production release for this.

While this seems like the right thing to do, it can unfortunately lead to errors.  The problem is, CLDR is on a [strict bi-annual release cadence](http://cldr.unicode.org/index#TOC-General-Schedule-).  This doesn't align well with either TZDB or Windows time zone releases.

TZDB releases on an as-needed basis, either when there is sufficient critical mass, or when reacting to a short-notice change by a government.

For Windows, traditionally, tz/dst hotfixes are released as necessary to react to changes, then cumulative updates are done in August and December.  This is a bit in flux, and the team is working on improvements to be more streamlined.  Either way, it doesn't align with CLDR's strict schedule.

This means that there's an interm period where CLDR maintainers know about a change and commit it [to the dev trunk here](http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml), until the final release version comes out.  Sometimes this delay is significant.

Consider recent events as an example.  TZDB introduced three new time zones in 2016b, `Europe/Astrakhan`, `Europe/Ulyanovsk` and `Asia/Barnaul`.  These aren't in CLDR 29, because CLDR 29 was released with TZDB 2016a data.  They are, however, in the dev version, mapping as best as possible to the current Windows time zones.  Additionally, [Windows will release an update on April 12th](https://blogs.technet.microsoft.com/dst2007/2016/03/15/time-zone-updates-for-russia/) that will add 4 new time zones.  After those are released, CLDR mappings will likely change once again.  However, none of these CLDR changes will be *released* until CLDR 30 comes out on September 15th.

It would seem to make sense to include in Noda Time's NZD files some delta of changes that we review manually.  That may also require NZD releases more frequently than just with each TZDB release, potentially.

jskeet:
This fixes #473 in terms of code - process still needs definition.

### DIFF
--- a/src/NodaTime.Test/Tests.nunit
+++ b/src/NodaTime.Test/Tests.nunit
@@ -1,9 +1,0 @@
-﻿<NUnitProject>
-  <Settings activeconfig="Debug" />
-  <Config name="Debug" binpathtype="Auto">
-    <assembly path="bin\Debug\NodaTime.Test.dll" />
-  </Config>
-  <Config name="Release" binpathtype="Auto">
-    <assembly path="bin\Release\NodaTime.Test.dll" />
-  </Config>
-</NUnitProject>
--- a/src/NodaTime.TzdbCompiler.Test/NodaTime.TzdbCompiler.Test.xproj
+++ b/src/NodaTime.TzdbCompiler.Test/NodaTime.TzdbCompiler.Test.xproj
@@ -14,5 +14,8 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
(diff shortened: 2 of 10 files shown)


mj1856:
I'll try this out, and build an overrides file to match the Windows data being released on Tuesday (June 14th).

mj1856:
Here's the overrides file.  Still need to try it out.  Will work on that this weekend.
https://gist.github.com/mj1856/85da329b72f298a5824abfb25c7f1dfd

jskeet:
Ping?

mj1856:
Sorry... Been dealing with some personal issues.  I'll get back to this soon.  Thanks.

jskeet:
Merging this as the code is pretty simple; Matt, let me know if it needs any more changes later.

Issue #372, opened 2015-07-28
Consider all target platforms and dependencies for 2.0.


jskeet:
Related to issue #329.

With .NET Core on the horizon, it would seem foolish not to support that. The details of exactly how that should be supported are still changing (or at least being revealed), but NodaTime-2.0.0-alpha20150728 supports it. However, this meant significant changes:

- Json.NET was updated to 7.0.1 (as that supports the same profile)
- Our PCL target has changed quite a bit, and probably doesn't support everything that 1.3.1 did

I'm *hoping* there'll be a way of supporting .NET Core without updating Json.NET quite as far - although "New feature - Added JsonDictionaryContract.DictionaryKeyResolver" suggests that we might be able to fix or just close issue #237 if we do.

It will partly depend on when we release, of course. I suspect there's still another 3-6 months of work on 2.0, by the time we've looked at serialization, revisited bits of the API around time zones etc.

mj1856:
Highly relevant:  http://oren.codes/2015/07/29/targeting-net-core/

jskeet:
Thanks. Not *entirely* sure where all of that leaves us - I need to read through it a few more times, I think. (IMO, MS still has a way to go in streamlining PCL development.)

onovotny:
From what I can see, your current PCL targets .net 4 and SL5 in addition to windows 8. That PCL won't work on .NET Core-based runtimes like .NET Native or CoreCLR.

There are really two options here:
1. Drop .NET 4 and SL5 and support a minimum of net45. This will give you a PCL that will work. 
2. Create a new csproj PCL project using VS 2015 with the UWP tools installed. Select .NET 4.6, Windows Universal 10 and ASPNet 5 as the targets. The resulting PCL project will reference the new .NET Core packages. You can use shared projects or linked files to include all of your existing source files. 

After either of those options, you just need to put an additional copy in `\lib\dotnet` and add the nuspec dependencies to the BCL packages. Adding a reference to my tool, NuSpec.ReferenceGenerator, to your PCL project will do that for you automatically.

jskeet:
Right - the first option is what I believe I've implemented in the Profile259 branch. It's not clear to me whether that means we *have* to update the Json.NET dependency to 7.0.1 or not, although it's probably okay to do so anyway.

onovotny:
[...]

PR #537, opened 2016-06-28
Most of the way to hitting .NET Core 1.0 RTM.

jskeet:
Two issues remain:
- Some unmapped time zones (possibly fixed by an override file); probably unrelated to .NET Core
- NodaTime.Serialization.Test fails to execute due to conflicting Json.NET versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,15 @@ addons:
 mono:
   - latest
 
+# TODO: See whether we could use apt sources above instead.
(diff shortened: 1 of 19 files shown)


jpsingleton:
Does this mean https://github.com/nodatime/nodatime/issues/372 can be closed?

jskeet:
Nearly. Still want to consider whether there are any benefits in targeting portable *and* netstandard - will ask Oren on the issue...

PR #541, opened 2016-07-06
Support for net45, and netstandard1.1 and simplified dependencies

fubar-coder:
Using NETStandard.Library seems to be the preferred way to include all standard dependencies. The new `TypeExtensions` class is required for .NET 4.0 support, because the code generated for the resource files use `GetTypeInfo().Assembly` to get the assembly of a given type.

When the change from `decimal` to `BigInteger` is reverted, then we should be able to support net35 and PCLs too.


jskeet:
Will look at this soon, but just to say that "when the change from decimal to BigInteger is reverted" seems to be assuming that it *will* be reverted... and I have no intention in that direction at the moment. From what I can see, it looks like `BigInteger` should be fine in various PCL profiles. It's definitely a cleaner solution, and I don't want to saddle 2.x with a messier solution just for the sake of older platforms (which can continue using 1.x). The installation numbers I've seen around desktop framework versions certainly don't support continuing with 3.5 when there's any kind of technical cost involved (as there is here).

fubar-coder:
The "when" was **not** meant as "will", just to be clear. I'm fine with the dependency on `BigInteger`, but I really need net45 (netstandard1.1) compatibility and it's quite easy to make this possible (as you can see in the PR).

jskeet:
The dependencies are deliberately trimmed, as per
https://docs.microsoft.com/en-us/dotnet/articles/core/deploying/reducing-dependencies
I gather there's some discussion at the moment as to what the best practice genuinely is, and of course I'll continue to update the project file until this has stabilized.

However, I'm concerned about the fact that just changing the dependency allows it to build for netstandard1.1 - when with the trimmed dependencies, netstandard1.3 is required. This is something that has bothered me about the NETStandard library - it seems to be a single metapackage which contains everything, regardless of whether it's *actually* available. I'm worried that the packages which aren't available for netstandard1.1 when using the trimmed dependencies will fail at execution time in a netstandard1.1-only environment. If you have links which would resolve that concern, I'd love to see them - I'd really like to target netstandard1.0 if we possibly can.

As for targeting .NET 4.0... I'm less interested in this, to be honest. I doubt that many developers will be genuinely tied to .NET 4.0. The change right now is manageable, but I worry about what future opportunities will be blocked by this. (Async is the most obvious problem, but that would be fixed by a 4.0-only dependency on Microsoft.Bcl.Async, I guess.)

fubar-coder:
Both `XmlDocument` and `System.Threading.Thread` are not needed. I'll reduce the dependencies and test it again.

fubar-coder:
I changed `NodaTime.Serialization.JsonNet` to be usable by `net45` and `netstandard1.1`. I also reduced/optimized some dependencies. The nunit tests run without problems.

jskeet:
Thanks - I suspect that an earlier version (back in DNX days) required those dependencies.

Definitely happy to support net45 and netstandard1.1.

Could you revert the net40 support for the moment (including TypeExtensions)? I'd then be happy to accept this PR (I'll squash it in Github, so don't worry about doing that). The net40 support can be in a different PR for later consideration.

jskeet:
(Oh, and then let me know if you'd like to be included in AUTHORS.txt, and how...)

fubar-coder:
Done

jskeet:
Cracking, thanks. Merging now.

fubar-coder:
Is it possible to create a new alpha package? It's somewhat blocking me for [BeanIO for .NET](https://github.com/FubarDevelopment/beanio-net)

jskeet:
Will try to do so before the end of the weekend. Tracked in #544.

fubar-coder:
Thank you

PR #553, opened 2016-07-16
Fix travis (maybe)

jskeet:
Suspect this won't help... haven't seen this travis failure before.


jskeet:
This didn't fix Travis, but is worth doing anyway. See https://github.com/dotnet/core/issues/227.

PR #558, opened 2016-07-23
Add the ability to coalesce zone intervals

jskeet:
...  when asking for all the zone intervals in an interval.

This is slightly odd as it reuses an existing enum and not every
option will affect the result, but I think it's okay.

(I needed this for ZoneDumper, and it feels like it's generally useful.)

### DIFF
--- a/src/NodaTime/DateTimeZone.cs
+++ b/src/NodaTime/DateTimeZone.cs
@@ -630,5 +630,39 @@ namespace NodaTime
                 current = zoneInterval.RawEnd;
             }
         }
+
+        /// <summary>
+        /// Returns the zone intervals within the given interval, potentially coalescing some of the
+        /// original intervals according to options.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This is equivalent to <see cref="GetZoneIntervals(Interval)"/>, but may coalesce some intervals.
+        /// For example, if the <see cref="ZoneEqualityComparer.Options.OnlyMatchWallOffset"/> is specified,
+        /// and two consecutive zone intervals have the same offset but different names, a single zone interval
+        /// will be returned instead of two separate ones. When zone intervals are coalesced, all aspects of
+        /// the first zone interval are used except its end instant, which is taken from the second zone interval.
+        /// </para>
+        /// <para>
+        /// As the options are only used to determine which intervals to coalesce, the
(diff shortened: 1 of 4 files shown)


src/NodaTime/DateTimeZone.cs
```diff
@@ -630,5 +630,39 @@ public IEnumerable<ZoneInterval> GetZoneIntervals(Interval interval)
...
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public IEnumerable<ZoneInterval> GetZoneIntervals(Interval interval, ZoneEqualityComparer.Options options)
+        {
+            if ((options & ~ZoneEqualityComparer.Options.StrictestMatch) != 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(options),
+                    $"The value {options} is not defined within ZoneEqualityComparer.Options");
+            }
+            var zoneIntervalEqualityComparer = new ZoneEqualityComparer.ZoneIntervalEqualityComparer(options, interval);
```
malcolmr: Any reason not to use the public API i.e. `ZoneEqualityComparer.ForInterval(interval).WithOptions(options)`?

src/NodaTime/DateTimeZone.cs
```diff
@@ -630,5 +630,39 @@ public IEnumerable<ZoneInterval> GetZoneIntervals(Interval interval)
...
+        /// the intervals returned.
+        /// </para>
+        /// </remarks>
+        /// <param name="interval">Interval to find zone intervals for. This is allowed to be unbounded (i.e.
+        /// infinite in both directions).</param>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public IEnumerable<ZoneInterval> GetZoneIntervals(Interval interval, ZoneEqualityComparer.Options options)
+        {
+            if ((options & ~ZoneEqualityComparer.Options.StrictestMatch) != 0)
```
malcolmr: Unless you particularly need to avoid extra frames on the stack trace, this check (and message) are already part of constructing a `ZoneEqualityComparer`.

malcolmr:
LGTM

src/NodaTime/DateTimeZone.cs
jskeet: Indeed. I wondered about making it public, but it felt like you shouldn't need to create an equality comparer just to get the zone intervals in a particular way.

Issue #561, opened 2016-07-28
Support for C++/CLI


fubar-coder:
I'll create a pull request for this (special) use case. C++/CLI (since 2013, tools version 13?) can consume net45 assemblies, but NuGet support for C++/CLI is worse than for C++/native.

fubar-coder:
Any estimated time frame for a new build? Otherwise we'll just use our private feed with a NuGet package version `2.0.0-alpha20160727`, which should be updatable by the official package.

jskeet:
I can do a new build in the next couple of days, but it would still only be alpha of course.

jskeet:
Presumably I need this in NodaTime.Testing and NodaTime.Serialization.JsonNet as well?

fubar-coder:
It shouldn't be necessary, because using NodaTime.Testing from a C# or VB assembly should be sufficient. Newtonsoft.Json currently doesn't support C++/CLI, so I guess that it's not necessary for NodaTime.Serialization.JsonNet. However, I'll try to get the PR for C++/CLI merged as soon as possible (and when James NK approves...).

jskeet:
I've pushed 2.0.0-alpha20160729 - if you could check that it works for you and report back, we can close this :)

fubar-coder:
Works as expected. Thank you very much!

PR #562, opened 2016-07-28
Add C++/CLI support (nodatime/nodatime#561)

fubar-coder:


jskeet:
Do you have a reference for this? Aside from anything else, this looks like it's only covering the net45 version - what if you want to use C++/CLI when targeting .NET Core? Or does that not work at all?

fubar-coder:
Visual C++ 2013 targets .NET 4.5, Visual C++ 2015 targets .NET 4.5.2. I don't think that it's possible to develop .NET Core assemblies with C++/CLI.

* [.NET Programming with C++/CLI (Visual C++)](https://msdn.microsoft.com/en-us/library/68td296t.aspx)
* [Will CoreCLR support C++/CLI crossplat? (dotnet/coreclr#659)](https://github.com/dotnet/coreclr/issues/659)
* [How to: Modify the Target Framework and Platform Toolset](https://msdn.microsoft.com/library/ff770576.aspx)

I just assume that .NET 4.5 is the minimum supported version and don't catch usage from VS 2012 (.NET 4.0).

src/NodaTime/project.json
```diff
@@ -13,6 +13,12 @@
     "requireLicenseAcceptance": false,
     "repository": {
       "url": "https://github.com/nodatime/nodatime.git"
+    },
+    "files": {
+      "mappings": {
+        "build/": "../../packaging/NodaTime.targets",
+        "lib/native/": [ "../../packaging/_._" ]
```
jskeet: Do we need this line? We don't have any native code in Noda Time.

jskeet:
I'm happy enough with the NodaTime.targets file - not sure about the _._ file and link though.

src/NodaTime/project.json
fubar-coder: Yes, we need this, because the `_._` creates an empty directory (`lib/native`), which is found/used by NuGet while installing the package. Otherwise, the package cannot be installed to a C++/CLI project. The TFM for C++ and C++/CLI is both `native`.

jskeet: Okay. This is highly weird black magic, IMO - I hate the idea that every library author needs to include this due to one awkward language. I hope it gets fixed in later tooling.

PR #565, opened 2016-08-01
The targets file only needs to be added to the C++ projects.

fubar-coder:


jskeet:
I'll take your word for it :)

PR #574, opened 2016-08-30
Pedantic - added Northern Hemisphere to "winter"

tymski:
This is a tiny change to be extra precise about the fact that the help guid talks about winter in the Northern Hemisphere. Being pedantic about time/date is good, right? ;)


jskeet:
Yup, pedantry definitely welcomed - thanks :)

Craigfis:
More pedantry: I think it's a help guide, not a help guid. ;-)

PR #604, opened 2016-12-15
Use reference equality for time zones other than FixedDateTimeZone

### LINKED ISSUE #180: Remove IEquatable support from DateTimeZone in favour of ZoneEqualityComparer
```
We don't support this properly (see e.g. issue 135), and so it's better not to 
pretend that we do.  We should remove the interface and the Equals/EqualsImpl 
(etc) support from DateTimeZone.
```

Original issue reported on code.google.com by `malcolm.rowe` on 25 Jan 2013 at 9:32

jskeet:
IDateTimeZoneProvider is now documented to return equal zones when asked for the same
ID, which CachedDateTimeZoneProvider will do by way of caching for all non-fixed-offset zones, and
regular equality for FixedOffsetDateTimeZone.

Fixes #180.

### DIFF
--- a/src/NodaTime.Test/DateTimeZoneTest.cs
+++ b/src/NodaTime.Test/DateTimeZoneTest.cs
@@ -30,7 +30,6 @@ namespace NodaTime.Test
             var zone2 = DateTimeZone.ForOffset(offset);
 
             Assert.AreNotSame(zone1, zone2);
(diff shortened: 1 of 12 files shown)


malcolmr:
LGTM.

(I haven't exhaustively checked whether there's anything else to update, but this looks fine as-is.)

PR #613, opened 2016-12-28
Revisit time zone mapping

### LINKED ISSUE #282: Reconsider the SPI for time zone mapping
```
IDateTimeZoneSource.MapTimeZoneId() accepts a TimeZoneInfo, but we only ever 
call it with one value, the local time zone.  It seems like this is 
unnecessarily general, and we should probably reconsider for 2.0
```

Original issue reported on code.google.com by `malcolm.rowe` on 10 May 2014 at 10:13

### LINKED ISSUE #434: Reconsider BCL TimeZoneInfo support in light of .NET Core

### LINKED ISSUE #543: Investigate use of TimeZoneInfo.Id in .NET Core
It looks like Id is now a supported property, when it wasn't before in PCLs.
However, we'll need to work out what to do about its behaviour on OS-X and Linux...

jskeet:
- IDateTimeZoneSource now only exposes GetSystemDefaultId
- Use TimeZoneInfo.Id instead of TimeZoneInfo.StandardName
  (requires netstandard1.3)
- Ignore the standard names field when reading NZD files. We
  still generate them for 1.x clients.
- If we can't find the local ID as a Windows ID, see if it's already
  a TZDB ID
- If we still can't find it, guess based on transitions.
- Fix missing standard name mappings.

Fixes #282.

Fixes #543.

Fixes #434.

(Apologies for several changes in one... they're all inter-related, and it was simpler to do them all in one go.)

### DIFF
--- a/src/NodaTime/TimeZones/DateTimeZoneCache.cs
+++ b/src/NodaTime/TimeZones/DateTimeZoneCache.cs
@@ -80,15 +80,10 @@ namespace NodaTime.TimeZones
         /// <inheritdoc />
         public DateTimeZone GetSystemDefault()
         {
-            TimeZoneInfo bcl = TimeZoneInfo.Local;
-            string id = source.MapTimeZoneId(bcl);
+            string id = source.GetSystemDefaultId();
             if (id == null)
             {
-#if PCL
-                throw new DateTimeZoneNotFoundException("TimeZoneInfo name " + bcl.StandardName + " is unknown to source " + VersionId);
-#else
-                throw new DateTimeZoneNotFoundException("TimeZoneInfo ID " + bcl.Id + " is unknown to source " + VersionId);
-#endif
+                throw new DateTimeZoneNotFoundException($"System default time zone is unknown to source {VersionId}");
             }
             return this[id];
         }
@@ -114,7 +109,8 @@ namespace NodaTime.TimeZones
                     zone = source.ForId(id);
                     if (zone == null)
                     {
-                        throw new InvalidDateTimeZoneSourceException("Time zone " + id + " is supported by source " + VersionId + " but not returned");
+                        throw new InvalidDateTimeZoneSourceException(
+                            $"Time zone {id} is supported by source {VersionId} but not returned");
                     }
                     timeZoneMap[id] = zone;
                 }
@@ -130,7 +126,7 @@ namespace NodaTime.TimeZones
                 var zone = GetZoneOrNull(id);
                 if (zone == null)
                 {
-                    throw new DateTimeZoneNotFoundException("Time zone " + id + " is unknown to source " + VersionId);
+                    throw new DateTimeZoneNotFoundException($"Time zone {id} is unknown to source {VersionId}");
                 }
                 return zone;
             }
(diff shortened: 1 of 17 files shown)


src/NodaTime/TimeZones/IO/TzdbStreamData.cs
```diff
@@ -63,12 +62,6 @@ internal sealed class TzdbStreamData
         /// </summary>
         public IList<TzdbZone1970Location> Zone1970Locations { get; }
 
-        /// <summary>
-        /// Additional mappings from Windows standard name to TZDB ID. Primarily used in
-        /// the PCL build, where we can't get at the system ID. This never returns null.
-        /// </summary>
-        public IDictionary<string, string> WindowsAdditionalStandardNameToIdMapping { get; }
```
malcolmr: Update `docs/developer/tzdb-file-format.md` to note that we're no longer using this field's data?

src/NodaTime/TimeZones/DateTimeZoneCache.cs
```diff
@@ -80,15 +80,10 @@ public DateTimeZoneCache([NotNull] IDateTimeZoneSource source)
...
-            string id = source.MapTimeZoneId(bcl);
+            string id = source.GetSystemDefaultId();
             if (id == null)
             {
-#if PCL
-                throw new DateTimeZoneNotFoundException("TimeZoneInfo name " + bcl.StandardName + " is unknown to source " + VersionId);
-#else
-                throw new DateTimeZoneNotFoundException("TimeZoneInfo ID " + bcl.Id + " is unknown to source " + VersionId);
-#endif
+                throw new DateTimeZoneNotFoundException("TimeZoneInfo ID " + id + " is unknown to source " + VersionId);
```
malcolmr: `id` will always be null here; I don't think that's what you mean.

src/NodaTime/TimeZones/IO/TzdbStreamData.cs
jskeet: Done - and added field 7, which wasn't documented before...

src/NodaTime/TimeZones/DateTimeZoneCache.cs
jskeet: Whoops, yes. Have also changed the source in this file to use interpolated string literals.

jskeet:
Have made changes in a separate commit - will rebase before merging, but it should make it easier for you to review.

PR #614, opened 2016-12-28
Rename LocalDate.IsoDayOfWeek (et al) to DayOfWeek

### LINKED ISSUE #472: Reconsider IsoDayOfWeek vs DayOfWeek
All our calendars use normal week days. We don't support non-7-day weeks etc.

Maybe keep the name IsoDayOfWeek for the enum (it has different values to the BCL DayOfWeek, after all) but ditch the separation between the enum and the integer property in 2.0?

jskeet:
(Numeric versions are removed.)

This is done as several commits for clarity - will preserve them (rather than squashing) when I merge.
I don't *think* we refer to the properties in the user guide...

Fixes #472.

### DIFF
--- a/docs/userguide/unstable/migration-to-2.md
+++ b/docs/userguide/unstable/migration-to-2.md
@@ -74,6 +74,14 @@ methods introduced in `DateTimeOffset` in .NET 4.6:
 - The `Ticks` property is now `ToUnixTimeTicks()`
 - There are two new methods, `ToUnixTimeSeconds()` and `ToUnixTimeMilliseconds()`
 
+Static properties on the pattern classes have been renamed to remove the `Pattern` suffix. For example,
+`LocalDateTimePattern.ExtendedIsoPattern` is now just `LocalDateTimePattern.ExtendedIso`.
+
+The `IsoDayOfWeek` properties in `LocalDate`, `LocalDateTime`, `OffsetDateTime`
+and `ZonedDateTime` are now just called `DayOfWeek`. The previous numeric `DayOfWeek` properties
+have been removed, but in all cases if you were actually calling them, you can just cast the `IsoDayOfWeek`
+to `int` and always get the same result, as all calendar systems use ISO days of the week.
+
 Period
 ====
 
@@ -87,9 +95,6 @@ setting date-based units.
 
 Normalization of a period which has time units which add up to a "days" range outside the range
 of `int` will similarly fail.
-
-Static properties on the pattern classes have been renamed to remove the `Pattern` suffix. For example,
-`LocalDateTimePattern.ExtendedIsoPattern` is now just `LocalDateTimePattern.ExtendedIso`.
 
 Offset
 ====
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -153,10 +153,8 @@ namespace NodaTime
         internal int DaysSinceEpoch => Calendar.GetDaysSinceEpoch(yearMonthDayCalendar.ToYearMonthDay());
 
         /// <summary>
-        /// Gets the week day of this local date expressed as an <see cref="NodaTime.IsoDayOfWeek"/> value,
-        /// for calendars which use ISO days of the week.
+        /// Gets the week day of this local date expressed as an <see cref="NodaTime.IsoDayOfWeek"/> value.
         /// </summary>
-        /// <exception cref="InvalidOperationException">The underlying calendar doesn't use ISO days of the week.</exception>
         /// <value>The week day of this local date expressed as an <c>IsoDayOfWeek</c>.</value>
         public IsoDayOfWeek DayOfWeek => Calendar.GetIsoDayOfWeek(yearMonthDayCalendar.ToYearMonthDay());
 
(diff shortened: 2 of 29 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -639,7 +626,7 @@ public LocalDate WithCalendar([NotNull] CalendarSystem calendar)
         public LocalDate PlusWeeks(int weeks) => DatePeriodFields.WeeksField.Add(this, weeks);
 
         /// <summary>
-        /// Returns the next <see cref="LocalDate" /> falling on the specified <see cref="IsoDayOfWeek"/>.
+        /// Returns the next <see cref="LocalDate" /> falling on the specified <see cref="DayOfWeek"/>.
```
malcolmr: I think this is supposed to be a reference to the type, not the property, but I guess the fact that you've changed it suggests that this should actually be `<see cref="NodaTime.IsoDayOfWeek"/>`.

src/NodaTime/LocalDate.cs
```diff
@@ -667,7 +654,7 @@ public LocalDate Next(IsoDayOfWeek targetDayOfWeek)
         }
 
         /// <summary>
-        /// Returns the previous <see cref="LocalDate" /> falling on the specified <see cref="IsoDayOfWeek"/>.
+        /// Returns the previous <see cref="LocalDate" /> falling on the specified <see cref="DayOfWeek"/>.
```
malcolmr: (Likewise)

src/NodaTime/LocalDateTime.cs
```diff
@@ -801,7 +788,7 @@ public LocalDateTime WithCalendar([NotNull] CalendarSystem calendar)
         public LocalDateTime PlusNanoseconds(long nanoseconds) => TimePeriodField.Nanoseconds.Add(this, nanoseconds);
 
         /// <summary>
-        /// Returns the next <see cref="LocalDateTime" /> falling on the specified <see cref="IsoDayOfWeek"/>,
+        /// Returns the next <see cref="LocalDateTime" /> falling on the specified <see cref="DayOfWeek"/>,
```
malcolmr: (likewise)

src/NodaTime/LocalDateTime.cs
```diff
@@ -815,7 +802,7 @@ public LocalDateTime WithCalendar([NotNull] CalendarSystem calendar)
         public LocalDateTime Next(IsoDayOfWeek targetDayOfWeek) => new LocalDateTime(date.Next(targetDayOfWeek), time);
 
         /// <summary>
-        /// Returns the previous <see cref="LocalDateTime" /> falling on the specified <see cref="IsoDayOfWeek"/>,
+        /// Returns the previous <see cref="LocalDateTime" /> falling on the specified <see cref="DayOfWeek"/>,
```
malcolmr: ditto

src/NodaTime/LocalDate.cs
```diff
@@ -276,7 +263,7 @@ public static LocalDate FromYearMonthWeekAndDay(int year, int month, int occurre
             Preconditions.CheckArgumentRange(nameof(dayOfWeek), (int) dayOfWeek, 1, 7);
 
             // Correct day of week, 1st week of month.
-            int week1Day = (int) dayOfWeek - startOfMonth.DayOfWeek + 1;
+            int week1Day = dayOfWeek - startOfMonth.DayOfWeek + 1;
```
jskeet: For any enum `Foo`, there's a unary `-` operator as `int -(Foo x, Foo y)` (assuming `int` is the underlying type).

src/NodaTime/LocalDate.cs
jskeet: Oh yes. Looks like it's always been broken. Will fix in all places - thanks.

docs/userguide/unstable/migration-to-2.md
```diff
@@ -88,6 +88,11 @@ setting date-based units.
 Normalization of a period which has time units which add up to a "days" range outside the range
 of `int` will similarly fail.
 
+The `IsoDayOfWeek` properties in `LocalDate`, `LocalDateTime`, `OffsetDateTime`
+and `ZonedDateTime` are now just called `DayOfWeek`. The previous numeric `DayOfWeek` properties
+have been removed, but in call cases if you were actually calling them, you can just cast the `IsoDayOfWeek`
```
malcolmr: Typo: `call` -> `all`.

www/unstable/userguide/migration-to-2.md
```diff
@@ -81,6 +81,11 @@ methods introduced in `DateTimeOffset` in .NET 4.6:
 - The `Ticks` property is now `ToUnixTimeTicks()`
 - There are two new methods, `ToUnixTimeSeconds()` and `ToUnixTimeMilliseconds()`
 
+The `IsoDayOfWeek` properties in `LocalDate`, `LocalDateTime`, `OffsetDateTime`
+and `ZonedDateTime` are now just called `DayOfWeek`. The previous numeric `DayOfWeek` properties
+have been removed, but in call cases if you were actually calling them, you can just cast the `IsoDayOfWeek`
```
malcolmr: (ditto)

PR #619, opened 2016-12-29
Prefer fixed offset zones from the source rather than FixedOffsetZone

### LINKED ISSUE #332: Bcl provider returns non-bcl zones
```
I would expect something like this to work:

    var bcl = DateTimeZoneProviders.Bcl;
    var zones = bcl.Ids.Select(id => bcl[id])
                       .Cast<BclDateTimeZone>()
                       .Select(x => new {x.Id, x.DisplayName})
                       .ToList();

But it fails since the Bcl provider returns a few zones that are 
FixedDateTimeZone instead of BclDateTimeZone.   Specifically, they are "UTC", 
"UTC+12", "UTC-02", and "UTC-11".

While I understand that these zones *can* be represented as a fixed offset, 
shouldn't they still be delivered as a BclDateTimeZone since I used the Bcl 
provider?

Casting to a BclDateTimeZone is the only way to get at DisplayName, which is 
how I found this.  Other than calling TimeZoneInfo.FindSystemTimeZoneById, I 
can't think of another way to get back a display name such as "(UTC) 
Coordinated Universal Time" without hardcoding it.
```

Original issue reported on code.google.com by `mj1856` on 26 Sep 2014 at 8:58

jskeet:
The source is only consulted for IDs it advertises, but that includes ones that are fixed, e.g. "UTC-2".

Fixes #332.

### DIFF
--- a/docs/userguide/unstable/versions.md
+++ b/docs/userguide/unstable/versions.md
@@ -60,6 +60,8 @@ Other:
 - Added a `ReturnForwardShifted` resolver, which shifts values in the daylight saving time "gap" forward by the duration of the gap,
   effectively returning the instant that would have occurred had the gap not existed.  This was added to support the new behavior of
   the "lenient" resolver (see above), but can also be used separately.
+- When an `IDateTimeZoneSource` advertises a zone with an ID corresponding to a fixed-offset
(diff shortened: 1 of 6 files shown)


jskeet:
Whoops - had tried to find text to change in `IDateTimeZoneProvider`, but failed. Will do that and version history now.

PR #620, opened 2016-12-29
Document guarantees around DateTimeZone.Utc equality.

malcolmr:
In particular, document that it will be equal to (though not necessarily the same instance as[*]) whatever `DateTimeZone.ForOffset(Offset.Zero)` returns, and clarify that it may not be equal to any zone returned by `IDateTimeZoneProvider["UTC"]` (as after #332 is resolved, it won't be).

Refs #180.

[*] In practice, `DateTimeZone.ForOffset(Offset.Zero)` _will_ be the same instance as `DateTimeZone.Utc` (and we have tests to check that), but it doesn't need to be.

### DIFF
--- a/src/NodaTime/DateTimeZone.cs
+++ b/src/NodaTime/DateTimeZone.cs
@@ -97,10 +97,13 @@ namespace NodaTime
         internal const string UtcId = "UTC";
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


malcolmr:
(Note that the inverse relationship — that `DateTimeZone.ForOffset(Offset.Zero)` is equal to `DateTimeZone.Utc` — is already documented.)

PR #621, opened 2016-12-30
Implement additional equality comparers for DateInterval.

### LINKED ISSUE #383: Add an IEqualityComparer for DateInterval that canonicalises
Split out from #380, in which @jskeet says:

> I think it makes sense to have the default equality comparison be "naive", but expose an IEqualityComparer which does canonicalize.

In other words, while the innate equality comparison considers the range [2001-01-01, 2001-01-31] to be distinct from [2001-01-01, 2001-02-01), this one would consider them equivalent.

(Note that we should handle ranges covering minimum and maximum dates, which may mean that this isn't as simple as canonicalising the range.)

One open question: are empty intervals — e.g. [2001-01-01, 2001-01-01), [2001-02-01, 2001-02-01) — all equivalent? They cover the same set of dates (i.e. none), but have different start dates.

malcolmr:
Specifically, add `DateInterval.NormalizingEqualityComparer` and `ContainedDatesEqualityComparer`.

`NormalizingEqualityComparer` additionally considers as equal two intervals that differ only in whether they are closed or half-open (inclusive or exclusive), meaning that [2001-01-01, 2001-01-31] is considered equal to [2001-01-01, 2001-02-01).  However, empty intervals with different start dates are still considered unequal.

(This is documented in terms of normalizing all intervals to exclusive intervals, since that's mathematically simpler; however, in practice we cannot actually represent all inclusive intervals as exclusive, and so we implement this by normalizing in the _other_ direction, then treating empty intervals as a special case.)

`ContainedDatesEqualityComparer` additionally considers all empty intervals (for the same calendar) to be equal, effectively implementing equality on the set of dates contained within the interval.

Fixes #383.

### DIFF
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -4,12 +4,15 @@
 using NodaTime.Text;
 using NUnit.Framework;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace NodaTime.Test
 {
     public class DateIntervalTest
     {
         private static readonly CalendarSystem JulianCalendar = CalendarSystem.Julian;
+        private static readonly List<CalendarSystem> SupportedCalendars = CalendarSystem.Ids.ToList().Select(CalendarSystem.ForId).ToList();
 
         [Test]
         public void Construction_DifferentCalendars()
@@ -77,7 +80,6 @@ namespace NodaTime.Test
             LocalDate start1 = new LocalDate(2000, 1, 1);
             LocalDate end1 = new LocalDate(2001, 6, 19);
             // This is a really, really similar calendar to ISO, but we do distinguish.
-            // TODO: Should we?
             LocalDate start2 = start1.WithCalendar(CalendarSystem.Gregorian);
             LocalDate end2 = end1.WithCalendar(CalendarSystem.Gregorian);
             var interval1 = new DateInterval(start1, end1, false);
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -169,6 +171,423 @@ public void Equals_DifferentToOtherType()
...
+            IEqualityComparer<DateInterval> comparer = DateInterval.NormalizingEqualityComparer;
+            Assert.IsTrue(comparer.Equals(interval1, interval2));
+            Assert.AreEqual(comparer.GetHashCode(interval1), comparer.GetHashCode(interval2));
+        }
+
+        [Test]
+        public void NormalizingEqualityComparer_Extremes_Equal()
+        {
+            LocalDate start = new LocalDate(-9998, 1, 1);
+            LocalDate end1 = new LocalDate(9999, 12, 30);
```
jskeet: Would LocalDate.MinValue/MaxValue be useful here? They'd always be in the ISO calendar, but it feels like it's a handy thing if only to make tests like this simpler.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -169,6 +171,423 @@ public void Equals_DifferentToOtherType()
...
+            var interval2 = new DateInterval(start, end2, false);
+            Assert.AreNotEqual(interval1, interval2);
+
+            IEqualityComparer<DateInterval> comparer = DateInterval.NormalizingEqualityComparer;
+            Assert.IsFalse(comparer.Equals(interval1, interval2));
+            Assert.AreNotEqual(comparer.GetHashCode(interval1), comparer.GetHashCode(interval2));  // probably
+        }
+
+        [Test]
+        public void NormalizingEqualityComparer_Extremes_Unequal_DifferentCalendars()
```
jskeet: These don't look like extreme intervals to me... what am I missing?

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -169,6 +171,423 @@ public void Equals_DifferentToOtherType()
...
+        public void NormalizingEqualityComparer_DateIntervalToItself()
+        {
+            LocalDate start = new LocalDate(2000, 1, 1);
+            LocalDate end = new LocalDate(2001, 6, 19);
+            var interval = new DateInterval(start, end, false);
+            Assert.IsTrue(DateInterval.NormalizingEqualityComparer.Equals(interval, interval));
+        }
+
+        [Test]
+        public void ContainedDatesEqualityComparer_EqualValues()
```
jskeet: Are these the same tests as above, just with the different comparer? If so, I wonder whether it might be clearer to have a single set of tests, each of which check all comparers.

I suspect that with a bit of work, using parameterized test cases might be simpler too, but I'm not as sure about that.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -169,6 +171,423 @@ public void Equals_DifferentToOtherType()
...
+        {
+            LocalDate date1 = new LocalDate(2000, 2, 1);
+            LocalDate date2 = new LocalDate(2000, 2, 2, CalendarSystem.Gregorian);
+            var interval1 = new DateInterval(date1, date1, false);
+            var interval2 = new DateInterval(date2, date2, false);
+            Assert.AreEqual(0, interval2.Length);
+            Assert.AreEqual(0, interval1.Length);
+            Assert.AreNotEqual(interval1, interval2);
+
+            // Different calendar systems are still unequal, even if they're both empty.
```
jskeet: At some point we could *consider* having yet another equality comparer (or two) which effectively normalized calendars as well. (Every non-ISO date can be converted to an ISO date, so it would be simple to normalize both intervals to ISO ones, and then compare.)

src/NodaTime/DateInterval.cs
```diff
@@ -204,5 +246,108 @@ public override string ToString()
...
+                    return NormalizingDateIntervalEqualityComparer.Normalize(obj);
+                }
+                var calendar = obj.Start.Calendar;
+                return calendar == CalendarSystem.Iso ? CanonicalIsoEmptyInterval : CreateCanonicalEmptyInterval(calendar);
+            }
+
+            private static DateInterval CreateCanonicalEmptyInterval(CalendarSystem calendar)
+            {
+                // This can use any arbitrary date for the given calendar, so long as it's a valid date.
+                var date = new LocalDate(calendar.MinYear, 1, 1, calendar);
```
jskeet: I had to mentally check that this really *would* be valid, even in calendars where the first month of the year isn't 1, but I think it is...

src/NodaTime.Test/DateIntervalTest.cs
malcolmr: Oops, fixed, thanks.

src/NodaTime/DateInterval.cs
malcolmr: I wasn't completely sure either, which is why the parameterized `ContainedDatesEqualityComparer_EmptyRange` test exists, to check all the calendars.

src/NodaTime.Test/DateIntervalTest.cs
malcolmr: We could; let's see if there's demand first, I think.

src/NodaTime.Test/DateIntervalTest.cs
malcolmr: (Of course, this particular case would be `LocalDate.MinValue.PlusDays(-1)`.)

From what I can see, it would only really help this test and `Calendars/SimpleWeekYearRuleTest.cs`, but we could consider it.

I've pulled these out to test constants though, which seems a lot more sensible than what I had before.

src/NodaTime.Test/DateIntervalTest.cs
malcolmr: They are. Let's see what happens if I merge the test cases.

malcolmr: Okay, done, and it's a lot easier to follow.

I'm not so sure about parameterized test cases: it'd be possible, but I think somewhat confusing here.

What might be useful would be for `TestHelper` to be able to simply test innate equality and/or comparers when we simply want to express "X and Y should be/should not be equal", rather than the current helpers which require exactly one equal and one non-equal example.

PR #623, opened 2016-12-31
Address "tick-oriented" APIs

### LINKED ISSUE #617: Check LocalTime tick constructor
Currently we have this constructor (amongst others)

```csharp
public LocalTime(int hour, int minute, int second, int millisecond, int tickWithinMillisecond)
```

I'm nervous of any constructors using ticks... possibly change it to a static factory method? (We already have FromHourMinuteSecondTick, so this would be FromHourMinuteSecondMillisecondTick)

jskeet:
- Convert LocalTime constructors accepting ticks to factory methods
- Remove LocalDateTime constructors accepting ticks
- Add Instant.PlusNanos to match Instant.PlusTicks. (Both are probably still useful.)

Fixes #617, although we should probably look over all core APIs at some point before 2.0.

### DIFF
--- a/docs/userguide/unstable/migration-to-2.md
+++ b/docs/userguide/unstable/migration-to-2.md
@@ -82,6 +82,9 @@ and `ZonedDateTime` are now just called `DayOfWeek`. The previous numeric `DayOf
 have been removed, but in all cases if you were actually calling them, you can just cast the `IsoDayOfWeek`
 to `int` and always get the same result, as all calendar systems use ISO days of the week.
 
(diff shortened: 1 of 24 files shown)


jskeet:
(I'd suggest reviewing one commit at a time :)

jskeet:
Will merge as is on the grounds that it's better than before, but I take your point... I wouldn't personally mind removing them. (Lots of tests would need to change, but that's easy enough to do.) Let's think about it when we review the whole API.

PR #624, opened 2016-12-31
Changes to IFormatProvider handling

### LINKED ISSUE #134: ToString(pattern, IFormatProvider) probably broken for non-CultureInfo format providers
```
I haven't tested this, but Instant (and LocalDate, LocalDateTime, etc) have a 
ToString(String pattern, IFormatProvider) method that I suspect is broken if 
the IFormatProvider is not a CultureInfo.

That is, it seems that it would be reasonable to pass a NumberFormatInfo or 
DateTimeFormatInfo directly, but NodaFormatInfo.GetInstance() will ignore 
those, and just return a NodaFormatInfo for the current thread's culture 
instead.
```

Original issue reported on code.google.com by `malcolm.rowe` on 27 Oct 2012 at 9:44

jskeet:
This behaviour is centralized in NodaFormatInfo.GetInstance, but is
visible via public methods such as ToString and when constructing
patterns.

New handling:

- A null argument means "use the current culture"
- A CultureInfo argument will be used for resource lookups, text
  comparisions, and for date/time strings.
- A DateTimeInfo argument will be used for date/time strings, with
  the invariant culture used for resource lookups and text
  comparisons.
- Any other value will be rejected (with an ArgumentException)

Fixes #134.


src/NodaTime/Globalization/NodaFormatInfo.cs
```diff
@@ -73,13 +78,30 @@ internal sealed class NodaFormatInfo
...
         /// <summary>
-        /// Initializes a new instance of the <see cref="NodaFormatInfo" /> class.
+        /// Initializes a new instance of the <see cref="NodaFormatInfo" /> class based solely
+        /// on a <see cref="System.Globalization.CultureInfo"/>.
         /// </summary>
-        /// <param name="cultureInfo">The culture info to base this on.</param>
+        /// <param name="cultureInfo">The culture info to use.</param>
+        [VisibleForTesting]
         internal NodaFormatInfo([NotNull] CultureInfo cultureInfo)
+            : this(cultureInfo, cultureInfo?.DateTimeFormat)
```
malcolmr: This seems odd: are you deliberately allowing a `[NotNull]` parameter to be `null` here so that you can delegate to another constructor with `[NotNull]` parameters (which can verify null-ness) rather than fail here?

src/NodaTime/Globalization/NodaFormatInfo.cs
```diff
@@ -388,22 +417,34 @@ internal static NodaFormatInfo GetFormatInfo([NotNull] CultureInfo cultureInfo)
...
-            if (provider != null)
+            // TODO: Use C# 7 pattern matching for this...
+            if (provider == null)
             {
-                var cultureInfo = provider as CultureInfo;
-                if (cultureInfo != null)
-                {
-                    return GetFormatInfo(cultureInfo);
-                }
+                return GetInstance(CultureInfo.CurrentCulture);
```
malcolmr: Not simply `provider = CultureInfo.CurrentCulture;` and allow the function to continue?

(On the other hand, I do quite like the delegation, so you could keep it too.)

malcolmr: I assume there's no way `CurrentInfo.CurrentCulture` can be `null`?

jskeet: I certainly hope not :)

jskeet: Could do that, yes. Would probably be a little clearer - will change.

src/NodaTime/Globalization/NodaFormatInfo.cs
malcolmr: I think as-is is fine provided that it works as you want.

PR #625, opened 2017-01-01
Testing extensions

jskeet:
We might want more at some point, but these are probably the most important ones.

(I'm going to investigate why we need the different overloads. It may be a C# compiler bug, but we won't be able to fix that immediately...)


jskeet:
Ah, this is why they fail... a requirement in extension methods is:

> An implicit identity, reference or boxing conversion exists from `expr` to the type of the first parameter of `Mj`.

src/NodaTime.Testing/Extensions/LocalDateConstruction.cs
```diff
@@ -0,0 +1,168 @@
+﻿// Copyright 2016 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+namespace NodaTime.Testing.Extensions
+{
+    /// <summary>
+    /// Extension methods for constructing <see cref="LocalDate"/> values.
```
malcolmr: Do we want to mention the ISO calendar anywhere here?

PR #626, opened 2017-01-10
Handle truncation and range issues for BCL conversions.

### LINKED ISSUE #395: Instant.ToDateTimeOffset Method Throws ArgumentException When Offset Contains Seconds
Instant.ToDateTimeOffset method will throw a SystemArgumentException "Offset must be specified in whole minutes" from the DateTimeOffset class when the Instant contains an offset with seconds (ie. -04:19:18). See the [MSDN TimeSpan ArgumentException](https://msdn.microsoft.com/en-us/library/bb351980%28v=vs.110%29.aspx) for more details.

This issue can occur very frequently when a system deals with dates prior to the late 1890's. It was around this time in history that various locales standardized their time zone. For the period before this time the offset looks to be more of a calculated value which typically contains seconds.

**Examples of these Offsets**

| Time Zone | Offset | Until |
| --- | --- | --- |
| Atlantic/Bermuda | -04:19:18 | 1930 Jan 1 02:00 |
| America/Mazatlan | -07:05:40 | 1921 Dec 31 23:54:20 |
| Pacific/Honolulu | -10:31:26 | 1896 Jan 13 12:00 |

@jskeet post here contains hundreds of examples of the points in history when these offsets containing seconds occur:
https://github.com/nodatime/nodatime/blob/master/data/tzdb/2015a/northamerica

**Example To Reproduce The Exception**

``` C#
DateTime dt = DateTime.Parse("1896-01-12 00:00:00.000");
DateTimeZone tz = DateTimeZoneProviders.Tzdb["Pacific/Honolulu"];
LocalDateTime localDt = LocalDateTime.FromDateTime(dt);
ZoneLocalMappingResolver customResolver = Resolvers.CreateMappingResolver(Resolvers.ReturnLater, Resolvers.ReturnStartOfIntervalAfter);
ZonedDateTime zdt = localDt.InZone(tz, customResolver);  // dzt.Offset results in -10:31:26
DateTimeOffset dto = zdt.ToDateTimeOffset();  // <- Exception is thrown here.
```

**Work Around**
Create a new DateTimeOffset object using a TimeSpan object created from the Hours and Minutes of the Offset from the converted ZonedDateTime object and set the seconds value to zero.

**Work Around Example**

``` C#
...
ZonedDateTime zdt = localDt.InZone(tz, customResolver);  // dzt.Offset results in -10:31:26
TimeSpan tsTemp = zdt.Offset.ToTimeSpan();
TimeSpan ts = new TimeSpan(tsTemp.Hours, tsTemp.Minutes, 0); // Set seconds to zero.
DateTimeOffset dto = new DateTimeOffset(zdt.Year, zdt.Month, zdt.Day, zdt.Hour, zdt.Minute, zdt.Second, zdt.Millisecond, ts);
```

jskeet:
- Document and test that we truncate towards the start of time
- Throw an appropriate exception for date/time and offset values
  being out of range
- Round offsets towards zero if they have non-zero second values

Fixes #395.

### DIFF
--- a/src/NodaTime/LocalDateTime.cs
+++ b/src/NodaTime/LocalDateTime.cs
@@ -290,18 +290,29 @@ namespace NodaTime
         /// of <see cref="DateTimeKind.Unspecified"/>.
         /// </summary>
         /// <remarks>
+        /// <para>
         /// <see cref="DateTimeKind.Unspecified"/> is slightly odd - it can be treated as UTC if you use <see cref="DateTime.ToLocalTime"/>
         /// or as system local time if you use <see cref="DateTime.ToUniversalTime"/>, but it's the only kind which allows
         /// you to construct a <see cref="DateTimeOffset"/> with an arbitrary offset, which makes it as close to
         /// the Noda Time non-system-specific "local" concept as exists in .NET.
+        /// </para>
+        /// <para>
+        /// If the date and time is not on a tick boundary (the unit of granularity of DateTime) the value will be truncated
+        /// towards the start of time.
+        /// </para>
         /// </remarks>
+        /// <exception cref="InvalidOperationException">The date/time is outside the range of <c>DateTime</c>.</exception>
         /// <returns>A <see cref="DateTime"/> value for the same date and time as this value.</returns>
         [Pure]
-        public DateTime ToDateTimeUnspecified() =>
-            new DateTime(
-                TickArithmetic.DaysAndTickOfDayToTicks(date.DaysSinceEpoch, time.TickOfDay) + NodaConstants.BclTicksAtUnixEpoch,
-                DateTimeKind.Unspecified);
-            
+        public DateTime ToDateTimeUnspecified()
+        {
+            long ticks = TickArithmetic.DaysAndTickOfDayToTicks(date.DaysSinceEpoch, time.TickOfDay) + NodaConstants.BclTicksAtUnixEpoch;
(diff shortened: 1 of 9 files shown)


src/NodaTime/LocalDateTime.cs
```diff
@@ -290,18 +290,29 @@ internal LocalDateTime(LocalDate date, LocalTime time)
...
-            new DateTime(
-                TickArithmetic.DaysAndTickOfDayToTicks(date.DaysSinceEpoch, time.TickOfDay) + NodaConstants.BclTicksAtUnixEpoch,
-                DateTimeKind.Unspecified);
-            
+        public DateTime ToDateTimeUnspecified()
+        {
+            long ticks = TickArithmetic.DaysAndTickOfDayToTicks(date.DaysSinceEpoch, time.TickOfDay) + NodaConstants.BclTicksAtUnixEpoch;
+            if (ticks < 0)
+            {
+                throw new InvalidOperationException("LocalDateTIme was out of range of DateTime");
```
malcolmr: typo: `LocalDateTime`. Also, elsewhere the message is "NodaTimeType out of range for BclType" (no "was").

jskeet: Thanks, will fix both. Admittedly in some cases we may see the "wrong" type in the exception, as we'll be going via this method - but I think it'll be descriptive enough in all cases.

PR #633, opened 2017-01-19
NodaTime.targets condition fix for Visual Studio

### LINKED ISSUE #632: Unresolved references showing up in Visual Studio 2015
If I include the NodaTime v2.0.0-alpha20160729 NuGet package into a .NET Standard 1.3
portable class library, Visual Studio displays three warnings after closing and reopening the project:

> The referenced component 'NodaTime' could not be found.			
> The referenced component 'System.Xml' could not be found.			
> The referenced component 'System.Numerics' could not be found.			

image

[Here](https://github.com/epsitec/rnd-nodatime-ref) is a minimal [example project](https://github.com/epsitec/rnd-nodatime-ref) which demonstrates the issue.

Has this something to do with the tooling, or is there something wrong with the nuget package itself?

emgarten:
The ImportGroup condition in NodaTime.targets is not evaluated correctly in Visual Studio 2015. This leads to unresolved reference warnings for C# netstandard projects. The fix for this is to use a Choose/When instead of the ImportGroup condition.

Fixes #632

### DIFF
--- a/packaging/NodaTime.targets
+++ b/packaging/NodaTime.targets
@@ -1,15 +1,19 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- for C++/CLI projects only -->
-  <ItemGroup Condition="'$(Language)' == 'C++'">
-    <Reference Include="NodaTime">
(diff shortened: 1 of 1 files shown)


jskeet:
Thanks very much for this - I've personally no idea whether the fix works, but I'm happy to trust you on it :)

PR #651, opened 2017-02-19
A few optimizations

### LINKED ISSUE #567: Performance considerations
Why does NodaTime 2 use `(nanoOfDay, day)` as the data model? Wouldn't it be nicer to use `(overflow, unixTimestampNanos)`?

My concern is basically, that NodaTime 2 introduces a huge overhead on converting a 64 bit integer to an `Instant` object, just because of this convention (compared to NodaTime 1, where instead of multiple arithmetic operations it was a straight assignment).

Nanosecond precision in a 64bit unix timestamp is enough to express the interval `[1824, 2115]`, so in that whole time you could ignore the `overflow` part completely.

jskeet:
These are the optimizations I'd been considering earlier. I've tested them, and the differences are not huge, but they're not insignificant either.

Fixes #567.

### DIFF
--- a/src/NodaTime.Test/NodaConstantsTest.cs
+++ b/src/NodaTime.Test/NodaConstantsTest.cs
@@ -3,6 +3,7 @@
 // as found in the LICENSE.txt file.
 
 using NUnit.Framework;
+using System;
 
 namespace NodaTime.Test
 {
@@ -16,6 +17,22 @@ namespace NodaTime.Test
             var localEpoch = new LocalDateTime(-4712, 1, 1, 12, 0, CalendarSystem.Julian);
             var epoch = localEpoch.InZoneStrictly(DateTimeZone.Utc).ToInstant();
             Assert.AreEqual(epoch, NodaConstants.JulianEpoch);
+        }
+
+        [Test]
+        public void BclTicksAtEpoch()
+        {
+            Assert.AreEqual(
+                new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Ticks,
+                NodaConstants.BclTicksAtUnixEpoch);
+        }
+
+        [Test]
+        public void BclDaysAtEpoch()
+        {
+            Assert.AreEqual(
+                new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Ticks,
+                NodaConstants.TicksPerDay * NodaConstants.BclDaysAtUnixEpoch);
         }
     }
 }
--- a/src/NodaTime/Utility/TickArithmetic.cs
+++ b/src/NodaTime/Utility/TickArithmetic.cs
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+using NodaTime.Annotations;
 using static NodaTime.NodaConstants;
 
 namespace NodaTime.Utility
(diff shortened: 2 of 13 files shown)


src/NodaTime/Utility/TickArithmetic.cs
```diff
@@ -11,48 +12,90 @@ namespace NodaTime.Utility
...
-            // very significantly faster under the x64 JIT (and no slower under the x86 JIT).
-            // See http://stackoverflow.com/questions/22258070 for the inspiration.
-            return ticks >= 0 ? unchecked((int)((ticks >> 14) / 52734375L))
-                    // TODO: Optimize with shifting at some point. Note that this must *not* subtract from ticks,
-                    // as it could already be long.MinValue.
-                    : (int)((ticks + 1) / TicksPerDay) - 1;
+            unchecked
+            {
+                Preconditions.DebugCheckArgument(ticks >= 0, nameof(ticks), "Ticks must be non-negative");
+                int days = unchecked((int) ((ticks >> 14) / 52734375L));
```
malcolmr: (ditto)

src/NodaTime/Utility/TickArithmetic.cs
```diff
@@ -11,48 +12,90 @@ namespace NodaTime.Utility
...
+            unchecked
+            {
+                // First work out the number of days, always rounding down (so that ticks * TicksPerDay is always the
+                // start of the day).
+                // The shift approach here is equivalent to dividing by NodaConstants.TicksPerDay, but appears to be
+                // very significantly faster under the x64 JIT (and no slower under the x86 JIT).
+                // See http://stackoverflow.com/questions/22258070 for the inspiration.
+                if (ticks >= 0)
+                {
+                    int days = unchecked((int)((ticks >> 14) / 52734375L));
```
malcolmr: Is it useful to nest `unchecked()` within an `unchecked {}` block?

src/NodaTime/Utility/TickArithmetic.cs
```diff
@@ -11,48 +12,90 @@ namespace NodaTime.Utility
...
 
-        // TODO(2.0): Check whether this is still required. I don't believe we'll ever see
-        // a value with such a small number of ticks now, due to the decreased date range in 2.0.
-        internal static long DaysAndTickOfDayToTicks(int days, long tickOfDay) =>
+
+        /// <summary>
+        /// Cautiously computes a number of ticks from day/tick-of-day value. This may overflow,
+        /// but will only do so if it has to.
+        /// </summary>
+        internal static long DaysAndTickOfDayToTicksChecked(int days, long tickOfDay) =>
```
malcolmr: Rather than ...Checked and ...Unchecked, can you name the "Unchecked" variant to indicate the precondition that must hold, and the drop the ...Checked suffix from the other?

(basically, make these two follow the same pattern as (NonNegative)TicksToDaysAndTickOfDay() above.)

src/NodaTime.Test/NodaConstantsTest.cs
```diff
@@ -17,5 +18,21 @@ public void JulianEpoch()
...
+                new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Ticks,
+                NodaConstants.BclTicksAtUnixEpoch);
+        }
+
+        [Test]
+        public void BclDaysAtEpoch()
+        {
+            Assert.AreEqual(
+                (int) (new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Ticks / NodaConstants.TicksPerDay),
+                NodaConstants.BclDaysAtUnixEpoch);
```
malcolmr: Could we test by multiplying rather than a truncating division? I know it's not testing directly, but I suspect this would still pass if `DateTime.Ticks` weren't exactly divisible by `TicksPerDay`.

src/NodaTime/Utility/TickArithmetic.cs
jskeet: Not at all - will fix, thanks :)

src/NodaTime/Utility/TickArithmetic.cs
jskeet: Yes, SGTM.

jskeet: The name is somewhat trickier here - I've called it "Bounded", but other suggestions welcome.

malcolmr: Bounded sounds reasonable. I was originally thrown because the 'unsafe' version of `TicksToDaysAndTickOfDay()` still used `unchecked`...

jskeet:
Changes made and squashed - will merge when green.

PR #657, opened 2017-02-20
Add a delegating converter for JSON

### LINKED ISSUE #656: Unseal NodaPatternConverter
While most of the time we wouldn't expect to derive from it, when specifying a particular converter for a property using `[JsonConverter]`, you have to identify a converter by type... which is much more easily done if you can just subclass `NodaPatternConverter`.

jskeet:
Fixes #656 (but without unsealing NodaPatternConverter).

See http://stackoverflow.com/questions/42351272 for more context.


jskeet:
I think I'll suggest a change for Json.NET, but I very much doubt that we'll be able to get it "in" before we want to release 2.0.

jskeet:
Merging this now, and can remove it if the Json.NET issue gets traction in time :)

PR #658, opened 2017-02-21
Fix a lot of TODOs and FIXMEs

jskeet:
Quite a few to go, but this is a good start.
I'd suggest reviewing one commit at a time - most commits are trivial.

### DIFF
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -736,7 +736,8 @@ namespace NodaTime
         /// Days property, and all time properties are normalized to their natural range.
         /// Subsecond values are normalized to millisecond and "nanosecond within millisecond" values.
         /// So for example, a period of 25 hours becomes a period of 1 day
-        /// and 1 hour. Aside from months and years, either all the properties
+        /// and 1 hour. A period of 1,500,750,000 nanoseconds becomes 1 second, 500 milliseconds and
+        /// 750,000 nanoseconds. Aside from months and years, either all the properties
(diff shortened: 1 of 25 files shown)


src/NodaTime/Period.cs
```diff
@@ -725,33 +726,30 @@ public Duration ToDuration()
...
         /// </summary>
         /// <remarks>
         /// Months and years are unchanged
         /// (as they can vary in length), but weeks are multiplied by 7 and added to the
-        /// Days property, and all time properties are normalized to their natural range
-        /// (where ticks are "within a millisecond"), adding to the larger property where
-        /// necessary. So for example, a period of 25 hours becomes a period of 1 day
+        /// Days property, and all time properties are normalized to their natural range.
+        /// Subsecond values are normalized to millisecond and "nanosecond within millisecond" values.
+        /// So for example, a period of 25 hours becomes a period of 1 day
```
malcolmr: Possibly add a second example: "1 day and 1 hour, while a period of 1500750μs  becomes a period of 1 second, 500 milliseconds, and 750,000 nanoseconds-with-millisecond" (assuming my maths is correct).

jskeet: Could do - although the fact that we don't support microseconds as a unit makes it a little odd. (In order to test such an assertion, you'd probably create a period of 1500750000 nanoseconds.)

jskeet:
Rebased, added extra example - will merge when green.

Issue #607, opened 2016-12-21
LocalTime text parsing doesn't support ISO 8601 formats


lundmikkel:
According to the [ISO 8601 standard regarding times](https://en.wikipedia.org/wiki/ISO_8601#Times):

> either the seconds, or the minutes and seconds, may be omitted from the basic or extended time formats for greater brevity but decreased accuracy: [hh]:[mm], [hh][mm] and [hh] are the resulting reduced accuracy time formats.

These formats are not supported by the `LocalTimePattern.ExtendedIsoPattern`, nor does it seem to be possible at all using a normal pattern text. Furthermore, you cannot even create your own `IPattern<LocalTime>` because the return type `ParseResult<T>` is sealed and only has private constructors (unless you make the pattern a wrapper around the three patterns: `"HH"`, `"HH':'mm"`, and `"HH':'mm':'ss;FFFFFFF"`).

Only including the necessary digits is already supported, but only with fractional seconds. It doesn't seem to be possible with minutes nor seconds.

jskeet:
Right - at the moment this would be a bit of a pain.

We should be able to support it without too much hassle though - we already have the internal mechanism for a composite pattern, where on parsing we try to parse one pattern then another. That would handle parsing, but wouldn't do what you _might_ want for formatting, where you format to the shortest form that doesn't lose information. That will require a bit more code. Not hard to do, but not supported yet. I don't think this should block the 2.0 release - I think I'd expose that as a separate pattern - but it's definitely a useful thing to implement at some point... thanks for raising it.

lundmikkel:
It would be nice to be able to specify it in the pattern. I'm in a situation, where I would like to require minutes, but not seconds. A convention with capital letters as known from fractional seconds could maybe make sense, i.e. `"HH':'mm':'SS;FFFFFFF"`. Not sure if that conflicts with anything else though.

Slightly off topic: it seems that the custom pattern documentation for LocalTime's hours (`h` and `hh`) has a wrong example, which uses capital letters.

jskeet:
Raised #608 for the docs issue. I don't *think* I want to introduce whole extra format specifiers for this, but I may just make sure there are all the patterns you might want...

jskeet:
(This shouldn't have been closed yet, as the type is still internal.)

PR #660, opened 2017-02-23
First pass at a public composite pattern

jskeet:
This is designed to provide a way of fixing #607.
I'm not entirely convinced about the design though... the "parse
using patterns in one direction, format using patterns in the other
direction" is useful for us internally (because we need partial
patterns), but not obvious.

Putting out there for comment... I'll add more tests once we've got
the design sorted.


jskeet:
cc @lundmikkel and @malcolmr - this PR isn't ready for submission, but is ready for design consideration.

src/NodaTime/Utility/Preconditions.cs
```diff
@@ -112,11 +112,13 @@ internal static void DebugCheckArgumentRange([InvokerParameterName] string param
         [Conditional("DEBUG")]
         internal static void DebugCheckArgument(bool expression, [InvokerParameterName] string parameter, string messageFormat, params object[] messageArgs)
         {
+#if DEBUG
```
malcolmr: Changes in this file seem useful but unrelated?

jskeet: I needed to introduce DebugCheckState, and that led me to find the inconsistent exception use. It may well be worth me pulling out these changes into a separate PR though...

src/NodaTime/Text/CompositePatternBuilder.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// A composite pattern is a combination of multiple patterns. When parsing, these are checked
+    /// in the order in which they are added to the builder with the <see cref="Add(IPattern{T}, Func{T, bool})"/>
+    /// method, by trying to parse and seeing if the result is a successful one. When formatting,
+    /// the patterns are checked in the reverse order, using the predicate provided along with the pattern
+    /// when calling <c>Add</c>. The intention is that patterns are added in "most precise first" order,
+    /// and the predicate should indicate whether it can fully represent the given value - so the "less precise"
+    /// (and therefore usually shorter) pattern can be used first.
+    /// </remarks>
+    /// <typeparam name="T">The type of value to be parsed or formatted by the resulting pattern.</typeparam>
+    internal sealed class CompositePatternBuilder<T> : IEnumerable<IPattern<T>>
```
jskeet: This was meant to be public... but if we submit while it's internal, we can still reflect on the API a bit further.

I'll add some more tests, fix whitespace-only changes and then merge.

PR #662, opened 2017-02-24
Update NUnit packages

jskeet:
This is a bit verbose when running tests by default.

(Still looking into that...)


jskeet:
Closing this PR for now - see https://github.com/nunit/dotnet-test-nunit/issues/78

PR #665, opened 2017-02-26
Improve binary serialization

jskeet:
First commit is mostly refactoring, although it does change the serialization format a little. (In particular, it changes the Offset representation to only use seconds, losing compatibility with Noda Time 1.x, but all the other types lost this compatibility a long time ago anyway...)

Second commit adds some validation we'd previously noted was missing.


src/NodaTime/LocalTime.cs
```diff
@@ -742,17 +742,25 @@ void IXmlSerializable.WriteXml([NotNull] XmlWriter writer)
...
         /// <param name="info">The <see cref="SerializationInfo"/> to fetch data from.</param>
         /// <param name="context">The source for this deserialization.</param>
         private LocalTime([NotNull] SerializationInfo info, StreamingContext context)
+            : this(info)
+        {
+        }
+
+        /// <summary>
+        /// Constructor only present for serialization; internal to allow
+        /// construction from other
```
malcolmr: incomplete sentence?

jskeet: Whoops, will fix, thanks.

PR #667, opened 2017-02-27
Website: fix link to roadmap

jorgebay:


jskeet:
Thanks! Merging now...

jorgebay:
ok, that was fast! 😃

jskeet:
Well, that was just the PR... I've now regenerated the site, and it's pushing at the moment, so the change should be visible soon. I've also added redirection so that `roadmap.html` would have worked (but non-optimally) so that old links are still valid.

PR #671, opened 2017-03-05
Rename DateInterval.Inclusive to EndInclusive

jskeet:
This improves clarity - the property *could* still be removed
if #2 is resolved to make all date intervals inclusive.

### DIFF
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -43,12 +43,12 @@ namespace NodaTime.Test
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


malcolmr:
LGTM. (I must admit that I am generally supportive of baking whether an interval is open or half-open into the type itself, rather than being a property of the instance, but as I recall you had use cases that supported both forms.)

jskeet:
Will look into the AppVeyor change separately - it's not related to this change.

jskeet:
Will add a comment on #2 about the "different types" option. I have an idea.

PR #673, opened 2017-03-06
Increase the number of bits available to months

jskeet:
We now support months in the range [1, 32]. This is
necessary for the Wondrous calendar which has 19 months.

I've taken the extra bit from the calendar range, so we can now "only"
support 64 calendars... but we've only got 18 at the moment, and the
Wondrous calendar will be the 19th. I think that's safe.

This needs to be done before 2.0 as it affects the serialization
format.

// cc @glittle

### DIFF
--- a/src/NodaTime/YearMonthDay.cs
+++ b/src/NodaTime/YearMonthDay.cs
@@ -12,10 +12,11 @@ namespace NodaTime
     /// </summary>
     /// <remarks>
     /// <para>
-    /// The day is represented in bits 0-5. The month is represented in bits 6-9. The
-    /// year is represented in bits 10-24. This type is naive: comparisons are performed
-    /// assuming that a larger month number always comes after a smaller month number,
-    /// etc. This is suitable for most, but not all, calendar systems.
+    /// See <see cref="YearMonthDayCalendar"/> for the number of bits per component,
+    /// but this doesn't have the calendar component, so bit 0 is part of the day value.
+    /// This type is naive: comparisons are performed assuming that a larger month number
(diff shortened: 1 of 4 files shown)


src/NodaTime/YearMonthDay.cs
```diff
@@ -12,10 +12,11 @@ namespace NodaTime
     /// </summary>
     /// <remarks>
     /// <para>
-    /// The day is represented in bits 0-5. The month is represented in bits 6-9. The
-    /// year is represented in bits 10-24. This type is naive: comparisons are performed
-    /// assuming that a larger month number always comes after a smaller month number,
-    /// etc. This is suitable for most, but not all, calendar systems.
+    /// See <see cref="YearMonthDayCalendar"/> for the number of bits per component,
+    /// but this type shifts all information over (so bit 0 is part of the day value).
```
malcolmr: perhaps something like "but this type excludes the calendar field (so that bit 0..."

jskeet: Yup, SGTM - will make that change.

jskeet:
We certainly *could* remove this from the binary serialization side of things - but it would increase the serialized size. (The compactness here is reflected in the serialized compactness.)

The alternative would probably be to serialize this `YearMonthDayCalendar` as 4 separate fields. It's a bit more fiddly, but would indeed allow backward-compatible changes. I think I'd still want to rely on the calendar ordinals not changing though.

Will merge this PR as is, but raise another issue - then I can create a speculative PR to address it, which is easier to think about than the abstract idea.

PR #678, opened 2017-03-07
First pass at an "update TZDB" script

### LINKED ISSUE #360: NuGet package for latest TZDB

jskeet:
This only tackles 1.3. I think it's going to be easiest to
have one of these scripts per minor release, and then one
to call them all (moving some of this script there).

It doesn't rebuild any documentation, so the docs will still show
(e.g.) 1.3.2... but I think that's okay.

I've used this to build a 1.3.3, but I won't push that just yet - I
want to make the web site automatically detect the new release too,
before I do that.

Fixes #360.

### DIFF
--- a/build/updatehistorictzdb.sh
+++ b/build/updatehistorictzdb.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -e
+
+declare -r GSUTIL=gsutil.cmd
+declare -r ROOT=$(realpath $(dirname $0)/..)
+
+rm -rf tmp
+mkdir -p tmp/releases
+cd tmp/releases
+
+# Work out the current release, fetch and extract it
+declare -r RELEASE=$(\
+    $GSUTIL ls gs://nodatime/releases | \
+    grep -o -E 'NodaTime-1\.3\.[0-9]+\.zip' | \
+    sort -r | \
+    head -n 1 | \
+    sed s/NodaTime-// | \
+    sed s/.zip//)
+echo "Fetching and expanding release ${RELEASE}"
+$GSUTIL -q cp gs://nodatime/releases/NodaTime-${RELEASE}.zip .
+$GSUTIL -q cp gs://nodatime/releases/NodaTime-${RELEASE}-src.zip .
+curl -s -S https://api.nuget.org/packages/nodatime.${RELEASE}.nupkg -o nodatime.${RELEASE}.nupkg
+unzip -q NodaTime-${RELEASE}.zip
+unzip -q NodaTime-${RELEASE}-src.zip
+unzip -q nodatime.${RELEASE}.nupkg -d NodaTime-${RELEASE}-nupkg
+
+# Handy "increment version number" code from http://stackoverflow.com/questions/8653126
+declare -r NEW_RELEASE=`echo $RELEASE | perl -pe 's/^((\d+\.)*)(\d+)(.*)$/$1.($3+1).$4/e'`
+
+# Update all the files that need updating
+echo "Updating to reflect ${NEW_RELEASE}"
+mv NodaTime-${RELEASE} NodaTime-${NEW_RELEASE}
+mv NodaTime-${RELEASE}-src NodaTime-${NEW_RELEASE}-src
+cp $ROOT/src/NodaTime/TimeZones/Tzdb.nzd NodaTime-${NEW_RELEASE}-src/src/NodaTime/TimeZones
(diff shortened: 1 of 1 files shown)


jskeet:
(Malcolm: I think this just needs a brief sanity check... anything you can think of that I've obviously forgotten?)

build/updatehistorictzdb.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+set -e
+
+rm -rf tmp
```
malcolmr: If you're expecting this to be run from a particularly directory, perhaps add something like:

```sh
cd $(dirname $0)/..
```

before this.

Alternatively (or additionally) use something like

```sh
declare -r TEMPPATH=$(mktemp -d)
trap "rm -rf ${TEMPPATH}" EXIT
```

To give yourself a temporary directory to build into.

build/updatehistorictzdb.sh
```diff
@@ -0,0 +1,57 @@
...
+mkdir tmp
+
+# Get the list of files, just once.
+
+mkdir -p tmp/releases
+cd tmp/releases
+FILES=files.txt
```
malcolmr: Prefer to mark variables as readonly where appropriate.

If you're happy with bashisms, then

```bash
declare -r FILES=files.txt
```

otherwise

```sh
FILES=files.txt
readonly FILES
```

(likewise other constant-like variables below.)

build/updatehistorictzdb.sh
```diff
@@ -0,0 +1,57 @@
...
+
+gsutil.cmd ls gs://nodatime/releases > $FILES
```
malcolmr: `gsutil` is just `gsutil` on Linux. Do you need the `.cmd` suffix on Windows?

Either way, perhaps pull out a

```bash
declare -r GSUTIL=gsutil.cmd
```

at the top of the file, then use `${GSUTIL} ls ...` here, that way there's only one place to change.

build/updatehistorictzdb.sh
```diff
@@ -0,0 +1,57 @@
...
+RELEASE=`cat $FILES | grep -o -E 'NodaTime-1\.3\.[0-9]+\.zip' | sort -r | head -n 1 | sed s/NodaTime-// | sed s/.zip//`
+gsutil.cmd cp gs://nodatime/releases/NodaTime-${RELEASE}.zip .
+gsutil.cmd cp gs://nodatime/releases/NodaTime-${RELEASE}-src.zip .
+curl -s -S https://api.nuget.org/packages/nodatime.${RELEASE}.nupkg -o nodatime.${RELEASE}.nupkg
+unzip -q NodaTime-${RELEASE}.zip
+unzip -q NodaTime-${RELEASE}-src.zip
+unzip -q nodatime.${RELEASE}.nupkg -d NodaTime-${RELEASE}-nupkg
+
+# Handy "increment version number" code from http://stackoverflow.com/questions/8653126
+NEW_RELEASE=`echo $RELEASE | perl -pe 's/^((\d+\.)*)(\d+)(.*)$/$1.($3+1).$4/e'`
```
malcolmr: nit: prefer

```sh
$(foo)
```

over

```sh
`foo`
```

(it has saner nesting behaviour, if nothing else)

build/updatehistorictzdb.sh
jskeet: I'm expecting to run it from the directory in which it resides - so presumably `cd $(dirname $0)` with no `..`? Or I could just stick to knowing that it'll only be me running it, probably :)

build/updatehistorictzdb.sh
jskeet: Bashism is fine - will do that.

build/updatehistorictzdb.sh
jskeet: Will do.

build/updatehistorictzdb.sh
jskeet: Could do. At the moment it also relies on nuget and msbuild, neither of which I'm expecting to be callable "just like that" for a little while.

jskeet:
Urgh - I hadn't spotted that I'd lost some changes which *do* use `dirname`... will have to reimplement those.

jskeet:
(Oh, and to answer a comment I hadn't spotted - yes, you need the .cmd for gsutil and the .exe for msbuild. Not sure why.)

PR #684, opened 2017-03-07
Make DateInterval always end-inclusive.

### LINKED ISSUE #2: Support ranges
```
Several applications need the idea of a range - I suspect the most common 
case is local dates.

It would be nice if we could use:

    LocalDateRange range = new LocalDateRange(start, end);
    foreach (LocalDate date in range)
    {
    }

We can use the code in MiscUtil (http://pobox.com/~skeet/csharp/miscutil) as 
a starting point for this. We may want to consider whether to "go generic" or 
just include the ranges for the most common cases.
```

Original issue reported on code.google.com by `jonathan.skeet` on 4 Dec 2009 at 8:09

jskeet:
Fixes #2.

### DIFF
--- a/src/NodaTime/DateInterval.cs
+++ b/src/NodaTime/DateInterval.cs
@@ -18,16 +18,9 @@ namespace NodaTime
     /// The two dates must be in the same calendar, and the end date must not be earlier than the start date.
     /// </para>
     /// <para>
-    /// By default, the end date is deemed to be part of the range, as this matches many real life uses of
+    /// The end date is deemed to be part of the range, as this matches many real life uses of
     /// date ranges. For example, if someone says "I'm going to be on holiday from Monday to Friday," they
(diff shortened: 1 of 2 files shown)


src/NodaTime/DateInterval.cs
```diff
@@ -214,141 +137,30 @@ public bool Contains(LocalDate date)
...
             Preconditions.CheckArgument(date.Calendar.Equals(Start.Calendar), nameof(date),
                 "The date to check must be in the same calendar as the start and end dates");
-            return Start <= date && (EndInclusive ? date <= End : date < End);
+            return Start <= date && date <= End;
         }
 
         /// <summary>
         /// Gets the length of this date interval in days.
         /// </summary>
-        /// <remarks>
```
malcolmr: Could perhaps say that this will always be at least 1?

jskeet: I was wondering whether or not to do that - will do.

PR #691, opened 2017-03-17
Update build files to use csproj

### LINKED ISSUE #685: Transition to csproj files
Big blocker at the moment: inability to run nunit tests

Minor issues:

- Warnings around package version downgrades (may be fine)
- Unable to run benchmarks under .NET Core

jskeet:
This will require all developers to use Visual Studio 2017 (or latest Visual Studio code etc).

Fixes #685.

(Initially incomplete - trying to get Travis working, then will look at the other CI systems. Also need to check the various build scripts will still work.)


jskeet:
Docfx still doesn't work - we need https://github.com/dotnet/docfx/issues/1254 to be fixed, basically. At that point, I think we're ready. Would still *like* to do this before 2.0 if possible...

jskeet:
Okay, I think this is ready to go now. Malcolm, please let me know if anything doesn't make sense.

jskeet:
I'd entirely forgotten we even *had* a Makefile :)

Yes, we should either update or possibly just delete it. And building.md should probably be updated too. As you say, we can do both a bit later.

PR #634, opened 2017-01-20
update format info to use C# 7 patterns

BillWagner:
This is much clearer with C# 7 pattern matching expressions.

PR #698, opened 2017-03-22
Update format info to use C# 7 patterns

jskeet:
This is a locally-rebased version of #634.

cc @BillWagner


jskeet:
Will merge when green. Thanks Bill :)

PR #699, opened 2017-03-22
Web site fixes

jskeet:
(See commits for details)


jskeet:
Will merge when green, then redeploy to the web site when I can. (That may be in a couple of hours...)

PR #702, opened 2017-03-23
Explicitly request deterministic builds

### LINKED ISSUE #359: Explore whether we can produce reproducible binaries
Now that we're pushing the website to nodatime/nodatime.org, I've noticed that `bin/NodaTime.Web.dll` is changing on every commit.

This is because the binaries contain embedded timestamp (etc) data that changes on every build.

That's not great for nodatime.org, but it's not good for our releases either, since it means that we can't reproduce a binary that we've built in the past (even if we have the same version of the tools that we built it with).

I'd like to explore how hard it'd be to redact the timestamps, etc, to generate reproducible binaries.  There doesn't seem like there's a pre-existing solution for this, so we might have to see if we can do it by hand as a post-build step.

Glancing at the diffs for `bin/NodaTime.Web.dll`, I see, in file order:
1. `TimeDateStamp` in the PE header is set by the linker.
2. The assembly has a random GUID in the MVID field; see`Module.ModuleVersionId`. (Including a UUID is required by ECMA-335, but there's no really good reason that I can see for generating a version 4 (random) UUID rather than a hash-based UUID; other than probably it's harder to do).
3. The Debug PE data directory contains a timestamp of the debugging information, and then the (CodeView RSDS) debugging data contains a GUID.

It should be fairly easy patch the first two, and I think that suppressing generation of PDB files when building will fix the last (for NodaTime.Web; that's less reasonable for the other binaries).

jskeet:
This appears to be the default now anyway, but it's possible that
it's a weaker version, or not the default everywhere. No harm in
being explicit.

Fixes #359.

### DIFF
--- a/src/NodaTime.Benchmarks/NodaTime.Benchmarks.csproj
+++ b/src/NodaTime.Benchmarks/NodaTime.Benchmarks.csproj
@@ -10,6 +10,7 @@
     <PackageId>NodaTime.Benchmarks</PackageId>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">1.1.1</RuntimeFrameworkVersion>
+    <Deterministic>True</Deterministic>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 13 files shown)


jskeet:
(Can't yet tell why Travis is failing - logs aren't showing up. Unrelated to this, I'm sure, so merging...)

PR #705, opened 2017-03-24
New "build release" script

jskeet:
Seems to work for prerc1...


build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$1" = "" ]]
```
malcolmr: Prefer `if [[ $# -ne 1 ]]`.

(If you're using bash, `[[` has less sharp edges that `test` / `[`, and you may as well check the number of arguments directly.)

Also prefer to format these as:

```bash
if expr; then
  ...
fi
```

rather than place the `then` on the next  line.

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+then
+  echo "Usage: buildrelease.sh version"
```
malcolmr: nit: prefer single-quotes where no variable substitution is required.

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo "e.g. buildrelease.sh 2.0.0"
+  echo "or buildrelease.sh 2.0.0-rc1"
+  echo "It is expected that a git tag will already exist"
```
malcolmr: You could optionally use a here document, as:

```bash
if ...; then
   cat <<EOF
Usage: $0 VERSION
blah blah
blah
EOF
fi
```

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+  exit 1
+fi
+
+declare -r VERSION=$1
+declare -r SUFFIX=$(echo $VERSION | cut -s -d- -f2)
+declare -r BUILD_FLAG=$([ "$SUFFIX" != "" ] && echo "--version-suffix $SUFFIX")
```
malcolmr:
```bash
declare -r BUILD_FLAG=${SUFFIX:+--version-suffix ${SUFFIX}}
```

is a much simpler way to do that if you can assume bash.

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+declare -r OUTPUT=artifacts
+
+rm -rf releasebuild
```
malcolmr: Again, if you're going to assume a particular starting directory, please put something like

```sh
cd $(dirname $0)/..
```

near the top of the script.

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+dotnet pack --include-symbols --no-build -c Release $BUILD_FLAG src/NodaTime.Testing
+cp src/NodaTime/bin/Release/*.nupkg $OUTPUT
+cp src/NodaTime.Testing/bin/Release/*.nupkg $OUTPUT
+git archive $VERSION -o $OUTPUT/NodaTime-$VERSION-src.zip --prefix=NodaTime-$VERSION-src/
+
+mkdir NodaTime-$VERSION
+cp AUTHORS.txt NodaTime-$VERSION
+cp LICENSE.txt NodaTime-$VERSION
+cp NOTICE.txt NodaTime-$VERSION
+cp "NodaTime Release Public Key.snk" NodaTime-$VERSION
```
malcolmr: nit: prefer single-quotes when you don't need interpolation.

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+if [[ "$1" = "" ]]
+then
+  echo "Usage: buildrelease.sh version"
+  echo "e.g. buildrelease.sh 2.0.0"
+  echo "or buildrelease.sh 2.0.0-rc1"
+  echo "It is expected that a git tag will already exist"
+  exit 1
+fi
+
+declare -r VERSION=$1
```
malcolmr: (Things fall apart later if `${VERSION}` contains spaces, but I think it's reasonable to assume that won't be the case.)

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+declare -r SUFFIX=$(echo $VERSION | cut -s -d- -f2)
+declare -r BUILD_FLAG=$([ "$SUFFIX" != "" ] && echo "--version-suffix $SUFFIX")
+declare -r OUTPUT=artifacts
+
+rm -rf releasebuild
+git clone https://github.com/nodatime/nodatime.git releasebuild
+cd releasebuild
+git checkout $VERSION
```
malcolmr:
```sh
git checkout "${VERSION}"
```
just to make sure that we pass a single argument to `checkout`.

(This if the first use of `${VERSION}`, and so will hopefully fail if we've passed a version containing a space because it won't exist.)

build/buildrelease.sh
```diff
@@ -0,0 +1,64 @@
...
+cp src/NodaTime/bin/Release/netstandard1.3/NodaTime.xml NodaTime-$VERSION/Portable
+cp src/NodaTime.Testing/bin/Release/netstandard1.3/NodaTime.Testing.xml NodaTime-$VERSION/Portable
+
+# Assemblies
+cp src/NodaTime/bin/Release/net45/NodaTime.dll NodaTime-$VERSION
+cp src/NodaTime.Testing/bin/Release/net45/NodaTime.Testing.dll NodaTime-$VERSION
+cp src/NodaTime/bin/Release/netstandard1.3/NodaTime.dll NodaTime-$VERSION/Portable
+cp src/NodaTime.Testing/bin/Release/netstandard1.3/NodaTime.Testing.dll NodaTime-$VERSION/Portable
+
+zip -r -9 $OUTPUT/NodaTime-$VERSION.zip NodaTime-$VERSION
```
malcolmr: I think we'd like the zipfile to be reproducible, no?

I believe this should work (cribbed largely from the wiki page linked below):

```sh
declare -r BUILD_DATE=$(git show -s --format=%cI)

# see https://wiki.debian.org/ReproducibleBuilds/TimestampsInZip.
find NodaTime-$VERSION -print0 | \
  xargs -0r touch --no-dereference --date="${BUILD_DATE}"
TZ=UTC zip -r -9 --latest-time --no-extra $OUTPUT/NodaTime-$VERSION.zip NodaTime-$VERSION
```

build/buildrelease.sh
jskeet: Done.

jskeet: Done. It's good to gradually get to grips with how to write bash scripts :)

build/buildrelease.sh
jskeet: Done.

build/buildrelease.sh
jskeet: Ack, but will keep it like this for now.

build/buildrelease.sh
jskeet: Definitely assuming bash. Will have to find a resource to understand how that works...

build/buildrelease.sh
jskeet: Done - will try calling it from the root directory as a test...

build/buildrelease.sh
jskeet: Done.

build/buildrelease.sh
jskeet: Done.

build/buildrelease.sh
jskeet: Ooh, nice. Will give it a try, building prerc1 twice in a row...

PR #706, opened 2017-03-24
Remove serialization

jskeet:
The buildapidocs bit is a bit fiddly, and I suspect it'll get *more* fiddly before we're done, but it's probably as good as it's going to get *right now*.


jskeet:
Ooh, thought I'd removed the jenkins file already. I meant to. Will do appveyor and the Makefile, although the makefile is massively out of date anyway :)

Yes, the Sandcastle removals are incidental - I just happened to spot them, and didn't want to bother with another PR.

Will add these extra changes as another commit.

jskeet:
Ah - I have done appveyor - the yml file just delegates to build/appveyor.sh, which has changed.

PR #716, opened 2017-03-27
Add a new docs directory for 2.0.x

jskeet:


src/NodaTime.Web/Markdown/2.0.x/versions.md
```diff
@@ -0,0 +1,4 @@
+@Title="Version history"
```
malcolmr: Possibly don't need this one?

jskeet: I was wondering about that. I thought it might be worth keeping for consistency, but equally I'd be happy to be convinced otherwise. Now is the time to do it, of course.

PR #744, opened 2017-04-04
"Handle" Feb 29th rules more cleanly

### LINKED ISSUE #743: Malta BCL zone issue on Linux/Mono
From #97.

Test NodaTime.Test.TimeZones.BclDateTimeZoneSourceTest.ProviderReturnsBclDateTimeZoneForAllAdvertisedIds fails with this exception

```text
System.InvalidOperationException : Requested transition for a ZoneYearOffset of February 29th in a non-leap year, not moving backwards to find a day-of-week
  at NodaTime.TimeZones.ZoneYearOffset.GetOccurrenceForYear (System.Int32 year) [0x0006d] in <c90fbbee5fa9426392c49904a3d9b7d0>:0 
```

The problem is in Europe/Malta, and I suspect it's due to this BCL adjustment rule:

01/01/1948 00:00:00 - 31/12/1948 00:00:00:
 Transition start: Day: 29 / DayOfWeek: Sunday / Fixed: True / Month: 2 / Week: 1 / TimeOfDay: 01/01/0001 02:00:00
 Transition end: Day: 3 / DayOfWeek: Sunday / Fixed: True / Month: 10 / Week: 1 / TimeOfDay: 01/01/0001 03:00:00

We're asking for the transition in 1949 for some reason, despite the rule only applying until 1948. That's odd to say the least.

jskeet:
Previously, we were following zic: if a rule says "exactly Feb 29th"
or "at least Feb 29th" it was invalid in a non-leap year. This is
painful due to the way we ask rules for transitions; it's simpler to
ignore the rule, and just treat Feb 29th as Feb 28th in non-leap
years.

It shouldn't make any different to validly-defined rules in TZDB,
although it does make us more lenient with badly-defined fules.
Importantly, it allows us to handle some valid BCL rules.

Fixes #743.

### DIFF
--- a/src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
+++ b/src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
@@ -142,6 +142,55 @@ namespace NodaTime.Test.TimeZones
             Assert.AreEqual(Offset.Zero, nodaZone.GetUtcOffset(Instant.FromUtc(-100, 10, 1, 0, 0)));
         }
 
+        [Test]
+        public void AwkwardLeapYears()
+        {
+            // This mimics the data on Mono on Linux for Europe/Malta, where there's a BCL adjustment rule for
+            // each rule for quite a long time. One of those years is 1948, and the daylight transition is Feburary
+            // 29th. That then fails when we try to build a ZoneInterval at the end of that year.
+            // See https://github.com/nodatime/nodatime/issues/743 for more details. We've simplified this to just
+            // a single rule here...
+
+            // Amusingly, trying to reproduce the test on Mono with a custom time zone causes Mono to throw -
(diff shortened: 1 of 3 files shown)


jskeet:
(Before I merge this, I'll want to pull it on my Linux box and rerun all the tests on Mono...)

jskeet:
Yeah, I'm not thrilled about it either. It just seems better than the alternatives.

I'll wait until Travis is sorted (see https://github.com/travis-ci/travis-ci/issues/7565) and I've tested on Linux before merging. We then need to decide whether this is worth a 2.0.1. (I haven't checked whether 1.3.x is broken in the same way. A lot of code was different then.)

jskeet:
Regenerating tzdb2017b gives the same results as before, which is encouraging :)

jskeet:
Humbug. Code now works on Mono, but the test makes Mono's time zone support break... should be feasible to add a couple more rules to fix it.

PR #745, opened 2017-04-05
Run more tests on Mono

jskeet:
Remove filtering of tests which used to be excluded on Mono due to Mono bugs, but where it now works (with 4.8.0 at least).


jskeet:
Just checking that it still works on Mono on Windows too, then will merge. Thanks for the quick review :)

Issue #746, opened 2017-04-05
"0 delta" daylight periods in TimeZoneInfo are annoying


jskeet:
Consider these BCL adjustment rules for America/Louisville, as seen on Linux under Mono:

```text
[1973-01-01 - 1973-1231]; Delta: 01:00:00
  DST start: Month: 4; Week: 1; Day: 29; DoW: Sunday; Time: 02:00; Fixed? True
  DST end: Month: 10; Week: 1; Day: 28; DoW: Sunday; Time: 02:00; Fixed? True
[1974-01-01 - 1974-1231]; Delta: 00:00:00
  DST start: Month: 1; Week: 1; Day: 6; DoW: Sunday; Time: 02:00; Fixed? True
  DST end: Month: 10; Week: 1; Day: 27; DoW: Sunday; Time: 02:00; Fixed? True
[1975-01-01 - 1975-1231]; Delta: 01:00:00
  DST start: Month: 2; Week: 1; Day: 23; DoW: Sunday; Time: 02:00; Fixed? True
  DST end: Month: 10; Week: 1; Day: 26; DoW: Sunday; Time: 02:00; Fixed? True
```

Note the "Delta: 00:00:00". That means "use the standard offset for the time zone, but let's pretend it's DST." Noda Time will get the right offsets, but end up saying that it's not on DST. I believe this is due to TimeZoneInfo not having a (public) representation of "standard time changes by rule".

Possible solution: if there's a delta of 0, pretend that the standard offset (for this ZoneInterval) is 1 hour off, and that there's a 1 hour delta.

jskeet:
Looking, this may be a problem for America/Resolute as well. Basically, our IsDaylight property perhaps incorrectly uses "0 savings" as an indication of DST. Needs more research - perhaps `ZoneInterval` needs an explicit `bool` in it. (And then we need to think about the nzd format. Sigh.)

jskeet:
@mj1856 @malcolmr: Any thoughts on this?

I think we either have to say "IsDaylight isn't the same as Savings==0" and keep a separate bool in ZoneInterval, or we have to effectively invent data within `BclDateTimeZone`, e.g. if we have time zone standard = +3 and we see a savings of 0, we say that's equivalent to a standard of +2 and a savings of +1. I've a horrible feeling we couldn't do that within the ZoneRecurrence, as we need both recurrences for a StandardDaylightAlternatingMap to have the same standard offset, but we could fake it up somewhere.

[...]

PR #750, opened 2017-04-05
Fake savings of 1 hour when TimeZoneInfo says the saving is 0 hours

jskeet:
This would "fix" #746, providing what's likely to be more accurate
data anyway. On the other hand, it's making up data, which always
makes me nervous.

I've tested this on Mono on Linux (the only environment I expect to see it at the moment), and it reduces the failure count from 36 to 26 - it looks like it knocks out all those problems (leaving other ones which I think are really Mono issues).

### DIFF
--- a/src/NodaTime/TimeZones/BclDateTimeZone.cs
+++ b/src/NodaTime/TimeZones/BclDateTimeZone.cs
@@ -193,8 +193,44 @@ namespace NodaTime.TimeZones
                 {
                     var daylightRecurrence = new ZoneRecurrence(zone.DaylightName, Savings, ConvertTransition(rule.DaylightTransitionStart), int.MinValue, int.MaxValue);
                     var standardRecurrence = new ZoneRecurrence(zone.StandardName, Offset.Zero, ConvertTransition(rule.DaylightTransitionEnd), int.MinValue, int.MaxValue);
-                    var recurringMap = new StandardDaylightAlternatingMap(StandardOffset, standardRecurrence, daylightRecurrence);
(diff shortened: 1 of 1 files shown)


jskeet:
(The fix ended up being pretty isolated, which makes me more confident in it than I would be otherwise.)

src/NodaTime/TimeZones/BclDateTimeZone.cs
```diff
@@ -193,12 +193,44 @@ internal BclAdjustmentRule(TimeZoneInfo zone, TimeZoneInfo.AdjustmentRule rule)
                 {
                     var daylightRecurrence = new ZoneRecurrence(zone.DaylightName, Savings, ConvertTransition(rule.DaylightTransitionStart), int.MinValue, int.MaxValue);
                     var standardRecurrence = new ZoneRecurrence(zone.StandardName, Offset.Zero, ConvertTransition(rule.DaylightTransitionEnd), int.MinValue, int.MaxValue);
-                    var recurringMap = new StandardDaylightAlternatingMap(StandardOffset, standardRecurrence, daylightRecurrence);
+                    IZoneIntervalMap recurringMap = new StandardDaylightAlternatingMap(StandardOffset, standardRecurrence, daylightRecurrence);
+                    // Fake 1 hour savings if the adjustment rule claims to be 0 savings.
```
malcolmr: Add a comment cross-referencing https://github.com/nodatime/nodatime/issues/746 ?

jskeet: Good plan, will do. Completely agree that it's weird - definitely contrary to my normal instincts, but it feels like it just about crosses the line into being appropriate.

PR #754, opened 2017-04-06
Fix redirect for latest API

### LINKED ISSUE #753: nodatime.org/api/ and /userguide/ "latest" redirects still use 1.3.x
```sh
$ curl -si http://nodatime.org/api/ | grep Location
Location: http://nodatime.org/1.3.x/api/
$ curl -si http://nodatime.org/userguide/ | grep Location
Location: http://nodatime.org/1.3.x/userguide/
```

Now that 2.0 is out, these version-less URLs should redirect to 2.0.x.

jskeet:
I thought I'd removed the IIS-based redirection, but I hadn't...
so I hadn't been testing the one in code either, which was broken.

Fixes #753.

### DIFF
--- a/src/NodaTime.Web/Startup.cs
+++ b/src/NodaTime.Web/Startup.cs
@@ -96,7 +96,7 @@ namespace NodaTime.Web
                 .AddRedirect($@"^{anyVersion}/userguide$", "$1/userguide/")
                 .AddRedirect($@"^developer$", "developer/")
                 // Make /api and /userguide links to the latest stable release.
-                .AddRedirect("^(api|userguide)((?:/.*))$", "1.3.x/$1$2");
+                .AddRedirect("^(api|userguide)(/.*)?$", "2.0.x/$1$2");
             app.UseRewriter(rewriteOptions);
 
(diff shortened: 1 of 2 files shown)


src/NodaTime.Web/Startup.cs
```diff
@@ -96,7 +96,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
                 .AddRedirect($@"^{anyVersion}/userguide$", "$1/userguide/")
                 .AddRedirect($@"^developer$", "developer/")
                 // Make /api and /userguide links to the latest stable release.
-                .AddRedirect("^(api|userguide)((?:/.*))$", "1.3.x/$1$2");
```
malcolmr: Am I right in thinking that this was doing a non-greedy match for `/.*` but then adding a `$` anchor after that match?

(and so the non-greedy part is uninteresting, which is why you've removed it?)

jskeet: I'm not sure why I made it greedy at all. I just needed it to be optional. Basically I experimented until it worked :)

PR #761, opened 2017-04-08
Fix website links

### LINKED ISSUE #760: Fix 404s and 302s on the nodatime.org website
There are some places where we link to URLs or resources that are 404 (e.g. http://nodatime.org/2.0.x/userguide/localtime-patterns, which has a link to `localdatetime-patterns.md` that's 404 because the suffix should be absent or `.html`). We should fix those.

There are also several places where we link to a URL that 302s to another URL. We should collapse those.
(Sometimes we do that deliberately, for `/api` and `/userguide`, but even then we should be linking to `/userguide/` to avoid a chain of two 302s in a row.)

We probably also want to check any external sites we link to. Now that I can run the website locally, this should all be fairly easy.

malcolmr:
Fixes various broken links within nodatime.org, collapses chains of 302, and rewrites URLs as https:// where possible.

See individual commits for more details, but in summary:

- fixes links from the user guide to API parameterised types (`IParseResult<T>`, etc) and namespaces, which were broken.
- fixes or removes some other miscellaneous broken links.
- removes `.html` suffixes from links to the user guide; these result in a 302.
- switches URLs for sites like Wikipedia, MSDN, etc to https://.
- update various URLs that have changed (e.g. blogs.msdn.com -> blogs.msdn.microsoft.com, with a different path).

Fixes #760.

### DIFF
--- a/src/NodaTime.Web/Markdown/1.0.x/concepts.md
+++ b/src/NodaTime.Web/Markdown/1.0.x/concepts.md
@@ -197,7 +197,7 @@ but they're still available within periods. [`Period`][Period] is used for arith
 See the [arithmetic](arithmetic) page for more information.
 
 [api]: ../api/
-[2]: http://blogs.msdn.com/b/bclteam/archive/2007/06/18/a-brief-history-of-datetime-anthony-moore.aspx
+[2]: https://blogs.msdn.microsoft.com/bclteam/2007/06/18/a-brief-history-of-datetime-anthony-moore/
 [LocalTime]: noda-type://NodaTime.LocalTime
 [LocalDate]: noda-type://NodaTime.LocalDate
 [LocalDateTime]: noda-type://NodaTime.LocalDateTime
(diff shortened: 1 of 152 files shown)


malcolmr:
Note that while most of these changes are within the `Markdown/` directory, there are some within the code itself.

jskeet:
Great, thanks. Haven't reviewed each individual file, but I've looked at the code changes and a smattering of the rest. Will pull and redeploy now.

jskeet:
Did you use something to check all the links, that we could put into CI of some kind? It would be good to try to avoid getting into this state again.

malcolmr:
@jskeet for broken internal links, I just ran `wget --spider -r -l inf -p -nd -o wget.log http://localhost:5000/`, which produces a report at the end (and told me about redirects). We could probably automate something like that without too much difficulty.

For external links, I used grep/sort/uniq to get a canonical list, and ran them through `curl` in order and inspected the results. We probably don't want to automate that one.

PR #762, opened 2017-04-08
Repeated check

jnyrup:
Without knowing the intended behavior, I guess this is the intended code.


jskeet:
Looks probable, yes. I'll need to check a bit more carefully and add a test, which I can make fail then merge your change to make it pass...

PR #763, opened 2017-04-08
Document how to build and run the website app

malcolmr:
Add a new page in the developer guide documenting how the nodatime.org website works, and include a script that can be used to quickly repopulate the docfx/ directory from the production content, for local development.

### DIFF
--- a/src/NodaTime.Web/Markdown/developer/website.md
+++ b/src/NodaTime.Web/Markdown/developer/website.md
@@ -1,0 +1,86 @@
+@Title="The nodatime.org website"
+
+## Overview
+
+[The nodatime.org website](http://nodatime.org/) is currently served from an
+ASP.NET Core web application hosted by Azure. The canonical source for the
+website is the [nodatime.org GitHub
+repository](https://github.com/nodatime/nodatime.org); changes to this
+repository are automatically deployed to the website (via a webhook that
+pings Azure to perform a pull/redeploy).
+
+However, the nodatime.org repository is typically _not_ edited directly,
+except in an emergency. Instead, changes are made in the nodatime
+repository, and then the nodatime.org repository is generated from that.
+
+The source for most of the website is the [NodaTime.Web
+project](https://github.com/nodatime/nodatime/tree/master/src/NodaTime.Web).
+This contains both the web application and the website content,
+particularly:
+
+- [`Markdown/`](https://github.com/nodatime/nodatime/tree/master/src/NodaTime.Web/Markdown)
+  containing the Markdown content used to generate the user and developer
+  guides.
+- [`wwwroot/`](https://github.com/nodatime/nodatime/tree/master/src/NodaTime.Web/wwwroot)
+  containing static content for the site.
+- [`Views/`](https://github.com/nodatime/nodatime/tree/master/src/NodaTime.Web/Views)
+  containing the programmatic content, for e.g. the home page and downloads
+  page.
+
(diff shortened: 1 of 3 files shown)


src/NodaTime.Web/Markdown/developer/website.md
```diff
@@ -0,0 +1,89 @@
...
+
+The [downloads](/downloads/) and [TZDB NZD files](/tzdb/) are stored on
+Google Cloud Storage, under the `https://storage.googleapis.com/nodatime/`
+bucket. The downloads page links directly to those files, but the NZD files
+are fetched and re-served via the nodatime.org web application (mostly
+because we do not want to serve redirects or non-nodatime.org URLs to
+clients).
+
+Both of these (downloads and TZDB files) are read from the Google Cloud
+Storage bucket on startup (by default).
```
jskeet: Add that they're refreshed periodically? That's why we don't need to restart the website when uploading a new release etc.

src/NodaTime.Web/Markdown/developer/website.md
```diff
@@ -0,0 +1,89 @@
...
+- [`Views/`](https://github.com/nodatime/nodatime/tree/master/src/NodaTime.Web/Views)
+  containing the programmatic content, for e.g. the home page and downloads
+  page.
+
+The [downloads](/downloads/) and [TZDB NZD files](/tzdb/) are stored on
+Google Cloud Storage, under the `https://storage.googleapis.com/nodatime/`
+bucket. The downloads page links directly to those files, but the NZD files
+are fetched and re-served via the nodatime.org web application (mostly
+because we do not want to serve redirects or non-nodatime.org URLs to
+clients).
```
jskeet: Do we want to mention that the NZD files are cached after fetching? Probably not vital. It does mean that if we ever need to modify an NZD file instead of releasing a new one, we'd need to restart the web app.

src/NodaTime.Web/Markdown/developer/website.md
```diff
@@ -0,0 +1,89 @@
...
+
+After this, visiting [http://localhost:5000/](http://localhost:5000/) should
+show the front page.
+
+Content is rendered at startup, so you'll need to restart the application if
+you change anything.
+
+## Deploying nodatime.org
+
+Well I run "buildweb.sh ../../nodatime.org" (which is where my local repo
```
jskeet: Given the paragraph below, just delete this one?
Although I typically wouldn't do it all in a single command - I use `git status` to check that the changes are roughly what I expect. Also, `git commit -a` won't add new files - `git add --all` followed by `git commit -m "Updated website"` might be better. (Or leave out the message to prompt.)

When using `;` between commands, is it still going to stop if any fail? We don't want to commit if the build failed, for example. I'd usually use `&&` but they may be equivalent...

src/NodaTime.Web/Markdown/developer/website.md
malcolmr: Done.

src/NodaTime.Web/Markdown/developer/website.md
malcolmr: Good point, done.

src/NodaTime.Web/Markdown/developer/website.md
malcolmr: Oops. That's why we have code reviews :). That was part of my draft notes (because yes, I basically copy/pasted our chat logs to start).

This was supposed to be informative rather than prescriptive, so I've simplified rather than listing an explicit command-line.

PR #767, opened 2017-04-09
Introduce test that deliberately fails

jskeet:
This will then be fixed by #762.


jskeet:
(Will validate that the only red test is the expected one before merging.)

PR #779, opened 2017-04-12
Reflection-based testing

jskeet:
- Make it simpler to do
- Add tests for names for conversions


jskeet:
That was deliberate actually - i realised that it would always be basically duplicating the test name :)

PR #781, opened 2017-04-14
Use a timer-based cache on the web site

### LINKED ISSUE #755: Website slows down (or halts?) every five minutes
After #751, we added an external (private) Prometheus/blackbox_exporter setup to probe nodatime.org.

One odd thing I noticed: there appear to be significant (350ms -> 1250ms) spikes at five-minute intervals, but only affecting the user guide / markdown content (I'm using http://nodatime.org/1.3.x/userguide/serialization, and requesting it every minute).

One possible clue: these _appear_ to be the times concurrent with requests for http://nodatime.org/tzdb/tzdb2016f.nzd, though the spikes continue even when we removed one of the NZD files and so that request was returning a 404 (which we did temporarily to test our monitoring).

Note that due to #751, we actually have most of Google Cloud Storage stubbed out at the moment (see cfb0164), so the only part involved should be the specific part that fetches and re-serves the NZD content (https://github.com/nodatime/nodatime/blob/master/src/NodaTime.Web/Models/TzdbDownload.cs).

jskeet:
This may or may not fix #755.

I've checked it out locally and it seems to do what we want. Will push it to the web site and we can see whether it improves things, then merge if so.

### DIFF
--- a/src/NodaTime.Web/Models/TimerCache.cs
+++ b/src/NodaTime.Web/Models/TimerCache.cs
@@ -1,0 +1,56 @@
+﻿using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Threading;
+
+namespace NodaTime.Web.Models
+{
+    public class TimerCache<T>
+    {
+        private readonly Timer timer;
+        private readonly Func<T> provider;
+        private readonly ILogger logger;
+        private readonly object padlock = new object();
+        private T value;
+        // TODO: Use Interlocked or similar to ensure proper volatility.
+        // It's possible that using volatile would be enough...
+        public T Value
+        {
+            get
+            {
+                lock (padlock)
+                {
+                    return value;
+                }
+            }
+            set
+            {
+                lock (padlock)
+                {
+                    this.value = value;
+                }
+            }
+        }
+
+        public TimerCache(IApplicationLifetime lifetime, Duration refreshPeriod, Func<T> provider, ILoggerFactory loggerFactory)
+        {
+            lifetime.ApplicationStopping.Register(() => timer?.Dispose());
+            logger = loggerFactory.CreateLogger(typeof(TimerCache<T>));
+            this.provider = provider;
+            timer = new Timer(Fetch, null, TimeSpan.Zero, refreshPeriod.ToTimeSpan());
+        }
+
+        private void Fetch(object state)
+        {
+            try
+            {
+                logger.LogInformation($"Refreshing cache for {typeof(T)}");
+                Value = provider();
+            }
+            catch (Exception e)
+            {
+                logger.LogError(0, e, $"Error fetching {typeof(T)}");
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/NodaTime.Web/Models/TimerCache.cs
```diff
@@ -0,0 +1,56 @@
...
+            logger = loggerFactory.CreateLogger(typeof(TimerCache<T>));
+            this.provider = provider;
+            timer = new Timer(Fetch, null, TimeSpan.Zero, refreshPeriod.ToTimeSpan());
+        }
+
+        private void Fetch(object state)
+        {
+            try
+            {
+                logger.LogInformation($"Refreshing cache for {typeof(T)}");
```
malcolmr: Worth adding a log line once the refresh is done?

src/NodaTime.Web/Models/TimerCache.cs
```diff
@@ -0,0 +1,56 @@
...
+                }
+            }
+        }
+
+        public TimerCache(IApplicationLifetime lifetime, Duration refreshPeriod, Func<T> provider, ILoggerFactory loggerFactory)
+        {
+            lifetime.ApplicationStopping.Register(() => timer?.Dispose());
+            logger = loggerFactory.CreateLogger(typeof(TimerCache<T>));
+            this.provider = provider;
+            timer = new Timer(Fetch, null, TimeSpan.Zero, refreshPeriod.ToTimeSpan());
```
malcolmr: I had to check the documentation to realise that this was "now and every refreshPeriod". Worth a comment? (or perhaps naming the `dueTime` parameter, but even that isn't completely clear.)

src/NodaTime.Web/Models/TimerCache.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Threading;
+
+namespace NodaTime.Web.Models
+{
+    public class TimerCache<T>
+    {
+        private readonly Timer timer;
+        private readonly Func<T> provider;
+        private readonly ILogger logger;
+        private readonly object padlock = new object();
```
malcolmr: (nit: first time I've seen `padlock` as the name of a lock variable; I would have expected just `lock` or `fooLock`.)

jskeet: `lock` is a keyword, otherwise I'd use it. I suppose it could be `valueLock`. I tend to use `padlock` unless I'm working in a convention which has underscores...

src/NodaTime.Web/Models/TimerCache.cs
jskeet: Will name the arguments and add a comment.

src/NodaTime.Web/Models/TimerCache.cs
jskeet: Done.

jskeet:
Rebuilding with NodaTime releases on a 6 minute cycle and TZDB on a 7 minute cycle. (Will take a few minutes before it's ready to deploy.)

Issue #711, opened 2017-03-27
Update our web site dependency packages


jskeet:
We're using old versions of Zepto, jQuery, Foundation etc - and almost certainly not using them in a way that's idiomatic for ASP.NET Core. (We should just be able to list them somewhere, I suspect....)

jskeet:
Okay, I've looked into this a bit.

In order to update to Foundation 6 (the current version) we'll need to tweak all our styles. I'm not competent to do this on my own (even though it's really just the _Layout that needs tweaking...

jskeet:
There's now a "bower" branch which is ugly as heck, but a reasonable start. Once we know what files we want, we can work out how to actually merge into master.

A few notes, so that Malcolm can have a go at making it look pretty:

- highlight-js is in a handlib directory as the Bower package is more a "package for making Javascript" than a ready-to-go highlighter, I think
- I've just noticed that we still have the foundicon fonts in the fonts directory. I think they can go now.
- Likewise images/conversions.png may well be redundant...
- Eventually we may want to remove the lib directory from git, so it's just restored on build
- On startup, the site looks on Google Cloud Storage for releases and TZDB files. The simplest way of handling that is to download a service account json file and set GOOGLE_APPLICATION_CREDENTIALS for it, or using `gcloud auth application-default login`; either way, let me know an account that needs access to the storage bucket. Maybe we should have a way of developing the site without this...
- You'll need a `docfx` directory which is normally built with `buildweb.sh`, but it's simplest just to grab a copy from https://github.com/nodatime/nodatime.org
- The files you'll want to edit most are likely to be:
  - Views/Documentation/Doc.cshtml
  - Views/Home/Index.cshtml
  - Views/Shared/_Layout.cshtml
  - wwwroot/css/main.css

malcolmr:
Having spent a little time with this tonight, I think I'm going to try dealing with this a slightly different way:

- while we could theoretically use Bower to pull foundation, Bower itself doesn't intend to give you a prebuilt binary, and Foundation 6 requires a load of build support, so I'd suggest instead that we download the prebuilt version from http://foundation.zurb.com/sites/download/ instead (which is where we originally pulled the version we're using, which I think is 4.2.1).
- we might find it easier to upgrade to Foundation 5 first: there's a [migration guide for that](http://foundation.zurb.com/sites/docs/v/5.5.3/upgrading.html).
- it might be nice to pull together the various snippets we have into `site.js`, then actually use the bundled/minified output in production mode.

[...]

PR #783, opened 2017-04-16
Serve bundled/minified CSS/JS on the website in production mode.

malcolmr:
Fold many of the CSS/JS assets into one bundled/minified CSS/JS resource (`site.min.css` and `site.min.js`), use that in the Staging/Production environments, and document how to regenerate them in the developer guide.

The only other significant changes here are:

- removal of relative links from font paths (which confuses the minifier).
- serving the resources for the NuGet button locally, rather than from a third-party site (and commenting out the CSS loader in favour of including the CSS directly).

Refs #711.


malcolmr:
Might be easier to review the individual commits, or alternatively you could just try it out.

Incidentally, according to [the MSDN documentation](https://docs.microsoft.com/en-us/aspnet/core/client-side/bundling-and-minification), the bundle/minify step should normally be performed on `dotnet build`, which would make a bit more sense. I tried adjusting the `csproj` file based on some examples I found, but I couldn't get that to work.

jskeet:
Will have a look later tonight or maybe tomorrow. I'd expect it to be minified when calling `dotnet publish` actually, but I could well be wrong... will look through that MSDN post carefully.

malcolmr:
Sorry, yes, it is also triggered by `dotnet publish`, but that doesn't get invoked by `dotnet run`. (From what I can see, it's supposed to hook into the build phase, which would be.)

PR #786, opened 2017-04-17
Update to highlight.js v9.10.0, and fix the included languages.

malcolmr:
Specifically, add `dos` in place of `java` (which we don't use), in addition to `bash` and `csharp`. This slightly fixes the display on e.g. `/developer/release-process` and `/unstable/userguide/tzdb` (it's more important on the `foundation-5` branch, where the default pre/code style looks a lot worse).


malcolmr:
Actually, reviewing this probably isn't too meaningful. I'll just merge myself.

jskeet:
I'm sure we used to have some Java code somewhere in the user guide, but I'm not sure where it was. Will take a look to check.

jskeet:
Nope, no idea. Maybe it was in an old bit of the developer guide. Anyway, fine to kill now - thanks :)

Issue #794, opened 2017-04-20
Web site resources should be cacheable


malcolmr:
At present, none of the resource files (CSS, JS, images, etc) served on nodatime.org are marked as cacheable in any way (no Expires/Cache-Control headers, etc). This means that browsers will apply heuristics to cache them (+10% of the difference between Date and Last-Modified, that kind of thing).

Additionally, we're not using any kind of cache busting, so changes to these resources (as I'm planning to do in #711) will roll out over a random time, effectively breaking the site for some people.

We should implement some kind of resource renaming, and then send proper cache headers for those resources. Something like [StaticWebHelper](http://madskristensen.net/post/optimizing-static-websites-on-iis) looks like it should do the trick.

malcolmr:
`StaticWebHelper` is .NET 4.5-only, but it turns out this is mostly built-in to ASP.NET; we just need to configure it properly.

PR #797, opened 2017-04-22
Upgrade from Foundation 4 to Foundation 5

### LINKED ISSUE #711: Update our web site dependency packages

malcolmr:
Upgrade the website from Foundation 4.something to 5.5.3.

This is mostly straightforward. There are a few minor differences:

- The navbar has slightly different padding for the menu items, and styling for the Install button (and the dividers look a little different).
- The `<code>` font isn't bold, and has a slight background (see e.g. the type names and links on http://nodatime.org/2.0.x/userguide/type-choices). I think this actually looks better.
- The side-bar menu in e.g. the user guide no longer has clickable section headers (Introduction, Core, Text, etc). These didn't do anything when clicked anyway.
- Padding and margins are slightly different throughout, though not by much.
- The site might not work in IE7/8 (that's <0.05% of our users).

Fixes #711.


malcolmr:
(Probably want to resolve #794 before pushing this live, btw.)

jskeet:
Given the comment about caching, do we want to try to build that into this PR? I haven't looked at caching (or cache-busting) in ASP.NET Core at all yet.

malcolmr:
I think the cache-busting (we already have caching implicitly, albeit not well-configured caching) is orthogonal to this, but we should probably wait to have a good solution before submitting this (or any large-scale change, really).

(I think the solution is going to be pretty easy, so that's the next thing I'm planning to look at.)

jskeet:
The simplest way of me testing this is probably just to merge it, pull and regenerate the website, then test it locally. I'll only push if it looks okay :)

jskeet:
Tested, and it looks fine. Pushing to nodatime.org.

PR #799, opened 2017-04-24
Provide sensible cache behaviour for static content

### LINKED ISSUE #794: Web site resources should be cacheable

malcolmr:
Add Cache-Control headers and cachebusting parameters to static content responses, as appropriate.

Fixes #794.

### DIFF
--- a/src/NodaTime.Web/Startup.cs
+++ b/src/NodaTime.Web/Startup.cs
@@ -86,7 +86,7 @@ namespace NodaTime.Web
                 {
                     Mappings = { [".nzd"] = "application/octet-stream" }
                 },
-                OnPrepareResponse = (context) => SetCacheControlHeaderForStaticContent(env, context.Context)
+                OnPrepareResponse = context => SetCacheControlHeaderForStaticContent(env, context.Context)
             });
 
             // API documentation
@@ -96,7 +96,8 @@ namespace NodaTime.Web
                 ContentTypeProvider = new FileExtensionContentTypeProvider
                 {
                     Mappings = { [".yml"] = "text/x-yaml" }
-                }
+                },
+                OnPrepareResponse = context => SetCacheControlHeaderForStaticContent(env, context.Context)
             });
             // Captures "unstable" or a specific version - used several times below.
             string anyVersion = @"((?:1\.[0-3]\.x)|(?:unstable)|(?:2\.0\.x))";
@@ -162,9 +163,10 @@ namespace NodaTime.Web
                 return;
             }
 
-            // Otherwise, the remaining content (/favicon.ico, /fonts/, /robots.txt, etc) should be good to use for a
-            // while without revalidation. When running in the Development environment, we'll use a much shorter time,
-            // since we might be iterating on it (in particular, this also covers the unminified JS/CSS).
+            // Otherwise, the remaining content (/favicon.ico, /fonts/, /robots.txt, /styles/docfx.js etc) should be
+            // good to use for a while without revalidation. When running in the Development environment, we'll use a
+            // much shorter time, since we might be iterating on it (in particular, this also covers the unminified
+            // JS/CSS).
             headers.CacheControl = new CacheControlHeaderValue
             {
                 Public = true,
--- a/src/NodaTime.Web/Startup.cs
+++ b/src/NodaTime.Web/Startup.cs
@@ -4,12 +4,15 @@
 
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Headers;
 using Microsoft.AspNetCore.Rewrite;
 using Microsoft.AspNetCore.StaticFiles;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
+using Microsoft.Net.Http.Headers;
 using NodaTime.Web.Controllers;
 using NodaTime.Web.Models;
 using NodaTime.Web.Providers;
@@ -20,6 +23,8 @@ namespace NodaTime.Web
 {
     public class Startup
     {
+        private static readonly MediaTypeHeaderValue TextHtml = new MediaTypeHeaderValue("text/html");
+
         // Not const to avoid unreachable code warnings.
         private static readonly bool UseGoogleCloudStorage = Environment.GetEnvironmentVariable("DISABLE_GCS") == null;
 
@@ -80,8 +85,10 @@ namespace NodaTime.Web
                 ContentTypeProvider = new FileExtensionContentTypeProvider
                 {
                     Mappings = { [".nzd"] = "application/octet-stream" }
-                }
+                },
+                OnPrepareResponse = (context) => SetCacheControlHeaderForStaticContent(env, context.Context)
             });
+
             // API documentation
             app.UseStaticFiles(new StaticFileOptions
             {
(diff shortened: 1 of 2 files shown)


malcolmr:
It's worth noting that we already send an ETag header for all of these, so the browser will already perform a conditional (If-None-Match) request and a get a 304 if the resource hasn't changed, but setting max-age will avoid the request in the first place.

src/NodaTime.Web/Startup.cs
```diff
@@ -80,6 +84,43 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
+                        // figure out when to revalidate.
+                        return;
+                    }
+
+                    // Otherwise send a sensible max-age value. If the request was made with a file version query
+                    // parameter (?v=..., as sent by asp-append-version=true), then cache the response 'indefinitely'.
+                    // Otherwise (e.g. /favicon.ico, /fonts/, /robots.txt), cache for a day.
+                    if (context.Context.Request.Query.ContainsKey("v"))
+                    {
+                        headers.CacheControl = new CacheControlHeaderValue()
```
malcolmr: I can write `headers.CacheControl.MaxAge = ...` here, but while it compiles, it doesn't do anything.

I can only assume that the getters for the `ResponseHeaders` type return mutable value-type objects, rather than proxies, and so it's actually necessary to use the setters to get anything to propagate through to the underlying `Context.Response.Headers` dictionary.

src/NodaTime.Web/Startup.cs
```diff
@@ -80,6 +84,43 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
                 ContentTypeProvider = new FileExtensionContentTypeProvider
                 {
                     Mappings = { [".nzd"] = "application/octet-stream" }
+                },
+                OnPrepareResponse = (context) => {
+                    // Set the Cache-Control header for static content.
+                    var headers = new ResponseHeaders(context.Context.Response.Headers);
+
+                    // All static content is cacheable by intermediates.
+                    headers.CacheControl = new CacheControlHeaderValue()
```
malcolmr: I'm wondering whether I should just remove this, as I'm not sure that "Cache-Control: public" with no max-age is that useful. (I added this originally when I thought I could create the object once; see below.)

src/NodaTime.Web/Startup.cs
```diff
@@ -80,6 +84,43 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
+                            Public = true,
+                            MaxAge = TimeSpan.FromDays(365)
+                        };
+                    }
+                    else
+                    {
+                        headers.CacheControl = new CacheControlHeaderValue()
+                        {
+                            Public = true,
+                            MaxAge = TimeSpan.FromDays(1)
```
malcolmr: This might potentially be annoying for local development. Perhaps I should guard this part in an envrionment != Development check?

src/NodaTime.Web/Startup.cs
jskeet: Right. No need for the `()` at the end of the line, btw.

jskeet: An alternative would be to build up the `CacheControlHeaderValue` object and then set it at the end. Possibly even just write a method for creating that, so it's not cluttering up this method?

src/NodaTime.Web/Startup.cs
jskeet: Maybe test locally that there isn't something that already does this, but if so, sure.

src/NodaTime.Web/Startup.cs
jskeet: Pass. I'm far from expert on this. However, this is all just about serving static content, yes? Should we do something about the Markdown we serve as well?

src/NodaTime.Web/Startup.cs
malcolmr: Done (the `()`), and I've split this out to a separate method, so I don't think the duplication is a problem so much.

src/NodaTime.Web/Startup.cs
```diff
@@ -80,8 +85,10 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
                 ContentTypeProvider = new FileExtensionContentTypeProvider
                 {
                     Mappings = { [".nzd"] = "application/octet-stream" }
-                }
+                },
+                OnPrepareResponse = (context) => SetCacheControlHeaderForStaticContent(env, context.Context)
```
jskeet: Nit: no need for the brackets around `context`. Should have spotted that before.

src/NodaTime.Web/Startup.cs
```diff
@@ -129,5 +136,40 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
+            var headers = new ResponseHeaders(context.Response.Headers);
+
+            if (headers.ContentType.IsSubsetOf(TextHtml))
+            {
+                // Don't set Cache-Control for HTML files (e.g. /tzdb/). The browser can figure out when to revalidate.
+                // (Which it can do easily, since we send an ETag with all static content responses).
+                return;
+            }
+
+            // Otherwise if the request was made with a file version query parameter (?v=..., as sent by
```
jskeet: Given that the logic is only about setting MaxAge as far as I can tell, I'd probably do that just with:

```csharp
var maxAge =
    context.Request.Query.ContainsKey("v") ? TimeSpan.FromDays(365)
    : env.IsDevelopment ? TimeSpan.FromMinutes(2)
    : TimeSpan.FromDays(1);

headers.CacheControl =...;
```

(With an explanatory comment, of course.) But this is fine too.

src/NodaTime.Web/Startup.cs
malcolmr: Removed this (it really only affects /tzdb/).

It would be nice to send an ETag for text/html content (like Markdown) and allow the browser to revalidate the cached version cheaply, but I'm wary of allowing the browser to avoid the round-trip entirely.

In particular, since ASP.NET doesn't validate the `?v=` parameter is actually correct, you can get a nasty problem if you allow an old referencing page to be used without revalidation:

1. Browser fetches `/foo/` (HTML) and then `/foo.css?v=1` (JS). The first has Cache-Control: max-age=5 minutes, the latter Cache-Control: max-age=1 year.
2. The site is updated. `/foo/` now uses `/foo.css?v=2`.
2. The user navigates back to `/foo/`. The browser reuses the HTML resource without revalidating it, but refetches `/foo.css?v=1` (because it's dropped out of the cache). This actually returns the _new_ content (I'd consider this a bug), and the browser permanently caches the new CSS with the old HTML content. (Alternatively: the server sensibly returns 404, and the page is also broken.)

src/NodaTime.Web/Startup.cs
malcolmr: Reduced to a few minutes in development mode. I think that's the best of both: it allows for iteration, but it's representative of the production mode too.

src/NodaTime.Web/Startup.cs
malcolmr: Actually, no reason I can't cache the docfx/ JS and CSS as well. Done.

src/NodaTime.Web/Startup.cs
malcolmr: True, but I think the separate version is a bit more readable here (and the minor duplication doesn't bother me too much).

src/NodaTime.Web/Startup.cs
malcolmr: Done, thanks.

src/NodaTime.Web/Startup.cs
jskeet: Fair enough :)

PR #805, opened 2017-04-27
Add the Google Analytics tracking code to the API documentation.

malcolmr:
Add the Google Analytics tracking code to the API documentation, by overriding the default DocFX `partials/scripts` template.


malcolmr:
Note that I have no easy way to test this, though it should be fine. If so, could you merge and regenerate the documentation?

(Also I have no idea in what case the corresponding Liquid template is used, but I'm guessing it's not important for us?)

jskeet:
Looks fine - will do my normal "Merge, build, test locally, deploy"

PR #817, opened 2017-05-07
Highlight the 1.3.x and 2.0.x versions of the API and User Guide.

malcolmr:
Place the current (1.3.x and 2.0.x) versions at the top of the API/User Guide menu options, and rename "Unstable" to "Development" (in the menu label only).

Also remove the dividers, which don't seem to be adding much.

### DIFF
--- a/src/NodaTime.Web/Views/Shared/_Layout.cshtml
+++ b/src/NodaTime.Web/Views/Shared/_Layout.cshtml
@@ -50,28 +50,28 @@
                     <li class="has-dropdown">
                         <a href="#">User Guide</a>
                         <ul class="dropdown">
-                            <li><label>Versions</label></li>
+                            <li><label>Current</label></li>
+                            <li><a href="/1.3.x/userguide/">1.3.x</a></li>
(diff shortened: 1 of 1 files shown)


src/NodaTime.Web/Views/Shared/_Layout.cshtml
```diff
@@ -50,28 +50,28 @@
                     <li class="has-dropdown">
                         <a href="#">User Guide</a>
                         <ul class="dropdown">
-                            <li><label>Versions</label></li>
+                            <li><label>Current</label></li>
+                            <li><a href="/1.3.x/userguide/">1.3.x</a></li>
+                            <li><a href="/2.0.x/userguide/">2.0.x</a></li>
+                            <li><label>Other</label></li>
```
jskeet: Maybe "Old"? And should this be under "Development", as the old releases are arguably less important?

malcolmr: Done. PTAL?

PR #818, opened 2017-05-07
Include symbols in nuget packages

jskeet:


malcolmr:
Is this something we're going to want in a 2.0.2?

jskeet:
Yup, hence merging it into the 2.0.x branch as well :) We haven't messed up anything, it's just a matter of changing *where* we say that we want symbols - so that I can just use `dotnet pack` instead of `dotnet pack --include-symbols`. I noticed due to the TZDB update scripts I've been writing.

PR #821, opened 2017-05-18
First pass at a snippet extractor

jskeet:
Lots more to do, but the results are pretty nice.
Next steps:

- Write a blog post
- Use a script to make all the simplifications instead of adding
  documents to the project
- Revisit all the Roslyn stuff
- Consider separating out into a new project

### DIFF
--- a/build/SnippetExtractor/Program.cs
+++ b/build/SnippetExtractor/Program.cs
@@ -1,0 +1,220 @@
+﻿// Copyright 2017 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Scripting;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.MSBuild;
+using Microsoft.CodeAnalysis.Scripting;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SnippetExtractor
+{
+    partial class Program
+    {
+        private static int Main(string[] args)
+        {
+            if (args.Length != 3)
+            {
+                Console.WriteLine("Arguments: <project file> <referenced assembly paths> <output directory>");
+                return 1;
+            }
+            try
+            {
+                ExtractSnippets(args[0], args[1], args[2]).Wait();
+                return 0;
+            }
+            catch (AggregateException e)
+            {
+                Console.WriteLine($"Error: {e.InnerException}");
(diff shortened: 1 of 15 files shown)


build/SnippetExtractor/Program.cs
```diff
@@ -0,0 +1,220 @@
...
+                return 1;
+            }
+        }
+
+        private static async Task ExtractSnippets(string projectFile, string assemblyPaths, string outputDirectory)
+        {
+            // Make sure snippets run in an invariant culture.
+            CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
+
+            // TODO: Work out why it doesn't work when targeting netcoreapp1.0
```
malcolmr: Don't you have a github issue for this already?

build/SnippetExtractor/Program.cs
```diff
@@ -0,0 +1,220 @@
...
+                    var doc = project.GetDocument(entry.Value);
+                    var root = await doc.GetSyntaxRootAsync();
+                    var tree = root.SyntaxTree;
+                    var usings = tree.GetCompilationUnitRoot().Usings.Select(uds => uds.ToString()).ToList();
+                    var method = root.DescendantNodes().OfType<MethodDeclarationSyntax>().Single();
+                    var lines = Trim(method.Body.GetLines());
+
+                    writer.WriteLine("---");
+                    writer.WriteLine($"uid: {uid}");
+                    // TODO: This will overwrite any existing remarks. We'd like a separate "example"
```
malcolmr: as in, this overwrites `<remarks/>`? That's not great :(

build/SnippetExtractor/Program.cs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        private async static Task<Document> RemoveUnusedImportsAsync(Document document)
+        {
+            // TODO: See if there's a better way of doing this.
+            var compilation = await document.Project.GetCompilationAsync();
+            var tree = await document.GetSyntaxTreeAsync();
+            var root = tree.GetRoot();
+            var unusedImportNodes = compilation.GetDiagnostics()
+                .Where(d => d.Id == "CS8019")
```
malcolmr: Oh, very nice.

build/SnippetExtractor/Program.cs
jskeet: Ah - no, it now doesn't. I fixed the code, but not the TODO. Oops.

I'm expecting to have another iteration within 24 hours, so I'll remove the TODO for that.

build/SnippetExtractor/Program.cs
jskeet: Assuming you were being sarcsastic - I know, it's horrible. I believe there's a nice way of doing this by adding a "simplify" annotation and then asking Roslyn to simplify everything. But I haven't worked it out yet, and I thought I'd give @DustinCampbell a rest after Build...

malcolmr: Actually, no! I think using the compiler diagnostics is actually a pretty smart way to do this - in the absence of a standard transformation, anyway.

PR #825, opened 2017-05-21
Fix Period computation bug

### LINKED ISSUE #824: Period.Between can report partially-negative values
This isn't meant to happen...

Reported on http://stackoverflow.com/questions/44087757

```csharp
var start = DateTime.Now;
var end = start.AddDays(7).AddMinutes(-1);
        
var period = Period.Between(LocalDateTime.FromDateTime(start), LocalDateTime.FromDateTime(end));
Console.WriteLine(period); // P6DT24H-1M
```

More investigation required.

jskeet:
The approach taken before was fundamentally unsound. It's clearer to
consider that when we get down to time-based fields, they're all
fixed-length, so we only need to consider the duration between the
start and end points. This is an initial fix to patch into 2.0.2 -
it's not as efficient as it might be, but it's reasonably minimal.

Further refactoring will come in a separate PR, but that can stay on
the master branch only.

Fixes #824.

### DIFF
--- a/src/NodaTime.Test/PeriodTest.cs
+++ b/src/NodaTime.Test/PeriodTest.cs
@@ -67,6 +67,22 @@ namespace NodaTime.Test
         }
 
         [Test]
+        public void BetweenLocalDateTimes_AcrossDays()
+        {
+            Period expected = Period.FromHours(23) + Period.FromMinutes(59);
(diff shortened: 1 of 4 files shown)


jskeet:
Wonderful, thanks. Will get cracking on a 2.0.2 release.

PR #826, opened 2017-05-21
Refactor period

jskeet:
This follows on from #825, performing several refactorings afterwards.

I believe they will improve performance too, but I haven't validated that yet. They certainly make the code easier to understand, IMO.

### DIFF
--- a/src/NodaTime/Calendars/HebrewYearMonthDayCalculator.cs
+++ b/src/NodaTime/Calendars/HebrewYearMonthDayCalculator.cs
@@ -159,26 +159,24 @@ namespace NodaTime.Calendars
             return new YearMonthDay(year, month, day);
         }
 
-        // Note to self: this is (minuendInstant - subtrahendInstant) in months. So if minuendInstant
-        // is later than subtrahendInstant, the result should be positive.
-        internal override int MonthsBetween(YearMonthDay minuendDate, YearMonthDay subtrahendDate)
+        internal override int MonthsBetween(YearMonthDay start, YearMonthDay end)
         {
             // First (quite rough) guess... we could probably be more efficient than this, but it's unlikely to be very far off.
-            int minuendCivilMonth = CalendarToCivilMonth(minuendDate.Year, minuendDate.Month);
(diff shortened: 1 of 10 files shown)


src/NodaTime/Calendars/HebrewYearMonthDayCalculator.cs
```diff
@@ -159,26 +159,24 @@ internal override YearMonthDay AddMonths(YearMonthDay yearMonthDay, int months)
...
-            double subtrahendTotalMonths = subtrahendCivilMonth + (subtrahendDate.Year * MonthsPerLeapCycle) / (double) YearsPerLeapCycle;
-            int diff = (int) (minuendTotalMonths - subtrahendTotalMonths);
+            int startCivilMonth = CalendarToCivilMonth(start.Year, start.Month);
+            double startTotalMonths = startCivilMonth + (start.Year * MonthsPerLeapCycle) / (double)YearsPerLeapCycle;
+            int endCivilMonth = CalendarToCivilMonth(end.Year, end.Month);
+            double endTotalMonths = endCivilMonth + (end.Year * MonthsPerLeapCycle) / (double) YearsPerLeapCycle;
+            int diff = (int) (endTotalMonths - startTotalMonths);
 
-            if (Compare(subtrahendDate, minuendDate) <= 0)
+            if (Compare(start, end) <= 0)
```
malcolmr: Perhaps `YearMonthDay` should implement operator< ?

Oh wait, it does. Maybe use it?

jskeet: It does, but we don't want to use it - because it might give the wrong answer for the Hebrew calendar because of the month numbering issue.

From the docs of `YearMonthDay`:

```csharp
/// This type is naive: comparisons are performed assuming that a larger month number
/// always comes after a smaller month number, etc.
/// This is suitable for most, but not all, calendar systems.
```

:)

jskeet:
Hmm... not sure why it thinks we've got conflicts here. Will look into that...

jskeet:
Ah, presumably it's due to #825 being merged. Will rebase and repush, which should sort it.

PR #832, opened 2017-05-29
Further tidy-ups of Period

jskeet:
- Use long arithmetic more widely, only using Duration arithmetic
  where absolutely necessary (big performance boost)
- Use DivRem where available
- Remove TODO comments after testing

### DIFF
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -432,36 +432,43 @@ namespace NodaTime
                     start.Date, endDate, units, out years, out months, out weeks, out days);
                 remaining = new LocalDateTime(remainingDate, start.TimeOfDay);
             }
-
-            long hours = 0, minutes = 0, seconds = 0, milliseconds = 0, ticks = 0, nanoseconds = 0;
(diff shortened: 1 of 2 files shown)


src/NodaTime/Period.cs
```diff
@@ -498,6 +505,47 @@ int UnitsBetween(PeriodUnits maskedUnits, ref LocalDate startDate, LocalDate end
...
         }
 
+        /// <summary>
+        /// Common code to perform the time parts of the Between methods for long-representable nanos.
+        /// </summary>
+        /// <param name="totalNanoseconds">Number of nanoseconds to compute the units of</param>
+        /// <param name="units">Units to compute</param>
+        /// <param name="hours">(Out) Hours component of result</param>
+        /// <param name="minutes">(Out) Minutes component of result</param>
+        /// <param name="seconds">(Out) seconds component of result</param>
```
malcolmr: very nit: "Seconds component"

jskeet: Ooh, thanks - will fix.

PR #833, opened 2017-05-29
Script to run benchmarks from freshly-cloned repo

jskeet:
The intention is to run this in a cron job / scheduled task.

### DIFF
--- a/build/runbenchmarks.sh
+++ b/build/runbenchmarks.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+set -e
+
+if [[ "$3" = "" ]]
+then
+  echo "Usage: runbenchmarks.sh <repo-directory> <results-directory> <target-framework> [target-framework...]"
+  echo "e.g. runbenchmarks.sj benchmark-tmp benchmark-results netcoreapp1.1 net45"
+  exit 1
+fi
+
+declare -r REPO=$1
+declare -r RESULTS=$(realpath $2)
+
+shift 2
+
+rm -rf $REPO
+git clone https://github.com/nodatime/nodatime.git $REPO --depth=1
+
+declare -r COMMIT=$(git -C $REPO rev-parse HEAD)
+
+if [[ -d $RESULTS/$COMMIT ]]
+then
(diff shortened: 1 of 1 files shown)


build/runbenchmarks.sh
```diff
@@ -0,0 +1,48 @@
...
+  OUTPUT=$RESULTS/$COMMIT/$TARGET_FRAMEWORK
+  mkdir -p $OUTPUT
+  echo "Running benchmarks for $TARGET_FRAMEWORK"
+  rm -rf BenchmarkDotNet.Artifacts
+  date -u -Iseconds > $OUTPUT/start.txt
+  dotnet run -f $TARGET_FRAMEWORK -c Release -- --exporter=briefjson '*'
+  date -u -Iseconds > $OUTPUT/end.txt
+  for report in BenchmarkDotNet.Artifacts/results/*-report-brief.json
+  do
+    dest=$(basename $report | sed s/-report-brief.json//g)
```
malcolmr: And actually I'd probably write `dest=$(...).json` to make it clearer that we were adding the suffix.

Or perhaps `dest=${$(basename report)/-report-brief/}`, which does the basename and then strips the middle part of the string.

build/runbenchmarks.sh
```diff
@@ -0,0 +1,48 @@
...
+declare -r RESULTS=$(realpath $2)
+
+shift 2
+
+rm -rf $REPO
+git clone https://github.com/nodatime/nodatime.git $REPO --depth=1
+
+declare -r COMMIT=$(git -C $REPO rev-parse HEAD)
+
+if [[ -d $RESULTS/$COMMIT ]]
```
malcolmr: I assume RESULTS can't contain spaces, otherwise we're going to have a bad time :)

build/runbenchmarks.sh
jskeet: Just as per that final part, or using sed? (Is sed somehow built into `${...}`?)

build/runbenchmarks.sh
jskeet: Yes, I'll make that assumption at least for now...

PR #851, opened 2017-06-09
Duration optimizations

### LINKED ISSUE #838: Perform arithmetic in decimal rather than BigInteger (was: Some BigInteger performance measurements)
I was most surprised by our experience in #837, so I've done a little bit of bench-marking of the relative performance of various ways `Duration` does some of its stuff - and also threw-in `Decimal` as an alternative to `BigInteger`.

The results look like this:

```
                        Method |      Mean |      Error |    StdDev |  Gen 0 | Allocated |
------------------------------ |----------:|-----------:|----------:|-------:|----------:|
   DurationFromNanosecondsLong |  39.98 ns |  0.8061 ns | 0.8626 ns |      - |       0 B |
 DurationFromNanosecondsBigInt | 518.35 ns | 10.2473 ns | 9.0840 ns | 0.0391 |     124 B |
               ConstructBigInt | 493.81 ns |  4.9034 ns | 4.3467 ns | 0.0448 |     144 B |
                 ConstructLong |  36.76 ns |  0.5417 ns | 0.5067 ns |      - |       0 B |
              ConstructDecimal | 254.47 ns |  2.8259 ns | 2.6433 ns |      - |       0 B |
```
The two 'DurationFromxxx' methods are using NodaTime's `Duration` constructors, and the three 'Construct' methods are really the innards of the `Duration.FromNanoseconds` methods, implemented with BigInt, `long` and `Decimal` respectively.

I have not bothered to check any of this for correctness - obviously `BigInteger` goes places that `long` doesn't, but it's very slow getting there.

I guess, though I haven't tried it, that many uses of `BigInteger` could be got rid of if there was a desire to do this - this would be a very direct trade-off of clean, simple code against performance.  With a 12:1 perf advantage this might feel worthwhile?

This isn't just pointless micro-bench-marking, there are clearly situations that run through these paths are lot, and even if in #837 the threshold between fast and slow gets tweaked to avoid the slow path more often, it's still going to be very slow when you do hit it.

jskeet:
Fixes #838 

(I've managed to delete the benchmark comparisons I had, but they were fairly significant - particularly avoiding using BigInteger on the way to double. When we've got benchmarks on the website I can edit this PR comment to link to an appropriate comparison :)

### DIFF
--- a/src/NodaTime/Duration.cs
+++ b/src/NodaTime/Duration.cs
@@ -18,6 +18,13 @@ using System.Numerics;
 
 namespace NodaTime
 {
+    // Implementation note:
+    // Although we use BigInteger in the public API as it's a better fit for "this could be a large integer"
+    // we use decimal for computations where we have to. It's more efficient in CPU (and space) than BigInteger,
+    // and still has plenty of range/precision. The FromNanoseconds(BigInteger) and ToBigIntegerNanoseconds() methods
+    // should not be called within Noda Time, except for in FromNanoseconds(double), as the double to decimal conversion
+    // loses information whereas double to BigInteger doesn't.
+
     /// <summary>
     /// Represents a fixed (and calendar-independent) length of time.
     /// </summary>
@@ -66,6 +73,8 @@ namespace NodaTime
         internal const int MinDays = ~MaxDays;
         internal static readonly BigInteger MinNanoseconds = (BigInteger)MinDays * NanosecondsPerDay;
         internal static readonly BigInteger MaxNanoseconds = (MaxDays + BigInteger.One) * NanosecondsPerDay - BigInteger.One;
+        internal static readonly decimal MinDecimalNanoseconds = (decimal) MinNanoseconds;
+        internal static readonly decimal MaxDecimalNanoseconds = (decimal) MaxNanoseconds;
         private static readonly double MinDoubleNanoseconds = (double) MinNanoseconds;
         private static readonly double MaxDoubleNanoseconds = (double) MaxNanoseconds;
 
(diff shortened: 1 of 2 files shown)


jskeet:
Urgh - test failure will need investigation.

jskeet:
Hmm. Test failure appears to basically show that it was broken before. Not sure why.

jskeet:
Aaargh. No. `double` to `decimal` conversion loses information :(

jskeet:
Okay, I think I've found the *only* conversion from `double` to `decimal` and changed it to use `BigInteger`. Looks like the tests pass now - please review :)

src/NodaTime/Duration.cs
```diff
@@ -1154,13 +1177,18 @@ void IXmlSerializable.WriteXml([NotNull] XmlWriter writer)
         [Pure]
         private long ToInt64NanosecondsUnchecked() => unchecked(days * NanosecondsPerDay + nanoOfDay);
 
+        // NOTE: Do not use from internal code. Use decimal instead. TODO: Add an attribute, and use a Roslyn analyzer to validate.
```
malcolmr: (Rather than add an attribute, might it be simpler to just disallow any `double` -> `decimal` conversions? It sounds like it would never be a good idea.)

src/NodaTime/Fields/TimePeriodField.cs
```diff
@@ -27,15 +27,15 @@ internal sealed class TimePeriodField
         internal static readonly TimePeriodField Hours = new TimePeriodField(NanosecondsPerHour);
 
         private readonly long unitNanoseconds;
+        // The largest number of units (positive or negative) we can multiply unitNanoseconds by without overflowing.
```
malcolmr: (perhaps "without overflowing a `long`"?)

src/NodaTime/Duration.cs
jskeet: It would be good to disallow that within Roslyn, yes. But we should *also* disallow using BigInteger in almost every other situation. Basically this is a bit of a tricky situation.

src/NodaTime/Fields/TimePeriodField.cs
jskeet: Sure, will do.

jskeet:
(Will merge on green.)

PR #853, opened 2017-06-10
First full pass at benchmarks

jskeet:
- Uploader (to be run on each benchmarking machine periodically)
- Crude UI

The benchmarks are currently stored in GCS as protobuf files, then loaded into memory (and polled for changes). Currently rebuilding the web site, at which point I'll push this... if it looks okay, I'll merge this PR.


jskeet:
Thinking about this more, I'm going to ditch BenchmarkContainer and add repeated fields to BenchmarkEnvironment, BenchmarkRun and BenchmarkType. Let's make the hierarchy easily navigable where possible. When we use a database instead, we can change more.

jskeet:
Done that - now pushing to nodatime.org. If the site is still working in the morning (and responds when I upload the benchmarks which should show the benefits of #838) I'll merge this PR. It doesn't affect the NodaTime library code at all.

Issue #311, opened 2015-03-15
Decide exactly what will be in 1.4.0


GoogleCodeExporter:
```
(marking this as 2.0-consider, because it's tightly coupled to what ends up in 
2.0.0.)

It would be nice to be able to write code that worked in Noda Time 1.x and 2.x. 
 However, due to changes to e.g. IClock for 2.0, that's not currently possible.

When we actually get a bit closer to 2.0.0 and can see what's in it, it'll be 
worth considering whether a 1.4 should exist, to bridge 1.x and 2.x.  

The obvious thing that it could contain so far would be a deprecated IClock.Now 
(not deprecated in 1.3; not present in 2.0) and an additional 
IClock.GetCurrentInstant() [though as Jon points out, adding a method to an 
interface could itself be considered a breaking change].

[It _might_ also be nice to backport some of the new-in-2.0 APIs back to 1.x to 
make sharing code between 1.x and 2.x a little easier: for example, 
Offset.FromSeconds() would be trivial to backport. That's far less important 
than the IClock case, though.]
```

Original issue reported on code.google.com by `malcolm.rowe` on 29 Jul 2014 at 2:03

malcolmr:
We've decided it should exist. See initial analysis/proposal at https://groups.google.com/d/msg/noda-time/NO5NgTAVfUw/aQYuSK2jDi4J.

I'm going to morph this to "decide exactly what goes into 1.4.0", since we'll still have work to do when we're closer to 2.0

jskeet:
http://nodatime.org/2.0.x/api/changes.html should be helpful here.

jskeet:
Reminder: we need to work out what "obsolete" really means. If there's no replacement in 2.0.x (e.g. for specifying the min-days-of-first-week) for a calendar system, do we want to make that obsolete, or are we only aiming at getting developers to a situation where they are in as good a position as they can be for 2.0?

jskeet:
This is now basically done, as 1.4.0 beta 01 is released :)

PR #860, opened 2017-06-13
First steps for 1.4

jskeet:
I'm assuming it's better to start than to wait until #311 is completely nailed down before doing anything...


.gitignore
```diff
@@ -69,3 +69,12 @@ nuget/**
 /src/**/*-*.csproj
 
 /docs
+
+
+# FIles present in master that get in the way when switching
```
malcolmr: nit: typo.

src/NodaTime/CalendarSystem.cs
```diff
@@ -189,6 +191,7 @@ public static CalendarSystem ForId(string id)
         /// this implementation align exactly with the BCL implementation.
         /// </remarks>
         /// <returns>A Persian calendar system.</returns>
+        [Obsolete("Use the PersianSimple property for compatibility with 2.0.")]
```
malcolmr: You're not obsoleting the others because there's not always an alternative?

(If so, I'm not sure about that: they are obsolete, even if they can't be directly replaced.)

jskeet: Mmm. It's a tricky one. I think the word "obsolete" is unfortunate in general for this. Will leave it as is right now, but let's revisit this when we have some more details.

PR #871, opened 2017-06-19
Added option to override default ToString implementation

Jivko1212:
I would like to be able to change the default ToString for the basic noda data types. My locale is Bulgaria and ToString results in cyrillic output that is very inconvenient to work with. Also I am working with financial data and I would like to have all LocalTimes output with nanosecond precision in the debugger or when I print or interpolate them.

I achieved this by adding two delegates, static DefaultStringFormatter and DefaultIFormattableFormatter .. they are used by the new versions of ToString() and ToString(string, IFormatProvider). They are initially set to their default values but by overriding them you can change the ToString behavior.

Here is some example code using my modifications along with the output before overriding ToString and the output after overriding ToString:

https://gist.github.com/Jivko1212/b7cf9c2581af786862b7fd85075e302c

### DIFF
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -708,6 +708,33 @@ namespace NodaTime
             Preconditions.CheckNotNull(adjuster, nameof(adjuster)).Invoke(this);
 
         #region Formatting
+        private static Func<LocalDate, string> defaultToStringFormatter = localDate => LocalDatePattern.BclSupport.Format(localDate, null, CultureInfo.CurrentCulture);
+        private static Func<LocalDate, string, IFormatProvider, string> defaultIFormattableFormatter = (localDate, patternText, formatProvider) =>
+            LocalDatePattern.BclSupport.Format(localDate, patternText, formatProvider);
+        /// <summary>
+        /// Provides the option to override the default ToString behavior
+        /// </summary>
+        public static Func<LocalDate, string> DefaultToStringFormatter {
+            get => defaultToStringFormatter;
+            set {
+                if (value == null)
+                    throw new ArgumentNullException("Default formatter cannot be null");
+                else
+                    defaultToStringFormatter = value;
+            }  
+        }
+        /// <summary>
+        /// Provides the option to override the default IFormattable behavior
+        /// </summary>
+        public static Func<LocalDate, string, IFormatProvider, string> DefaultIFormattableFormatter {
+            get => defaultIFormattableFormatter;
+            set {
+                if (value == null)
+                    throw new ArgumentNullException("Default formatter cannot be null");
+                else
+                    defaultIFormattableFormatter = value;
+            }
+        }
         /// <summary>
         /// Returns a <see cref="System.String" /> that represents this instance.
         /// </summary>
@@ -715,7 +742,7 @@ namespace NodaTime
         /// The value of the current instance in the default format pattern ("D"), using the current thread's
         /// culture to obtain a format provider.
         /// </returns>
-        public override string ToString() => LocalDatePattern.BclSupport.Format(this, null, CultureInfo.CurrentCulture);
+        public override string ToString() => DefaultToStringFormatter(this);
 
         /// <summary>
         /// Formats the value of the current instance using the specified pattern.
@@ -731,7 +758,7 @@ namespace NodaTime
         /// </param>
         /// <filterpriority>2</filterpriority>
         public string ToString(string patternText, IFormatProvider formatProvider) =>
-            LocalDatePattern.BclSupport.Format(this, patternText, formatProvider);
+            defaultIFormattableFormatter(this, patternText, formatProvider);
         #endregion Formatting
 
         #region XML serialization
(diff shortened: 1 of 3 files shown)


jskeet:
This would behave differently to every other type in .NET. We're trying to fit in with the existing BCL approach.

I'd suggest either:

- Using patterns everywhere to be explicit
- Changing your thread's current culture appropriately (slightly tricky if you use `DateTime` anywhere else, admittedly)

I would really rather not mutable global state for this.

malcolmr:
Agreed re: the culture formatting.

However, making `LocalTime` output nanoseconds by default in what should be a debug string sounds sensible. @jskeet, is there some reason we didn't do that to start with?

Jivko1212:
That would be a welcome change .. right now LocalTime loses information and
I would have to inspect private members during debugging.

jskeet:
This comes back to "what is `ToString()` for?" It's not necessarily debug output. We basically use the default time format - in other words, whatever a `DateTime` would give, but without any date part.

If .NET separated "diagnostic output" from "general output" I'd agree.

jskeet:
One thing we *could* potentially do is have another interface similar to `DateTimeFormatInfo`, and an interface indicating that a type supports that. Then if you create a subclass of `CultureInfo` that implements the interface, we could pick up extra defaults. It's not terribly nice, but it might be a bit better...

malcolmr:
Yup, I got confused between Java and C#, sorry.

Changing the `ToString()` implementation would therefore require us to wait for a 3.0.

Creating an interface that could be used as a `CultureInfo` mixin (and therefore be set as the thread default culture) would be more reasonable.

Would you think we'd replicate how `DateTimeFormatInfo` works, or do something more direct like:

```csharp
public interface NodaTimeFormatInfo<T>
{
    IPattern<T> GetPattern();
}
```

(Not sure that would allow you to implement e.g. `public class MyCultureInfo : CultureInfo, NodaTimeFormatInfo<LocalDate>, NodaTimeFormatInfo<Period>`, but I think you get the idea.)

jskeet:
@Jivko1212: Any updates on what you want to do with this?

Jivko1212:
I think the best solution would be to allow a way to override the ToString
behavior via the app.config file. Many libraries like Akka, NLog allow this.

I am currently using my own Time/Date implementation for a finance project
and the nanosecond precision is very important. I don't like how .NET
forces localization on the developer. I think NodaTime is very elegant but
I decided to continue using my own, because my implementation has a lot of
coupling with everything else in the project. Especially, having to deal
with setting DefaultThreadCulture on each module is too much work and might
introduce bugs if I forget to set it. There is also one issue with setting
DefaultThreadCulture with time formatter that uses "HH:mm:ss.fffffffff". If
there are 9 f's it works perfectly with NodaTime (for nanoseconds), but
TimeSpans start throwing exceptions.

I had looked at NodaTime before when it wasn't nanosecond precision and
this was the main reason to write my own Time/Date. Right now I think
NodaTime is great for finance applications. Adding the ability to format to
nanoseconds by default would be ideal for finance.

Thanks

jskeet:
To be honest, I'd say the answer isn't to change the default - it's to avoid relying on the default at all. If you want a particular format, be explicit about that. We try to avoid global defaults wherever possible in Noda Time - that's why nothing uses the default time zone by default, and nothing uses the system time by default. You have to be really clear about what you want.

That said, I'm happy to look at what Akka and NLog do - do you have any links to details of that?

jskeet:
(Sorry, the close was accidental.)

Jivko1212:
You avoid defaults, but the BCL has annoying defaults (in my case and I am
sure in many others) which NodaTime is piggybacking on. I think being able
to configure a library is a good thing and makes it more usable. Sure you
can specify options on every call but it is not convenient. On top of that
the CultureInfo does not allow formatting that uses nanoseconds (or rather
breaks TimeSpan formatting). You are piggybacking on CultureInfo which
doesn't allow nanosecond formatting which is a major feature of your
library.

Here is a link how Akka handles configuration:
http://getakka.net/docs/concepts/configuration

They have their own hocon format, probably due to their java heritage. But
it can be made simpler with just XML.

jskeet:
Hmm. I'll keep thinking about it, but I'm mostly not convinced. It's one thing for general configuration - it's another for something as mostly-low-level as `ToString` to be changed like this.

As an example of how this could cause problems, consider the case where your application uses another library which expects Noda Time to use the default culture settings - but then the app.config settings could override that.

Will ponder. (Will also see how Akka goes about *getting* config settings. It would have to be applicable to .NET Core, of course.)

Jivko1212:
I guess it would be a breaking change in this case. If a library uses
NodaTime and ToString was configurable the library should use explicit
formatting.

Allowing more configuration options makes it easier to use for the end
user, but library authors have to do the extra calls and explicit
configuration. I think it is a good trade off. NodaTime is probably used by
end users in the majority of the time.

Also do you see it as a problem that CultureInfo doesn't play well with
nanoseconds, while your resolution is nanoseconds?

jskeet:
> NodaTime is probably used by end users in the majority of the time.

I wouldn't want to assume that, to be honest.

> Also do you see it as a problem that CultureInfo doesn't play well with
nanoseconds, while your resolution is nanoseconds?

It's certainly not ideal, but I'm not sure there's a particularly clean solution.

jskeet:
I'm going to close this pull request as I still don't think we want to mess with `ToString()` in this way.

PR #875, opened 2017-06-25
Added IWeekYearRule to 1.4.x.

jskeet:
This will allow us to make GetGregorian(int) etc obsolete while
providing 2.0.x-compatible alternatives.

The implementation and tests are as close to 2.0.x as I could make
them - they're not necessarily how I'd normally write them with 1.x,
but I wanted to make as few changes as possible. (Hence the multiple
conversions between ticks and days, because 2.0 is day-based.)

(I'm not sure if there's an easy way of comparing this code with the 2.0 equivalent other than downloading it...)

### DIFF
--- a/src/NodaTime/Calendars/WeekYearRules.cs
+++ b/src/NodaTime/Calendars/WeekYearRules.cs
@@ -1,0 +1,124 @@
+﻿// Copyright 2016 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using JetBrains.Annotations;
+using NodaTime.Extensions;
+using NodaTime.Utility;
+using System;
+using System.Globalization;
+using static System.Globalization.CalendarWeekRule;
+
+namespace NodaTime.Calendars
+{
+    /// <summary>
+    /// Factory methods to construct week-year rules supported by Noda Time.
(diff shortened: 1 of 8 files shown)


src/NodaTime/Calendars/WeekYearRules.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </para>
+        /// <para>
+        /// For example, January 1st 2011 was a Saturday, so only two days of that week
+        /// (Saturday and Sunday) were in 2011. Therefore January 1st is part of
+        /// week 52 of week-year 2010. Conversely, December 31st 2012 is a Monday,
+        /// so is part of week 1 of week-year 2013.
+        /// </para>
+        /// </remarks>
+        /// <value>A <see cref="IWeekYearRule"/> consistent with ISO-8601.</value>
+        [NotNull] public static IWeekYearRule Iso { get; } = new SimpleWeekYearRule(4, IsoDayOfWeek.Monday, false);
```
malcolmr: These are only `[NotNull]` in master/2.1. In 2.0.x, they're not (see 24fdeef).

Does that matter? (I suspect not, since 2.0.x would be a breaking change from 1.4.x anyway.)

jskeet: Nope, that'll be fine.

PR #877, opened 2017-06-25
Remaining 1.4 obsolete attributes

jskeet:
If I've got this right, this (with #875) should complete the 1.4 work, I think...

We're left with the following which have been removed in 2.0 but can't be made obsolete in 1.4:

- The `IsoDayOfWeek` property in various types, which has been renamed to `DayOfWeek` in 2.0, but there's already a `DayOfWeek` property of type `int`
- Overriding `Equals/GetHashCode` from `object` in `DateTimeZone`
- The `ZoneInterval` constructor, which will be source compatible as it now takes `Instant?` values instead of `Instant` (with a slight wrinkle that `Instant.MinValue` and` Instant.MaxValue` are now "regular" extreme values rather than sentinel values for "beginning and end of time"

Each commit is self-contained - I'd suggest reviewing one commit at a time.

### DIFF
--- a/src/NodaTime/TimeZones/DateTimeZoneSourceExtensions.cs
+++ b/src/NodaTime/TimeZones/DateTimeZoneSourceExtensions.cs
@@ -7,7 +7,7 @@ using System;
 namespace NodaTime.TimeZones
 {
     /// <summary>
-    /// Extension methods for <see cref="IClock"/> to enable migration to Noda Time 2.0.
+    /// Extension methods for <see cref="IDateTimeZoneSource"/> to enable migration to Noda Time 2.0.
     /// </summary>
     public static class DateTimeZoneSourceExtensions
     {
--- a/src/NodaTime/TimeZones/DateTimeZoneSourceExtensions.cs
+++ b/src/NodaTime/TimeZones/DateTimeZoneSourceExtensions.cs
@@ -1,0 +1,22 @@
+﻿// Copyright 2017 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.using System;
+using NodaTime.Utility;
+using System;
+
+namespace NodaTime.TimeZones
+{
+    /// <summary>
+    /// Extension methods for <see cref="IClock"/> to enable migration to Noda Time 2.0.
+    /// </summary>
+    public static class DateTimeZoneSourceExtensions
+    {
+        /// <summary>
+        /// Returns this source's ID for the system default time zone.
+        /// </summary>
+        /// <param name="source">The source to consult its ID for the system default time zone</param>
+        /// <returns>The ID for the system default time zone for this source, or null if the system default time zone has no mapping in this source.</returns>
+        public static string GetSystemDefaultId(this IDateTimeZoneSource source) =>
+            Preconditions.CheckNotNull(source, nameof(source)).MapTimeZoneId(TimeZoneInfo.Local);
+    }
+}
--- a/src/NodaTime/TimeZones/IDateTimeZoneSource.cs
+++ b/src/NodaTime/TimeZones/IDateTimeZoneSource.cs
@@ -100,6 +100,7 @@ namespace NodaTime.TimeZones
         /// The ID for the given system time zone for this source, or null if the system time
         /// zone has no mapping in this source.
         /// </returns>
+        [Obsolete("Only the system default time zone can be mapped in 2.0, using GetSystemDefaultId. For other time zones, use source-specific members.")]
         string MapTimeZoneId(TimeZoneInfo timeZone);
     }
 }
--- a/src/NodaTime/ZonedDateTime.cs
+++ b/src/NodaTime/ZonedDateTime.cs
@@ -194,12 +194,14 @@ namespace NodaTime
         /// so is part of week 1 of WeekYear 2013.
         /// </para>
         /// </remarks>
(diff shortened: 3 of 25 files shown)


src/NodaTime/ZonedDateTime.cs
```diff
@@ -378,6 +378,7 @@ public override int GetHashCode()
         /// <param name="lhs">First operand of the comparison</param>
         /// <param name="rhs">Second operand of the comparison</param>
         /// <returns>true if the <paramref name="lhs"/> is strictly earlier than <paramref name="rhs"/>, false otherwise.</returns>
+        [Obsolete("Use ZonedDateTime.Comparer.Instant to compare values by instant for compatibility with 2.0, and compare calendars and zones explicitly.")]
```
malcolmr: Is this "Use ... to compare values by instance [...], and [to] compare calendars and zones explicitly", or "Use ... to compare values by instance [...], and [you should] compare calendars and zones explicitly"?

src/NodaTime/ZonedDateTime.cs
```diff
@@ -441,6 +445,7 @@ public override int GetHashCode()
         /// <returns>A value less than zero if the instant represented by this zoned date/time is earlier than the one in
         /// <paramref name="other"/>; zero if the instant is the same as the one in <paramref name="other"/>;
         /// a value greater than zero if the instant is later than the one in <paramref name="other"/>.</returns>
+        [Obsolete("Use ZonedDateTime.Comparer.Instant to compare values by instant for compatibility with 2.0")]
```
malcolmr: Also should these two have the same ", and compare" text as above?

src/NodaTime/TimeZones/DateTimeZoneSourceExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright 2017 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.using System;
+using NodaTime.Utility;
+using System;
+
+namespace NodaTime.TimeZones
+{
+    /// <summary>
+    /// Extension methods for <see cref="IClock"/> to enable migration to Noda Time 2.0.
```
malcolmr: Not `IClock` here!

src/NodaTime/TimeZones/IDateTimeZoneSource.cs
```diff
@@ -100,6 +100,7 @@ public interface IDateTimeZoneSource
         /// The ID for the given system time zone for this source, or null if the system time
         /// zone has no mapping in this source.
         /// </returns>
+        [Obsolete("Only the system default time zone can be mapped in 2.0, using GetSystemDefaultId. For other time zones, use source-specific members.")]
```
malcolmr: I wonder if this should say "the GetSystemDefaultId extension method"? (Likewise elsewhere)

src/NodaTime/ZonedDateTime.cs
jskeet: Update the comment to:

> Use ZonedDateTime.Comparer.Instant to compare values by instant for compatibility with 2.0. If calendar and time zone comparison is required, make those comparisons explicitly.

Does that cover it?

src/NodaTime/ZonedDateTime.cs
jskeet: No, because they were already only comparing by instant. The extra part of the comment for the operators was because comparing by instant wouldn't provide the exact same behaviour.

src/NodaTime/TimeZones/DateTimeZoneSourceExtensions.cs
jskeet: Fixed.

src/NodaTime/TimeZones/IDateTimeZoneSource.cs
jskeet: Well, in 2.0 it's not an extension method - and the first half of the sentence implies that it's talking about 2.0. I'd prefer to leave it slightly woolly to cover both 1.4 and 2.0.

jskeet:
Reasoning about IsoDayOfWeek...

If we make IsoDayOfWeek obsolete, we'd want to either say that it wasn't available in 2.0 or provide a compatible equivalent now... but we can't provide a compatible equivalent because that's DayOfWeek, which already exists with a different type.

We want users to fix their code until they have no warnings - that's the best they can do until they go to 2.0. We want them to change their code *after* moving to 2.0 by changing everywhere they were referring to IsoDayOfWeek to refer to DayOfWeek instead. Now arguably they *could* do that with a cast now, but that's a really ugly workaround IMO.

We could *potentially* make DayOfWeek obsolete, but we don't really want users to move away from it - we just want them to add a cast when they've moved to 2.0 if they really need the value as an `int`.

jskeet:
(Added a commit with the changes from code review rather than trying to backdate the fixes...)

src/NodaTime/ZonedDateTime.cs
malcolmr: SGTM, thanks!

src/NodaTime/TimeZones/IDateTimeZoneSource.cs
malcolmr: Well, q.v. `IClock.Now`?

(I'm nitpicking, though. This is fine as-is.)

malcolmr:
LGTM.

jskeet:
Merging this now, although it won't actually make full sense until #875 is in.

PR #884, opened 2017-07-04
Release build script for 1.4 branch

jskeet:
We don't build an offline user guide any more, or run through the Sandcastle XML doc stripper... so it's fairly simple now.

The new script was mostly adapted from the tzdb update script for 1.3 in master.


jskeet:
Just checked, and there's no PDB in NodaTime-1.3.1.zip.

PR #899, opened 2017-07-07
Introduce LocalDate.Max/Min/MaxIsoValue/MinIsoValue

### LINKED ISSUE #898: Add LocalDate.{Min,Max}Value
Much like #356 but for LocalDate.

jskeet:
Fixes #898.


jskeet:
We have Min/Max for Instant - indeed, that's where I copied the implementation from :) It can definitely be useful to be able to say "get the later value".

Maybe I should remove them from this PR, but start a new post-2.1 PR that adds it for other types too? (Anything that implements IComparable, basically).

malcolmr:
That sounds sensible, yes.

jskeet:
Done, PTAL.

PR #915, opened 2017-07-27
Provide Min/Max for all suitable types

### LINKED ISSUE #900: Introduce Min/Max methods for all appropriate types
We already have this for Instant and Offset. We should expand this to other suitable types:

- LocalDateTime (same calendar only)
- LocalDate (same calendar only)
- LocalTime
- Duration

Probably *not* OffsetDateTime and ZonedDateTime, for the same reasons those don't implement IComparable

jskeet:
- Duration
- LocalDate
- LocalDateTime
- LocalTime

These join Offset and Instant that already have the same methods.

OffsetDateTime and ZonedDateTime don't have these as they make less
sense; this mirrors which types have `<` and `>` operators.

Fixes #900.

### DIFF
--- a/src/NodaTime/Duration.cs
+++ b/src/NodaTime/Duration.cs
@@ -1190,6 +1190,35 @@ namespace NodaTime
         [Pure]
         internal decimal ToDecimalNanoseconds() => IsInt64Representable ? ToInt64NanosecondsUnchecked() : ((decimal)days) * NanosecondsPerDay + nanoOfDay;
 
+        /// <summary>
+        /// Returns the larger duration of the given two.
+        /// </summary>
+        /// <remarks>
+        /// A "larger" duration is one that advances time by more than a "smaller" one. This means
+        /// that a positive duration is always larger than a negative one, for example. (This is the same
+        /// comparison used by the binary comparison operators.)
+        /// </remarks>
+        /// <param name="x">The first duration to compare.</param>
+        /// <param name="y">The second duration to compare.</param>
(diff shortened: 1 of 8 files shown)


src/NodaTime/Duration.cs
```diff
@@ -1190,6 +1190,25 @@ void IXmlSerializable.WriteXml([NotNull] XmlWriter writer)
         [Pure]
         internal decimal ToDecimalNanoseconds() => IsInt64Representable ? ToInt64NanosecondsUnchecked() : ((decimal)days) * NanosecondsPerDay + nanoOfDay;
 
+        /// <summary>
+        /// Returns the larger duration of the given two.
```
malcolmr: Consider "longer"/"shorter" here throughout instead of "larger"/"smaller"?

jskeet: I was wondering how to express that. The trouble is that "longer/shorter" sound more like absolute values. Consider two durations: -2 hours and 1 hour. We want to express that 1 hour is "more than" -2 hours. I'm not sure whether either larger or longer cut it there. Any other ideas? (If not, I'm happy to just change it to longer/shorter and maybe add a remark.)

malcolmr: Argh, and I thought about negative durations too.

Yes, you're right. Larger/smaller is probably clearer even if it's a little awkward.

jskeet: Righto - will leave that terminology, but add a remark to each method to make it clear.

PR #924, opened 2017-08-07
Introduce smoke tests and web site autobuild script

### LINKED ISSUE #916: Automate web site build
The build procedure for the web site is now pretty stable. We should automate it - one of my benchmark machines should be okay to do that, either using a CI setup or just checking in a scheduled task (as we do for benchmarks).

jskeet:
I will schedule autobuildweb.sh to be run once every half hour
(excluding the benchmark period if I can). With the smoke tests as
part of the buildweb.sh script, we can be reasonably confident that
this won't break the site...

Fixes #916

### DIFF
--- a/build/autobuildweb.sh
+++ b/build/autobuildweb.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# This script is designed to be run on a scheduled basis, to build and deploy the web site
+# in an automated manner. It takes the following steps, with input as a
+# "web site build root directory" (henceforth known as $root).
+#
+# - Find the latest commit for the main repo
+# - If $root/$commit exists, assume there have been no changes, and abort
+# - Make a shallow clone (--depth 1) of the main repo to $root/$commit/nodatime
+# - Make a shallow clone (--depth 1) of the nodatime.org repo to $root/$commit/nodatime.org
+# - Run $root/$commit/build/buildweb.sh $root/$commit/nodatime.org
+# - In $root/$commit/nodatime.org
+#   - Create a new commit
+#   - Push to nodatime.org repo
+
+set -e
+
+if [[ "$1" = "" ]]
+then
+  echo "Usage: autobuildweb.sh output-directory"
+  echo "e.g. autobuildweb.sh c:\\users\\jon\\NodaTimeWeb"
+  exit 1
+fi
+
+declare -r root=$1
+declare -r commit=$(curl -H Accept:application/vnd.github.VERSION.sha https://api.github.com/repos/nodatime/nodatime/commits/master)
+
+if [[ -d $root/$commit ]]
+then
+  echo "Directory $root/$commit already exists. Aborting."
+  exit 0
+fi
+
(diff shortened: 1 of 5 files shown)


jskeet:
(My plan is to get this in and the scheduled task set up, then edit the roadmap and watch it magically get published...)

build/autobuildweb.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "Directory $root/$commit already exists. Aborting."
+  exit 0
+fi
+
+# Clone repos
+git clone https://github.com/nodatime/nodatime.git --depth 1 $root/$commit/nodatime
+git clone https://github.com/nodatime/nodatime.org.git --depth 1 $root/$commit/nodatime.org
+
+# Build site and run smoke tests
+(cd $root/$commit/nodatime/build; ./buildweb.sh ../../nodatime.org)
```
malcolmr: This runs in a subshell, as does the smoketest in `buildweb.sh`. Have you tested whether a smoke-test failure will actually abort this script? (e.g. by making the smoketest fail, then running this.)

src/NodaTime.Web/Program.cs
```diff
@@ -1,12 +1,35 @@
-﻿using Microsoft.AspNetCore.Hosting;
+﻿// Copyright 2017 The Noda Time Authors. All rights reserved.
```
malcolmr: 2016. This was introduced in 522cf028.

src/NodaTime.Web/Program.cs
```diff
@@ -16,8 +39,77 @@ public static void Main(string[] args)
...
+            await ExpectPageText("1.0.x/userguide/text", "There are two options for text handling");
+            await ExpectPageText("1.0.x/api/NodaTime.DateTimeZone.html", "The mapping is unambiguous");
+            await ExpectPageText("tzdb/index.txt", "http://nodatime.org/tzdb/tzdb2013h.nzd");
+            await ExpectBinarySize("tzdb/tzdb2013h.nzd", 125962);
+        }
+
+        private static async Task ExpectPageText(string relativeUrl, string expectedText)
+        {
+            var client = new HttpClient();
+            string page = await client.GetStringAsync($"{TestUrlBase}/{relativeUrl}");
```
malcolmr: Any idea what this does on a non-200 response?

src/NodaTime.Web/Program.cs
jskeet: Fixed.

src/NodaTime.Web/Program.cs
jskeet: Just checked, and we get a message of:

```text
Failure: Response status code does not indicate success: 404 (Not Found).
```

That doesn't show which URL was hit, but as the previous successes *are* reported, it's not too bad.

build/autobuildweb.sh
jskeet: Just checked, and it does.

PR #929, opened 2017-08-12
A mixture of changes

### LINKED ISSUE #905: Website considers 2.1.0-beta01 as newer than 2.1.0
See http://nodatime.org/downloads/. The betas released for 2.1.0 and 1.4.0 appear before the unsuffixed final release, even though the date is later.

(I think this is why the front page still lists the beta as "latest".)

Sorting by version number does make sense, but the beta/RC releases should precede the final releases.

jskeet:
I can break these out into separate changes if necessary...

- Move some web classes into a Helpers directory
- Add a VersionComparer to fix #905
- Add a test project for the web site
- Move Travis script into its own shell script in build, like AppVeyor
- Run web site tests on AppVeyor and Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,4 @@ dotnet: 1.0.1
 dist: trusty
 
 script:
-  - dotnet restore src/NodaTime-All.sln
-  - dotnet build src/NodaTime/*.csproj -f netstandard1.3
-  - dotnet build src/NodaTime.Testing/*.csproj -f netstandard1.3
-  - dotnet build src/NodaTime.Benchmarks/*.csproj -f netcoreapp1.1
-  - dotnet build src/NodaTime.Demo/*.csproj -f netcoreapp1.0
-  - dotnet build src/NodaTime.Test/*.csproj -f netcoreapp1.0
-  - dotnet run -p src/NodaTime.Test/*.csproj -f netcoreapp1.0 -- --where=cat!=Slow
+  - build/travis.sh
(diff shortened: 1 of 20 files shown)


src/NodaTime.Web.Test/Helpers/VersionComparerTest.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        [Test]
+        [TestCase("1.0.0", "1.0.1")]
+        [TestCase("1.0.0", "1.1.0")]
+        [TestCase("1.1.0", "2.0.0")]
+        [TestCase("1.0.0", "1.0.1-alpha01")]
+        [TestCase("1.0.0-alpha01", "1.0.0-alpha02")]
+        [TestCase("1.0.0-alpha01", "1.0.0-beta01")]
+        [TestCase("1.0.0-alpha01", "1.0.0")]
+        public void Compare(string earlier, string later)
```
malcolmr: Just for completeness, add some pairs that should be equivalent? (which I think means that the strings need to be equal?)

Also perhaps include something with a two-digit major/minor/patch number? (even if it's to document that that comparison doesn't work properly).

jskeet: Good plan, will add an extra test for equivalent versions.

Will actually fix it for two-digit major/minor/patch, otherwise it'll bug me :)

jskeet:
Okay, completely redone this PR now, I'm afraid - but it should be easier to review now that it's in four commits :)

PR #930, opened 2017-08-14
Move to .NET Core 2 SDK

jskeet:
This will probably fail, as I'm pretty sure AppVeyor hasn't got VS 15.3, and I doubt that Travis has 2.0 yet. But it's nice to have it ready to try...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,18 @@
 # TODO: Test against Mono as well, using matrix.include
 language: csharp
 mono: none
-dotnet: 1.0.1
+dotnet: 2.0.0
 dist: trusty
+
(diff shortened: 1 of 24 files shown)


jskeet:
Will tidy up the commits when we know exactly what we want to do.

It will fail on AppVeyor until we have .NET Core SDK 2.0 support there, but AppVeyor was mostly present for coverage - which isn't working at the moment anyway.

Thoughts welcome.

jskeet:
And now AppVeyor works too :) Will tidy this PR up...

jskeet:
Okay, refactored this PR into sensible commits. Will merge on green - I then need to modify the scheduled tasks to run benchmarks on .NET Core 2.

PR #954, opened 2017-08-20
Reinstate code coverage using dotCover and codecov

jskeet:
(Will modify github readme badge when we *have* some coverage.)

### DIFF
--- a/build/appveyor.sh
+++ b/build/appveyor.sh
@@ -3,17 +3,20 @@
 # Build script run by Appveyor. Might eventually be less
 # single-purpose, but let's get coverage going ASAP...
 
-cd $(dirname $0)/..
+set -e
+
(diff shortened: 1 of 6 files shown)


build/appveyor.sh
```diff
@@ -3,17 +3,20 @@
 # Build script run by Appveyor. Might eventually be less
 # single-purpose, but let's get coverage going ASAP...
 
-cd $(dirname $0)/..
```
malcolmr: Doesn't removing this mean that this script is now sensitive to `pwd`? (For the `dotnet` invocations below, e.g.)

jskeet: Ah, true. Will reinstate that.

jskeet: (Not that it matters given that it's always actually going to be run from the root :)

PR #958, opened 2017-08-22
Fix LocalInstant.SafeMinus near start/end of time

### LINKED ISSUE #957: LocalInstant.SafeMinus gives incorrect result around the start/end of time
The call to PlusSmallNanoseconds should be MinusSmallNanoseconds.

Discovered while adding test coverage. The fix is trivial - we should probably backport to the 2.2 branch, but it's sufficiently corner-casey to mean it's probably not worth doing a 2.2.1 build just for that.

jskeet:
Fixes #957.


jskeet:
Will cherry pick into the 2.2 branch when I'm not on the beach...

PR #961, opened 2017-08-24
Improve code coverage significantly

jskeet:
(Some more to come, but this is a good start.)

Separated into easy-to-read commits, but I'll self-review.


jskeet:
(Pushed without the changes to test code, just to make sure everything still passes before the test changes.)

PR #963, opened 2017-08-25
Coverage again

jskeet:
The first commit probably merits review, if you have time @malcolmr.
The rest are less controversial.

Most of the remaining uncovered code is mostly around time zones, and I have a plan for those...

### DIFF
--- a/src/NodaTime.Test/TestHelper.cs
+++ b/src/NodaTime.Test/TestHelper.cs
@@ -11,7 +11,9 @@ using System.Xml.Serialization;
 using NUnit.Framework;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Serialization;
 using NodaTime.Annotations;
+using NodaTime.TimeZones;
 
 #if !NETCORE
 using System.Runtime.Serialization.Formatters.Binary;
@@ -505,6 +507,20 @@ namespace NodaTime.Test
             Assert.AreEqual(value, rehydrated);
 #endif
         }
+
+#if !NETCORE
+        internal static void AssertBinaryDeserializationFailure<T>(Type expectedExceptionType, Action<SerializationInfo> fillInfoAction)
+        {
+            var info = new SerializationInfo(typeof(T), new FormatterConverter());
+            fillInfoAction(info);
+            var ctor = typeof(T).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance, null,
+                new Type[] { typeof(SerializationInfo), typeof(StreamingContext) }, null);
+            Assert.NotNull(ctor);
+            var context = new StreamingContext();
+            var exception = Assert.Throws<TargetInvocationException>(() => ctor.Invoke(new object[] { info, context }));
+            Assert.IsInstanceOf(expectedExceptionType, exception.InnerException);
+        }
+#endif
 
         /// <summary>
         /// Validates that a value can be serialized to the expected XML, deserialized to an equal
--- a/src/NodaTime/CalendarSystem.cs
+++ b/src/NodaTime/CalendarSystem.cs
@@ -10,6 +10,7 @@ using JetBrains.Annotations;
 using NodaTime.Annotations;
 using NodaTime.Calendars;
 using NodaTime.Utility;
+using System.Diagnostics.CodeAnalysis;
 
 namespace NodaTime
 {
@@ -602,6 +603,9 @@ namespace NodaTime
         /// </summary>
         /// <param name="yearMonthDay">The value to validate.</param>
         [Conditional("DEBUG")]
+#if NET45
+        [ExcludeFromCodeCoverage]
+#endif
         internal void DebugValidateYearMonthDay(YearMonthDay yearMonthDay)
         {
             // Avoid the line even being compiled in a release build...
--- a/src/NodaTime/Text/ValueCursor.cs
+++ b/src/NodaTime/Text/ValueCursor.cs
@@ -294,7 +294,7 @@ namespace NodaTime.Text
         {
             unchecked
             {
-                Preconditions.DebugCheckArgument(maximumDigits <= scale , nameof(maximumDigits),
+                Preconditions.DebugCheckArgument(maximumDigits <= scale, nameof(maximumDigits),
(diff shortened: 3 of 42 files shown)


src/NodaTime/TimeZones/ZoneRecurrence.cs
```diff
@@ -129,27 +129,17 @@ public bool Equals(ZoneRecurrence other)
             else if (safeLocal >= maxLocalInstant)
             {
                 // Asked for a transition after our final transition... or both are beyond the end of time (in which case
-                // we can return an infinite transition)
-                return maxLocalInstant == LocalInstant.AfterMaxValue ? new Transition(Instant.AfterMaxValue, newOffset) :  (Transition?) null;
+                // we can return an infinite transition). THis branch will always be taken for transitions beyond the end
```
malcolmr: typo: THis.

malcolmr: Also, while you're here, there's an odd construction I can't parse on L22 with "before that calendars".

src/NodaTime/TimeZones/ZoneRecurrence.cs
```diff
@@ -167,11 +157,9 @@ public bool Equals(ZoneRecurrence other)
             }
 
             // We've got a transition earlier than we were asked for. Try next year.
+            // Note that this will stil be within the FromYear/ToYear range, otherwise
```
malcolmr: This seems correct to me, though I'll admit that I'm more convinced by our existing test coverage.

src/NodaTime/Text/ValueCursor.cs
```diff
@@ -282,20 +284,18 @@ internal bool ParseDigits(int minimumDigits, int maximumDigits, out int result)
...
-        internal bool ParseFraction(int maximumDigits, int scale, out int result, int minimumDigits)
+        internal bool ParseFraction([Trusted] int maximumDigits, int scale, out int result, int minimumDigits)
         {
             unchecked
             {
-                if (scale < maximumDigits)
-                {
-                    scale = maximumDigits;
-                }
+                Preconditions.DebugCheckArgument(maximumDigits <= scale , nameof(maximumDigits),
```
malcolmr: nit: remove space-before-comma?

src/NodaTime.Test/TestHelper.cs
```diff
@@ -506,6 +508,23 @@ internal static void AssertBinaryRoundtrip<T>(T value)
...
+        {
+            var info = new SerializationInfo(typeof(T), new FormatterConverter());
+            fillInfoAction(info);
+            var ctor = typeof(T).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance, null,
+                new Type[] { typeof(SerializationInfo), typeof(StreamingContext) }, null);
+            Assert.NotNull(ctor);
+            var context = new StreamingContext();
+            var exception = Assert.Throws<TargetInvocationException>(() => ctor.Invoke(new object[] { info, context }));
+            Assert.That(exception.InnerException, 
+                Is.InstanceOf<OverflowException>()
```
malcolmr: This seems prone to us testing for one exception but getting another by mistake. Could we specify the expected exception at the callsites?

(possibly consider this as a followup, since what we have here is strictly better than what we already had.)

src/NodaTime/CalendarSystem.cs
```diff
@@ -602,6 +603,9 @@ internal Era GetEra([Trusted] int absoluteYear)
         /// </summary>
         /// <param name="yearMonthDay">The value to validate.</param>
         [Conditional("DEBUG")]
+#if NET45
+        [ExcludeFromCodeCoverage]
+#endif
         internal void DebugValidateYearMonthDay(YearMonthDay yearMonthDay)
         {
             // Avoid the line even being compiled in a release build...
```
malcolmr: (Incidentally, what does this mean?)

malcolmr: (Incidentally, what does this mean?)

src/NodaTime/CalendarSystem.cs
```diff
@@ -602,6 +603,9 @@ internal Era GetEra([Trusted] int absoluteYear)
         /// </summary>
         /// <param name="yearMonthDay">The value to validate.</param>
         [Conditional("DEBUG")]
+#if NET45
```
malcolmr: and `!DEBUG`?

src/NodaTime/CalendarSystem.cs
jskeet: The attribute isn't available in netstandard, but that's okay, because we only run coverage tests under net45. It's used in the coverage filter to detect methods we don't record coverage for.

src/NodaTime.Test/TestHelper.cs
jskeet: Can do, will just need to refactor a bit.

src/NodaTime/TimeZones/ZoneRecurrence.cs
jskeet: Not sure what you mean by the latter part - do you mean you're convinced by me being able to cover the lines I've now removed?

src/NodaTime/TimeZones/ZoneRecurrence.cs
jskeet: Yup, will fix that and the typo, thanks.

src/NodaTime/Text/ValueCursor.cs
jskeet: Thanks, will fix.

src/NodaTime/CalendarSystem.cs
jskeet: No need - we don't run the coverage tests under debug, and if we did (and ran not under net45) it would hit anyway :)

jskeet:
Nit fixed, and invalid binary serialization tests now show the exception that's thrown - I don't think it's desperately important that we harmonize those exceptions, but it's nice to know what they are.

Will merge on green, and we can talk about the ZoneRecurrence aspect more later.

PR #965, opened 2017-08-28
More coverage of time zone code

jskeet:
The important change introduced here is an interceptor for TimeZoneInfo calls.
Is this too heavy-weight given that there aren't many tests that rely on it?
(Many of the tests introduced in this PR *don't* depend on the introduction of the interceptor. Let me know if you'd like me to split them out into separate commits.)

This PR also fixes a bug where if you ask for the system default time zone in an environment where TimeZoneInfo.Local returns null, we blew up with a NullReferenceException instead of a DateTimeZoneNotFoundException. I think it's sufficiently corner-casey that we don't need to patch 2.2.

The remaining uncovered lines are pretty much all for time zone IO. We have the time zone *reading* code in NodaTime.dll, but the *writing* code is all in NodaTime.TzdbCompiler. We could move just FieldCollection and Field into NodaTime.dll, at which point it would be easier to test TzdbDateTimeZoneSource.Validate and TzdbStreamData.

### DIFF
--- a/src/NodaTime.Test/Extensions/ClockExtensionsTest.cs
+++ b/src/NodaTime.Test/Extensions/ClockExtensionsTest.cs
@@ -3,8 +3,10 @@
 // as found in the LICENSE.txt file.
 
 using NodaTime.Extensions;
+using NodaTime.Test.TimeZones;
 using NodaTime.Testing;
 using NUnit.Framework;
+using System;
 
 namespace NodaTime.Test.Extensions
 {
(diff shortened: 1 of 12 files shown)


jskeet:
We don't find out benchmark changes until it's merged, but these are rarely-used calls, and the delegation is trivial. `TimeZoneInfo.Local` is probably the most frequently-called one (as the `DateTimeZoneCache` avoids calling other things more often). Certainly the work in here is trivial compared with what we do with the results of each call :)

Issue #649, opened 2017-02-16
Intervals/ranges in NodaTime 2.0


lundmikkel:
NodaTime has a wonderful set of time and date related data types. However, it generally lacks support for intervals/ranges. NodaTime 1.x only contains `Interval` (a `Instant` interval), whereas NodaTime 2.0 seems to introduce a `DateInterval` as well. Most types of intervals are not supported in the library, e.g. `DateTimeInterval` and `TimeInterval`.

Neither `Interval` nor `DateInterval` support [interval relations](https://en.wikipedia.org/wiki/Allen's_interval_algebra). There are a lot of helper methods that are rather trivial, which are tedious to implement for each user of the library. Conversion methods would also be nice, allowing one type of interval to be converted into another, e.g. for converting a `DateTimeInterval` into a `ZoneDateTimeInterval`/`Interval`.

Are there any plans to support intervals in general in the future? E.g. with more interval data types, helper methods or even data structures. This could either be done within NodaTime, or as a separate project containing the existing interval types.

jskeet:
Each potential interval type would really need to justify its own existence. Not everything makes sense - for example, an interval from one ZonedDateTime to one in a different time zone isn't terribly useful, I suspect. I really think Interval and DateInterval will cover most cases. TimeInterval is a possibility, but with no plans right now - I'd want to see various use cases and a design that handles them nicely.

Interval relation methods *could* be useful - and pull requests are welcome, of course. It becomes slightly tricky when normal Interval values may or may not have a start/end, and DateIntervals can be exclusive or inclusive of end dates.

lundmikkel:
Yeah, ZonedDateTimeInterval might be a bad example (the DateTimeInterval could be converted directly into an Interval though). I work with scheduling and find that we often juggle pairs of values, which are actually a start and end of an interval. The values are given in local times by the user, but converted into global times. Having an interval from the beginning to the end would make it a lot easier to work with.

Relations are not really difficult to make. I have previously made an extensive interval library (with generic intervals), which had plenty of [extension methods](https://github.com/lundmikkel/C5Intervals/blob/master/C5.Intervals/Extensions/IntervalExtensions.cs) which worked with any combination of endpoint inclusions.

danygarss:
[...]

PR #977, opened 2017-10-02
Add Interval.Intersects method

mattjohnsonpint:
It is sometimes useful to see if an interval overlaps with another interval.  The math is straightforward, but somewhat non-intuitive.  Thus a built-in function seems appropriate.

### DIFF
--- a/src/NodaTime/Interval.cs
+++ b/src/NodaTime/Interval.cs
@@ -155,6 +155,14 @@ namespace NodaTime
         [Pure]
         public bool Contains(Instant instant) => instant >= start && instant < end;
 
+        /// <summary>
+        /// Returns whether or not this interval intersects with the given interval.
+        /// </summary>
+        /// <param name="interval">Interval to test.</param>
+        /// <returns>True if this interval intersects with the given interval; false otherwise.</returns>
+        [Pure]
+        public bool Intersects(Interval interval) => interval.start < end && interval.end > start;
+
         #region Implementation of IEquatable<Interval>
         /// <summary>
         /// Indicates whether the value of this interval is equal to the value of the specified interval.
--- a/src/NodaTime.Test/IntervalTest.cs
+++ b/src/NodaTime.Test/IntervalTest.cs
@@ -271,5 +271,25 @@ namespace NodaTime.Test
             // This would have been true under Noda Time 1.x
             Assert.IsFalse(interval.Contains(instant));
         }
+
+        [Test]
+        [TestCase("2000-01-01T00:00:00Z", "2020-01-01T00:00:00Z", "2020-01-01T00:00:00Z", "2040-01-01T00:00:00Z", ExpectedResult = false, Description = "First interval before second interval")]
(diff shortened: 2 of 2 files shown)


src/NodaTime/Interval.cs
```diff
@@ -155,6 +155,14 @@ public Instant End
         [Pure]
         public bool Contains(Instant instant) => instant >= start && instant < end;
 
+        /// <summary>
+        /// Returns whether or not this interval intersects with the given interval.
+        /// </summary>
+        /// <param name="interval">Interval to test.</param>
+        /// <returns>True if this interval intersects with the given interval; false otherwise.</returns>
+        [Pure]
+        public bool Intersects(Interval interval) => interval.start < end && interval.end > start;
```
jskeet: Would it be useful to change this to an `Intersection` method that returns a `Nullable<Interval>` I wonder?

jskeet (commented):
Implementation is fine, but I'm uncertain about how best to expose it.

malcolmr:
This is a good idea (and note that due to `start` and `end` not being exposed directly, it would be harder to implement in user code). However, I think it needs some work around empty intervals.

At present, two empty intervals (even those equal to one another) do not intersect. This is reasonable, especially if you consider the definition of intersect to be "contain an instant in common", as empty intervals contain no instants.

However, a non-empty interval will intersect with all empty intervals originating at [start+1, end), which seems odd (both because of the +1, and because of the inconsistency with `Contains()`). Would it make sense to filter out all empty intervals on both sides? (And, if so, to document the semantics of `Intersect` as based on the `Contains()` relation, and note what this means for empty intervals?)

Also, more test coverage around empty/half-open/infinite intervals would be nice.

src/NodaTime/Interval.cs
malcolmr: If you're going to add that, it's probably worth adding union and perhaps difference/symmetric difference.

jskeet:
I was going to check for prior art in the java.time API, but that doesn't even *have* interval, interestingly. Joda Time has `overlap` and `gap`, but it's not clear what `overlap` does in the face of an empty interval.

If we change to use `Intersection` returning an `Interval?` would we want to make `nonempty.Intersection(empty)` return `empty` if the start/end time of `empty` is within `nonempty`, and null otherwise?

jskeet:
See also #649.

malcolmr:
@jskeet I think it would be reasonable either to define intersection to be based on "intervals contain an instant common to both" _or_ on "ranges covered by the instants overlap" (where empty intervals have a defined empty range), just as long as we document it.

jskeet:
This is really old, but I've been thinking it would be good to clear up the repo in terms of both PRs and issues.

I think I'm most in favour of creating Intersection and Union methods over this. (And possibly Difference as well.) @mattjohnsonpint I know it's been years, but any thoughts on that?

mattjohnsonpint:
Hey, sorry but I haven't given much thought to this project in a long while.  I'd say do whatever you think is best.

PR #980, opened 2017-10-04
Fix Instant - Duration not performing validation

### LINKED ISSUE #979: Instant bounds checking bypassed
This should throw an exception, but doesn't:

```csharp
Instant.MinValue.Minus(Duration.FromDays(1))
```

jskeet:
Fixes #979.

### DIFF
--- a/src/NodaTime.Test/InstantTest.Operators.cs
+++ b/src/NodaTime.Test/InstantTest.Operators.cs
@@ -3,6 +3,7 @@
 // as found in the LICENSE.txt file.
 
 using NUnit.Framework;
+using System;
 
 namespace NodaTime.Test
 {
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/InstantTest.Operators.cs
```diff
@@ -81,6 +82,12 @@ public void OperatorPlusOffset_Zero_IsNeutralElement()
             Assert.AreEqual(new LocalInstant(0, 1L), one.Plus(Offset.Zero), "Instant(1) + Offset.Zero");
             Assert.AreEqual(new LocalInstant(0, NodaConstants.NanosecondsPerHour), NodaConstants.UnixEpoch.Plus(Offset.FromHours(1)), "UnixEpoch + offsetOneHour");
         }
+
+        [Test]
+        public void OperatorPlus_OutOfRange()
+        {
+            Assert.Throws<OverflowException>(() => (Instant.MaxValue + Duration.Epsilon).GetHashCode());
```
malcolmr: Obvious answer, I'm sure, but why the `GetHashCode()`?

jskeet: Because `Instance.MaxValue + Duration.Epsilon` isn't a statement expression, so the lambda expression can't be converted to `Action`. (It's a shame `Assert.Throws` doesn't have an overload accepting a `Func<T>`.)

PR #982, opened 2017-10-05
Be more persistent when parsing time zones

### LINKED ISSUE #981: Parse error, ZonedDateTime
This test fails:

```fsharp
let tzdb: IDateTimeZoneProvider =
  NodaTime.DateTimeZoneProviders.Tzdb

// ...
      testCase "parse failing zoned date time" <| fun _ ->
        let p = NodaTime.Text.ZonedDateTimePattern.ExtendedFormatOnlyIso.WithZoneProvider tzdb
        let res = p.Parse "1906-08-29T20:58:32 Etc/GMT-12 (+12)"

        if not res.Success then
          Tests.failtestf
            "Should successfully parse generated zoned date time, but got %s"
            res.Exception.Message
```

With this error message:

> [20:43:33 ERR] serialisation/base types/parse failing zoned date time failed in 00:00:00.0350000.
Should successfully parse generated zoned date time, but got The value string does not match a simple character in the format string " ". Value being parsed: '1906-08-29T20:58:32 Etc/GMT-1^2 (+12)'. (^ indicates error position.)

The value was generated thus:

```fsharp
type Arbs =
  static member Instant () =
    let genInst = gen {
      let! y = Gen.choose(1900,2100)
      let! m = Gen.choose(1, 12)
      let! d = Gen.choose(1, System.DateTime.DaysInMonth(y, m))
      let! h = Gen.choose(0,23)
      let! min = Gen.choose(0,59)
      let! sec = Gen.choose(0,59)
      return Instant.FromUtc(y, m, d, h, min, sec)
    }
    Arb.fromGen genInst

  static member Timezone() =
    gen {
      let! tz = Gen.choose (0, tzdb.Ids.Count - 1)
      return tzdb.Item tzdb.Ids.[tz]
    }
    |> Arb.fromGen

  static member ZonedDateTime() =
    gen {
      let! tz = Arb.generate<_>
      let! inst = Arb.generate<_>
      return ZonedDateTime(inst, tz)
    }
    |> Arb.fromGen
```

Version of things:

macOS 10.12.6, mono 5.2.0.215, NodaTime 1.4

More failing test strings:

 - `The value string does not match a simple character in the format string " ". Value being parsed: '1994-02-09T09:02:33 Etc/GMT-1^1 (+11)'. (^ indicates error position.)`
 - ...

jskeet:
We were previously stopping too early in the search.

Fixes #981. Should probably be backported to 1.4.x and 2.2.x.

Note for reviewers: would it be worth splitting this into an internal method that could then be tested more thoroughly? I'm pretty confident in it now, but presumably I was confident before, too...

### DIFF
--- a/src/NodaTime.Test/Text/ZonedDateTimePatternTest.cs
+++ b/src/NodaTime.Test/Text/ZonedDateTimePatternTest.cs
@@ -7,6 +7,7 @@ using NodaTime.Text;
 using NodaTime.TimeZones;
 using NUnit.Framework;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 
 namespace NodaTime.Test.Text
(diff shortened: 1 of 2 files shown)


jskeet:
A fake time zone provider would let us hit all the spots, yes - although the coverage side would be fragile to any further implementation change of the binary search, for example. Will have a go and add a new commit to this PR when I get a chance.

jskeet:
I've added the extra test into the single commit instead; will merge on green.

PR #987, opened 2017-10-14
Version history update

jskeet:


src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -4,6 +4,21 @@ User-visible changes from 1.0.0-beta1 onwards. See the
 [project repository](https://github.com/nodatime/nodatime) for more
 details.
 
+## 2.2.1, released 2017-10-14 with tzdb2017b
+
+Bug-fix release:
+
+- Fixes [issue 975]: `LocalInstant.SafeMinus` gives incorrect result around the start/end of time
```
malcolmr: Since `LocalInstant` isn't public, is there a way to describe what the end-user of this was, if any? (I was just having a look myself, and it looks like perhaps it was mostly theoretical unless you had crazy zone transitions?)

src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -4,6 +4,21 @@ User-visible changes from 1.0.0-beta1 onwards. See the
...
+- Fixes [issue 971]: Clarify exception for unknown standard patterns to point to %
+- Fixes [issue 979]: Instant bounds checking bypassed
```
malcolmr: "... when subtracting a `Duration`"?

(Also `Instant`, maybe? We use monospace style below.)

src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -4,6 +4,21 @@ User-visible changes from 1.0.0-beta1 onwards. See the
...
+- Fixes [issue 981]: Failure to parse time zones such as "Etc/GMT-12"
+
+## 1.4.1, released 2017-10-14 with tzdb2017b
+
+Bug-fix release:
+
+- Fixes [issue 981]: Failure to parse time zones such as "Etc/GMT-12"
```
malcolmr: "Failure to parse some time zones names that were prefixed by another valid time zone"?

src/NodaTime.Web/Markdown/root/versions.md
jskeet: Also it should be 957 :)

Will make it more relevant to end users, but still vague: "Time zone calculations could give incorrect results near the start/end of time"

src/NodaTime.Web/Markdown/root/versions.md
jskeet: SGTM

src/NodaTime.Web/Markdown/root/versions.md
jskeet: Well it's only if there's also another time zone in between - we'd be fine if the time zone list went "ABC", "ABCXX". It's only when it's "ABC", "ABCD", "ABCXX" that we have a problem. In reality, I think it's only going to be `Etc/GMT-1*`

PR #988, opened 2017-10-14
Fix text tests on some versions of Linux

### LINKED ISSUE #943: Investigate test failures on Gabriel
Gabriel should be very much like our Travis build - both Ubuntu 16.04.

Gabriel fails with two French tests though, formatting the day name as "Lundi" instead of the expected "lundi". It's `LocalDateTimePatternTest.Format` that fails. Will investigate. I suspect it's just a different version of localization data, and we should make the test less reliant on it.

jskeet:
In particular, my local Ubuntu machine uses capitalized day/month
names, causing tests to fail. This normalizes that just for the sake
of tests.

Fixes #943.

### DIFF
--- a/src/NodaTime.Test/Text/Cultures.cs
+++ b/src/NodaTime.Test/Text/Cultures.cs
@@ -110,8 +110,8 @@ namespace NodaTime.Test.Text
                 ShortTimePattern = "h:mm tt"
             }
         });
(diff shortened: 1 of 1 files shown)


jskeet:
Merging this now, partly to get a scheduled web site build to help diagnose #989.

PR #990, opened 2017-10-15
Fetch docfx version explicitly

### LINKED ISSUE #938: Fetch docfx during build
We shouldn't be using "whatever docfx version is on the machine" as we could easily get out of date between machines. Let's fetch it when we run the docs build.

jskeet:
The change to the annotation generator is due to a new version of
docfx which expects a comment at the start of each yml file (but the
comment isn't preserved by a load/save cycle).

I've got the history branch ready to push just before I merge this PR.

Fixes #938.


jskeet:
I'm going to self-merge this and some more PRs to unblock some tedious but useful work around improving the docs.

PR #1003, opened 2017-10-24
Cover InvalidOperationException in ZoneLocalMapping

heldersepu:


jskeet:
Hmm... I thought I'd covered these already. Will have a look tomorrow.

jskeet:
Ah, no - I see the problem, the count being invalid - your tests will fail because the constructor will (correctly) throw.

heldersepu:
ahhh I missed that ...
 Preconditions.DebugCheckArgumentRange(nameof(count), count, 0, 2);

PR #1005, opened 2017-10-30
Update TZDB downloads to https

### LINKED ISSUE #999: Check anywhere we're serving HTTP URLs
Example: https://nodatime.org/tzdb/index.txt and https://nodatime.org/tzdb/latest.txt both give HTTP URLs. Those URLs autodirect to the HTTPS ones.

jskeet:
I have a slight concern that this may trigger clients to get
confused, thinking that something's changed but then being surprised
when nothing has.

One option would be to use http for anything at or prior to 2017c,
but https from 2017d onwards.

Fixes #999. (I believe this is all there is left.)


jskeet:
Thoughts on the not-breaking-people part welcome, btw.

jskeet:
Let's go with it and see if it causes any problems - it's very hard to know in advance, really.

PR #1008, opened 2017-10-31
Add snippets for "Period" type showing basic construction using factory methods

carlosschults:


jskeet:
Thanks Carlos - will review later on (if I'm not too sleepy - I'm travelling at the moment).

Issue #299, opened 2015-03-15
Document exception behaviour for LocalDate operators (probably etc)


GoogleCodeExporter:
```
(Filing now to remind me to do this later, and also look in more detail.)

LocalDate's operator-(LocalDate) says "This is simply a convenience operator 
for calling Between(LocalDate, LocalDate). The calendar systems of the two 
dates must be the same."

However, the method documentation doesn't say what happens if that's not true. 
Also, Period.Between() doesn't document that precondition at all, though I 
would be surprised if it wasn't present.


Similarly, LocalDate's operator+(Period) and operator-(Period) say "Must not 
contain any (non-zero) time units", but don't document any exceptions.


Some of these probably apply to LocalDateTime and LocalTime, etc, as well.
```

Original issue reported on code.google.com by `malcolm.rowe` on 3 Jul 2014 at 9:38

PR #1011, opened 2017-11-04
Document exception behavior for Period

carlosschults:
Ok, trying to address #299.

I've just edited the xml documentation for the `Between` method in `Period` and for the "-" operator in `LocalDate`, and also added a small unit test asserting that the exception is being thrown.

### DIFF
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -661,7 +661,7 @@ namespace NodaTime
         /// <param name="end">End date</param>
         /// <returns>The period between the two dates, using year, month and day units.</returns>
         /// <exception cref="ArgumentException">
-        /// When <paramref name="start"/> and <paramref name="end"/> don't have the same calendar system.
+        /// <paramref name="start"/> and <paramref name="end"/> are not in the same calendar system.
         /// </exception>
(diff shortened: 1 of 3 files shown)


src/NodaTime/Period.cs
```diff
@@ -655,10 +655,14 @@ public static Period Between(LocalDate start, LocalDate end, PeriodUnits units)
         /// <remarks>
         /// If <paramref name="end"/> is before <paramref name="start" />, each property in the returned period
         /// will be negative.
+        /// The calendar systems of the two dates must be the same; an exception will be thrown otherwise.
         /// </remarks>
         /// <param name="start">Start date</param>
         /// <param name="end">End date</param>
         /// <returns>The period between the two dates, using year, month and day units.</returns>
+        /// <exception cref="ArgumentException">
+        /// When <paramref name="start"/> and <paramref name="end"/> don't have the same calendar system.
```
jskeet: Nit: remove the "When" for consistency

(For both, consider changing "don't have the same calendar system" to "are in different calendars", but I'm not massively fussed either way.)

jskeet:
Thanks very much :)

PR #1015, opened 2017-11-07
Document attempt of constructing a LocalDate passing a null calendar.

carlosschults:


src/NodaTime/LocalDate.cs
```diff
@@ -105,6 +105,7 @@ public LocalDate(int year, int month, int day)
         /// <param name="calendar">Calendar system in which to create the date.</param>
         /// <returns>The resulting date.</returns>
         /// <exception cref="ArgumentOutOfRangeException">The parameters do not form a valid date.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="calendar"/> is null.</exception>
```
jskeet: No need to do this - the fact that we have the NonNull annotation means these are already documented after generation. See https://nodatime.org/unstable/api/NodaTime.LocalDate.html#NodaTime_LocalDate__ctor_System_Int32_System_Int32_System_Int32_NodaTime_CalendarSystem_

src/NodaTime.Test/LocalDateTest.cs
```diff
@@ -57,6 +58,16 @@ public void XmlSerialization_Invalid(string xml, Type expectedExceptionType)
         }
 
         [Test]
+        public void Construction_NullCalendar_Throws()
+        {
+            Assert.Multiple(() =>
```
jskeet: Interesting - I haven't seen Assert.Multiple before.

carlosschults: Nice, isn't it? :) It's a new-ish feature in NUnit; I guess they borrowed that from xunit.

Issue #585, opened 2016-10-05
Add Deconstruct methods to support C# 7 deconstruction


jskeet:
Obvious examples:
- `ParseResult<T>`  - two overloads, one `Deconstruct(out bool success, out T value, out Exception error)`
- `LocalDate` - `Deconstruct(out int year, out int month, out int day)`
- `LocalTime` - `Deconstruct(out int hour, out int minute, out int second)`

We'll want to be careful with this, but it could be really nice...

([C# 7 feature description](https://github.com/dotnet/roslyn/blob/master/docs/features/deconstruction.md))

jskeet:
Punting for the moment in an effort to get 2.0 out.

carlosschults:
Can I take a shot at this?

jskeet:
Go for it :)

carlosschults:
Let's discuss here and try to decide for exactly which types are we adding deconstruct methods. Once we've  decided this, we could close these issue and open individual issues for each type.

Does that make sense?

jskeet:
I suspect it'll make sense to just keep a single issue anyway, to be honest - we could quite possibly get it all done in a single PR, possibly with one commit per type. I don't want to introduce too much issue churn :)

As a starting point, I'd suggest:

- LocalDate:
  - year/month/day
  - year/month/day/calendar
- LocalTime
  - hour/minute/second (subsecond becomes trickier with units)
- LocalDateTime
  - year/month/day/hour/minute/second
  - *Maybe* the equivalent with a calendar?
  - *Maybe* LocalDate/LocalTime?
- ZonedDateTime
  - year/month/day/hour/minute/second/DateTimeZone
  - *Maybe* LocalDate, LocalTime, DateTimeZone (and offset?)
- OffsetDateTime
  - year/month/day/hour/minute/second/Offset
  - *Maybe* LocalDate/LocalTime/Offset
- OffsetDate
  - year/month/day/Offset
  - Maybe the same with a calendar
- OffsetTime
  - hour/minute/second/Offset
- Interval
  - Start/End (nullable instants)
- DateInterval
  - Start/End (local dates)
- Instant, Duration, Period, DateTimeZone
  - Nothing?

carlosschults:
> Maybe LocalDate/LocalTime/DateTimeZone (and offset?)

I was thinking about LocalDateTime/DateTimeZone/Offset. Makes sense?

jskeet:
[...]

PR #1016, opened 2017-11-08
Implement deconstruction for LocalDate

carlosschults:
As suggested in #585.

### DIFF
--- a/src/NodaTime.Test/LocalDateTest.cs
+++ b/src/NodaTime.Test/LocalDateTest.cs
@@ -79,13 +79,32 @@ namespace NodaTime.Test
             var expectedYear = 2017;
             var expectedMonth = 11;
             var expectedDay = 7;
-            value.Deconstruct(out int actualYear, out int actualMonth, out int actualDay);
+            var (actualYear, actualMonth, actualDay) = value;
 
             Assert.Multiple(() =>
             {
-                Assert.AreEqual(expectedYear, actualYear, "Year comparison failed.");
-                Assert.AreEqual(expectedMonth, actualMonth, "Month comparison failed.");
-                Assert.AreEqual(expectedDay, actualDay, "Day comparison failed.");
+                Assert.AreEqual(expectedYear, actualYear);
+                Assert.AreEqual(expectedMonth, actualMonth);
+                Assert.AreEqual(expectedDay, actualDay);
+            });
+        }
+
+        [Test]
+        public void Deconstruction_IncludingCalendar()
+        {
+            var value = new LocalDate(2017, 11, 7, CalendarSystem.Gregorian);
+            var expectedYear = 2017;
+            var expectedMonth = 11;
+            var expectedDay = 7;
+            var expectedCalendar = CalendarSystem.Gregorian;
+            var (actualYear, actualMonth, actualDay, actualCalendar) = value;
+
+            Assert.Multiple(() =>
+            {
+                Assert.AreEqual(expectedYear, actualYear);
+                Assert.AreEqual(expectedMonth, actualMonth);
+                Assert.AreEqual(expectedDay, actualDay);
+                Assert.AreEqual(expectedCalendar, actualCalendar);
             });
         }
     }
--- a/src/NodaTime.Test/LocalDateTest.cs
+++ b/src/NodaTime.Test/LocalDateTest.cs
@@ -71,5 +71,22 @@ namespace NodaTime.Test
             Assert.AreEqual(CalendarSystem.Iso, value.Calendar);
             Assert.Throws<OverflowException>(() => value.PlusDays(-1));
         }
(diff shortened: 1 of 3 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -746,6 +746,20 @@ public LocalDate Previous(IsoDayOfWeek targetDayOfWeek)
         public LocalDate With([NotNull] Func<LocalDate, LocalDate> adjuster) =>
             Preconditions.CheckNotNull(adjuster, nameof(adjuster)).Invoke(this);
 
+        /// <summary>
```
jskeet: Possibly add another overload accepting a CalendarSystem (as a final parameter)?

src/NodaTime.Test/LocalDateTest.cs
```diff
@@ -71,5 +71,22 @@ public void MinIsoValue()
...
+        {
+            var value = new LocalDate(2017, 11, 7);
+            var expectedYear = 2017;
+            var expectedMonth = 11;
+            var expectedDay = 7;
+            value.Deconstruct(out int actualYear, out int actualMonth, out int actualDay);
+
+            Assert.Multiple(() =>
+            {
+                Assert.AreEqual(expectedYear, actualYear, "Year comparison failed.");
```
jskeet: No need for the messages here - if there are failures, we can tell what's failed by the stack trace, or probably just by the numbers.

src/NodaTime.Test/LocalDateTest.cs
```diff
@@ -71,5 +71,22 @@ public void MinIsoValue()
...
         }
+
+        [Test]
+        public void Deconstruction()
+        {
+            var value = new LocalDate(2017, 11, 7);
+            var expectedYear = 2017;
+            var expectedMonth = 11;
+            var expectedDay = 7;
+            value.Deconstruct(out int actualYear, out int actualMonth, out int actualDay);
```
jskeet: I'd use the deconstruction feature here to make it more obvious why we're doing this:

```csharp
var (actualYear, actualMonth, actualDay) = value;
```

carlosschults: Yeah, I try doing that, but then I'd have to add the System.ValueTuple package, and I wasn't sure I could.

And in that case, to which project should I add it? Just to the test project?

src/NodaTime.Test/LocalDateTest.cs
carlosschults: Got it.

src/NodaTime/LocalDate.cs
carlosschults: Sounds good.

src/NodaTime.Test/LocalDateTest.cs
jskeet: Yup - it's fine to have it in tests, but I don't want it as a dependency of Noda Time itself. (I believe fairly soon we won't need it at all for deconstruction to variables - I can find the Roslyn issue if you want.)

carlosschults:
@jskeet Thanks for the review.

Regarding which types should we add desconstruction to, maybe we should discuss in the issue?

jskeet:
Yup, absolutely - in fact, when I said "PR" I really meant "issue" :)

src/NodaTime/LocalDate.cs
```diff
@@ -760,6 +760,22 @@ public void Deconstruct(out int year, out int month, out int day)
...
+        /// Deconstructs the current instance into its components.
+        /// </summary>
+        /// <param name="year">The year component of the date.</param>
+        /// <param name="month">The month component of the date.</param>
+        /// <param name="day">The day component of the date.</param>
+        /// <param name="calendar">The <see cref="CalendarSystem"/> associated with the date.</param>
+        [Pure]
+        public void Deconstruct(out int year, out int month, out int day, [NotNull]out CalendarSystem calendar)
```
jskeet: I think it's probably worth removing the `[NotNull]` attribute for the moment - aside from anything else, at the moment the documentation generator would add an `ArgumentNullException` XML doc. Will need to modify that - we haven't had much in the way of out parameters before...

(It's fine to use "amend" and force-push for that - no need for another commit.)

carlosschults:
Well, it seems I've broken a test by removing the `[NotNull]` attribute :/

jskeet:
Will have a look when I get the chance, but I'm travelling today/tomorrow. I may just fix the doc generator :)

jskeet:
Okay, tell you what - let's get this in now *with* the NotNull attribute (which means tests should pass) and then I'll be able to fix up the tools in a way which is easy to verify. So if you could just put back the NotNull, that'd be grand - and please accept my sincere apologies for all the messing around :)

Issue #67, opened 2015-03-15
Consider whether to add PlusHours(), PlusMinutes(), etc to Period


GoogleCodeExporter:
```
Personally I'd rather have methods on Period that can be used to build a period 
by adding the various components one by one (returning a new Period each time).

    Period p = Period.Between(a,b).PlusDays(1).PlusHours(2);
    Period q = p.PlusDays(-1);
    ...

And therefore I'd also like these methods, and the operators to perform an 
automatic Normalize() each time they are used, so that PlusDays(4).PlusDays(3) 
would give you a Period of one week and PlusDays(1).PlusHours(-1) would give 
you a Period of 23 hours.



```

Original issue reported on code.google.com by `HighTech...@gmail.com` on 19 Apr 2012 at 3:21

jskeet:
Closing this for now. I don't think it's sufficiently useful to justify all the extra methods, and in the 7 years since the original report, we haven't had much interest.

PR #1017, opened 2017-11-08
Add a "PlusDays" method to Period type

carlosschults:
As suggested in #67.


jskeet:
I don't think we should add just `PlusDays` - I think we need to either add `PlusX` for *every* unit, or not do it at all. The extra "baggage" of all those methods is why I hadn't done it before.

@malcolmr @mj1856 thoughts?

carlosschults:
I guess it makes sense to close this PR for now and then discuss in #67 how we should approach this. Or decide we're not doing it and closing the issue, I think it's a win either way :)

PR #1020, opened 2017-11-08
Don't add exception docs for NotNull out parameters

### LINKED ISSUE #1019: NotNull on an out parameter should be documented differently
DocfxAnnotationGenerator.Program.AnnotateNotNullParameters needs fixing.

jskeet:
At some point we *may* want to automatically append something in the
parameter doc (to say that it won't be null on return) but that's
pretty low priority.

Fixes #1019.


jskeet:
(Have checked it locally - will merge on green.)

jskeet:
(Have checked it locally - will merge on green.)

Issue #918, opened 2017-08-02
Demo snippets for everything!


jskeet:
In the `NodaTime.Demo` project, we have snippets which are automatically turned into samples in the documentation. For example, [this code](https://github.com/nodatime/nodatime/blob/master/src/NodaTime.Demo/LocalDateDemo.cs#L13) is turned into the snippet [here in the docs](http://nodatime.org/unstable/api/NodaTime.LocalDate.html#NodaTime_LocalDate__ctor_System_Int32_System_Int32_System_Int32_)

The conversion process is NodaTime-specific at the moment and very much a work in progress (and it only works on the "Development" API docs at the moment) but that's something I can handle.

What I'd like help with is populating *lots* of snippets, for as much of the API surface as we can handle.

Please add a comment (or just file a PR!) if you're interested in helping out, particularly indicating an area of the API surface you'd like to cover - I'm expecting a bunch of feature requests for the snippet handling as go along, so don't be afraid to ask :)

carlosschults:
I'd love to help with that! What about the Period/Duration types, would this be a good place to start?

jskeet:
Sounds great to me, yes! For Period in particular, you can probably find interesting corner cases that would be well-demonstrated with snippets.

Are you happy to just dive in and try to use the snippet infrastructure as it is with the `LocalDate` code (first link in the first comment) as an example, or would you like me to set up a coupe of Period/Date snippets to start with?

carlosschults:
I believe I'd rather jump right in and start using the examples as you've said, and if I hit any obstacles I can contact you guys.

What do you think?

jskeet:
Sounds awesome to me - welcome aboard :)

carlosschults:
Hey @jskeet - or anyone else willing to help ;) - just a quick question.

Are there any special requirements in order to build Noda?

I'm using Visual Studio Community 2017 on Windows 10, and I have .net core sdk version 2.0 installed. 

I'm trying to build the NodaTime-All solution, but can't get it to work. Any tips?

Thanks!

jskeet:
VS 2017 and .NET Core SDK 2.0 should be all that's required, and NodaTime-All.sln is indeed the solution to add. Rather than discuss the problem on this issue, do you want to file a new issue with any error messages etc you get? That could easily help people who don't care about snippets.

Fortunately the Noda Time build environment is reasonably straightforward (for the main solution, anyway) so I'm sure we'll get you going soon :)

[...]

PR #1022, opened 2017-11-10
ZonedDateTime.TickOfDay snippet

mike-schenk:
For #918: 
Added documentation and a snippet to describe unexpected result from ZonedDateTime.TickOfDay.


mike-schenk:
Replacing with a different pull request that is more worthwhile.

PR #1023, opened 2017-11-11
Several snippets for the ZonedDateTime type.

mike-schenk:
For #918: Snippets for the ZonedDateTime type.
Also added documentation for NanosecondOfDay and TickOfDay to describe unexpected results.

### DIFF
--- a/src/NodaTime.Demo/ZonedDateTimeDemo.cs
+++ b/src/NodaTime.Demo/ZonedDateTimeDemo.cs
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+using System;
 using NUnit.Framework;
 
 namespace NodaTime.Demo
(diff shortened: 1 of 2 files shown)


src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+            Assert.AreNotEqual(duration.TotalTicks, nearEndOfDay.TickOfDay);
+        }
+
+        [Test]
+        public void AddDurationAndIsDaylightSavingTime()
+        {
+            // This is the date when DST ended at 2am in the given time zone.
+            var dt = new LocalDateTime(2017, 10, 29, 1, 45, 0);
+
+            DateTimeZone dublin = DateTimeZoneProviders.Tzdb["Europe/Dublin"];
```
jskeet: It's probably worth mentioning what the offsets are before and after the transition here.

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+
+            ZonedDateTime beforeTransition = new ZonedDateTime(dt, dublin, Offset.FromHours(1));
+
+            Assert.AreEqual(true, Snippet.For(beforeTransition.IsDaylightSavingTime()));
+            var result = Snippet.For(ZonedDateTime.Add(beforeTransition, Duration.FromHours(1)));
+            Assert.AreEqual(new LocalDate(2017, 10, 29), result.Date);
```
jskeet: A comment would be useful here: "Adding an hour of elapsed time takes us across the DST transition, so we have the same local time (shown on a clock) but a different offset" or similar.

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+            Assert.AreEqual(result, result2);
+
+            ZonedDateTime afterTransition = new ZonedDateTime(dt, dublin, Offset.FromHours(0));
+            result = ZonedDateTime.Add(afterTransition, Duration.FromDays(1));
+            Assert.AreEqual(new LocalTime(1, 45, 0), result.TimeOfDay);
+            Assert.AreEqual(new LocalDate(2017, 10, 30), result.Date);
+        }
+
+        [Test]
+        public void PlusDuration()
```
jskeet: Any reason for combining `Add` and `+`, but not `Plus`? It feels like it's similar enough that it's worth either splitting all three up, or combining all three. (I'd probably combine all three - it's quite a good way of saying "Here are three ways of doing exactly the same thing")

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+            Assert.AreEqual(new LocalDate(2017, 10, 30), result.Date);
+        }
+
+        [Test]
+        public void PlusHoursMinutesEtc()
+        {
+            DateTimeZone dublin = DateTimeZoneProviders.Tzdb["Europe/Dublin"];
+            var start = Instant.FromUtc(2017, 7, 17, 7, 17);
+            ZonedDateTime subject = new ZonedDateTime(start, dublin);
+            Assert.AreEqual(new ZonedDateTime(start + Duration.FromNanoseconds(1), dublin),
```
jskeet: I think it might be better to split these snippets up, one per unit. (I know that's annoying, but I think it'll give better results.)

What's slightly worrying is that the string representation may not go down to nanoseconds by default, so possibly the best format would be:

- Create the subject
- Create the result by adding the unit
- Assert a specific string result by formatting with the `F` standard pattern (which does down to the nanosecond, contrary to the documentation which I'm about to fix...)

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+        public void SubtractDuration()
+        {
+            // This is the date when DST ended at 2am in the given time zone.
+            var dt = new LocalDateTime(2017, 10, 29, 1, 45, 0);
+
+            DateTimeZone dublin = DateTimeZoneProviders.Tzdb["Europe/Dublin"];
+
+            ZonedDateTime afterDstTransition = new ZonedDateTime(dt, dublin, Offset.FromHours(0));
+
+            var result = Snippet.For(ZonedDateTime.Subtract(afterDstTransition, Duration.FromHours(1)));
```
jskeet: Again, show all of `Minus`, `Subtract` and `-` in one snippet?

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+            Assert.AreEqual(result, result2);
+
+            ZonedDateTime beforeTransition = new ZonedDateTime(dt, dublin, Offset.FromHours(1));
+            result = ZonedDateTime.Subtract(beforeTransition, Duration.FromDays(1));
+            Assert.AreEqual(new LocalTime(1, 45, 0), result.TimeOfDay);
+            Assert.AreEqual(new LocalDate(2017, 10, 28), result.Date);
+        }
+
+        [Test]
+        public void SubtractZonedDateTime()
```
jskeet: And again :)

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
...
+            ZonedDateTime other = new ZonedDateTime(Instant.FromUtc(2017, 7, 17, 9, 17), zone);
+
+            var difference = Snippet.For(other.Minus(subject));
+            Assert.AreEqual(Duration.FromHours(2), difference);
+        }
+
+        [Test]
+        public void NanosecondOfDay()
+        {
+            // This is a 25-hour day at the end of daylight saving time
```
jskeet: Same comments as per TickOfDay

mike-schenk:
I'll make the changes you've suggested.

jskeet:
Hi Mike - any update on this? If you don't have time, I could merge the PR in its current state then make the changes myself if you'd prefer.

src/NodaTime.Demo/ZonedDateTimeDemo.cs
zixuan75: what you should to do

src/NodaTime.Demo/ZonedDateTimeDemo.cs
```diff
@@ -39,5 +40,183 @@ public void SkippedLocalDateTime()
             DateTimeZone dublin = DateTimeZoneProviders.Tzdb["Europe/Dublin"];
```
zixuan75: just this for you

zixuan75:
@mike-schenk Hello mike, what are you doing? is this a bug?

src/NodaTime.Demo/ZonedDateTimeDemo.cs
zixuan75: just a comment?

jskeet:
As described, will merge and make the requested changes, to get this in.

PR #1031, opened 2017-11-15
Add more snippets to Period Demo class

carlosschults:
Added new snippets showing creation of periods via the `Between` method.

References #918.

### DIFF
--- a/src/NodaTime.Demo/PeriodDemo.cs
+++ b/src/NodaTime.Demo/PeriodDemo.cs
@@ -95,11 +95,8 @@ namespace NodaTime.Demo
                 new LocalDate(2017, 11, 10),
                 new LocalDate(2017, 11, 15)));
 
-            Assert.Multiple(() =>
-            {
-                Assert.AreEqual(5, period.Days);
-                Assert.AreEqual("P5D", period.ToString());
-            });
+            Assert.AreEqual(5, period.Days);
+            Assert.AreEqual("P5D", period.ToString());
         }
 
         [Test]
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -86,6 +86,146 @@ public void ConstructionFromNanoseconds()
...
+        }
+
+        [Test]
+        public void Between_LocalDatesDaysApart()
+        {
+            Period period = Snippet.For(Period.Between(
+                new LocalDate(2017, 11, 10),
+                new LocalDate(2017, 11, 15)));
+
+            Assert.Multiple(() =>
```
jskeet: I'd steer clear of Assert.Multiple in snippets. The snippet renderer isn't smart enough to understand that yet!
Just two separate assertions should be fine.

src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -86,6 +86,146 @@ public void ConstructionFromNanoseconds()
...
+                Assert.AreEqual("P27Y4M20D", period.ToString());
+            });
+        }
+
+        [TestCase(PeriodUnits.Months, 5, 0, 0, "P5M")]
+        [TestCase(PeriodUnits.Weeks, 0, 22, 0, "P22W")]
+        [TestCase(PeriodUnits.Months | PeriodUnits.Weeks, 5, 1, 0, "P5M1W")]
+        [TestCase(PeriodUnits.Months | PeriodUnits.Days, 5, 0, 7, "P5M7D")]
+        [TestCase(PeriodUnits.Days, 0, 0, 160, "P160D")]
+        public void Between_LocalDatesWithGivenUnits(PeriodUnits units, int expectedMonths, int expectedWeeks, int expectedDays, string expectedPattern)
```
jskeet: This would be great as a regular unit test, but won't work for snippets. It would be better to just choose the most interesting example - maybe a "years and days" one, as something a bit different?

src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -86,6 +86,146 @@ public void ConstructionFromNanoseconds()
...
+
+            Assert.Multiple(() =>
+            {
+                Assert.AreEqual(3, period.Hours);
+                Assert.AreEqual(5, period.Minutes);
+                Assert.AreEqual("PT3H5M", period.ToString());
+            });
+        }
+
+        [TestCase(PeriodUnits.Hours, 3, 0, 0, "PT3H")]
```
jskeet: Ditto - a single example, or multiple ones just expanded.

src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -86,6 +86,146 @@ public void ConstructionFromNanoseconds()
...
+                Assert.AreEqual(8, period.Months);
+                Assert.AreEqual(21, period.Days);
+                Assert.AreEqual(23, period.Hours);
+                Assert.AreEqual(32, period.Minutes);
+                Assert.AreEqual(2, period.Seconds);
+                Assert.AreEqual("P2Y8M21DT23H32M2S", period.ToString());
+            });
+        }
+
+        [TestCase(PeriodUnits.Years, 2, 0, 0, 0, 0, 0, "P2Y")]
```
jskeet: And again - not sure what interesting unit combination would be here.

src/NodaTime.Demo/PeriodDemo.cs
carlosschults: Got it.

Issue #864, opened 2017-06-14
Consider OffsetTime and OffsetDate types


mj1856:
Not saying per se that I want this, but there may be a need for an `OffsetTime` type - that is, a type that is `LocalTime` + `Offset`.

https://stackoverflow.com/questions/44514504/offsettime-in-nodatime

ISO8601 has this in section *4.2.5.2 Local time and the difference from UTC*, supporting the same basic/extended formats that we are used to with `OffsetDateTime`, sans the date portion.

Java 8 has this with [`java.time.OffsetTime`](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetTime.html).

PostgreSQL has this with [`time with time zone`](https://www.postgresql.org/docs/current/static/datatype-datetime.html), and mentions that it's in the SQL standard (ISO 9075), but discourages its use.  From their docs:

> The type `time with time zone` is defined by the SQL standard, but the definition exhibits properties which lead to questionable usefulness. In most cases, a combination of date, time, timestamp without time zone, and timestamp with time zone should provide a complete range of date/time functionality required by any application.

Do we want to add this to Noda Time?

mj1856:
We should also consider `OffsetDate`.  Though it's an odd duck, we have seen at least [one StackOverflow question on this](https://stackoverflow.com/q/44409687/634824).

In XML Schema Definition, both `xs:date` and `xs:time` have an offset field.  Though the offset is optional, it is reasonable that one might want to round-trip such a value through an object of some sort.

Alternatively, we could make an offset an option on the `LocalDate` and `LocalTime` objects, but I think that goes against our stated design philosophy.

mj1856:
Note that for `xs:date`, the offset is intended to reflect that on the first instant of the date.  If the offset changes over the date, that is not reflected.

Similarly, `xs:time` doesn't take into account that the offset could be different if applied to a different date.  Its use is limited, but is something to the effect of a recurring appointment in a fixed-offset time zone.

jskeet:
Okay - will start looking at this. We should probably think about what we want to do in terms of serialization (XML, binary, JSON) - that can come once we've got the basic functionality ready.

mj1856:
[...]

PR #1034, opened 2017-11-25
Implement serialization for OffsetDate and OffsetTime

jskeet:
I think this completes the code changes for #864. We then need to do
documentation, and when 2.3 has been released we can look at
Json.NET support.

### DIFF
--- a/src/NodaTime.Test/OffsetDateTest.cs
+++ b/src/NodaTime.Test/OffsetDateTest.cs
@@ -2,7 +2,11 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+using NodaTime.Calendars;
+using NodaTime.Text;
 using NUnit.Framework;
+using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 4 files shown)


src/NodaTime.Test/OffsetDateTest.cs
```diff
@@ -148,5 +151,39 @@ public void Deconstruction()
...
+        public void BinarySerialization()
+        {
+            var date = new LocalDate(2013, 4, 12);
+            var offset = Offset.FromHoursAndMinutes(5, 30);
+            var value = new OffsetDate(date, offset);
+            TestHelper.AssertBinaryRoundtrip(value);
+        }
+
+        [Test]
+        public void XmlSerialization_Iso()
```
malcolmr: Could you add tests covering BCE?

Specifically, for OffsetDate, I assume that the serialized form of January 1st, 2 BCE would be something like `-0001-01-01Z`, following (I assume) ISO 8601:2000, and not e.g. the craziness discussed in https://www.w3.org/TR/xmlschema-2/#dateTime.

malcolmr: Also a test that clarifies whether 123 AD is `123-01-01Z` or (what I believe might be required for `xs:date`) `0123-01-01Z` would be nice.

jskeet: Done, thanks. (And yes, we're following ISO-8601.)

jskeet: It's 0123-01-01 - all our default formatters use `uuuu` as the year specifier.

PR #1035, opened 2017-11-27
Add Wondrous calendar

glittle:
@jskeet I'm not sure how to collapse these into a single commit. It took a number of attempts just to get it this far!  Feel free to do so if you can.

### DIFF
--- a/src/NodaTime/CalendarSystem.cs
+++ b/src/NodaTime/CalendarSystem.cs
@@ -53,6 +53,9 @@ namespace NodaTime
         private const string CopticName = "Coptic";
         private const string CopticId = CopticName;
 
+        private const string WondrousName = "Wondrous";
+        private const string WondrousId = WondrousName;
+
         private const string JulianName = "Julian";
         private const string JulianId = JulianName;
 
@@ -151,6 +154,8 @@ namespace NodaTime
                     return Julian;
                 case CalendarOrdinal.Coptic:
                     return Coptic;
+                case CalendarOrdinal.Wondrous:
+                    return Wondrous;
                 case CalendarOrdinal.HebrewCivil:
                     return HebrewCivil;
                 case CalendarOrdinal.HebrewScriptural:
@@ -200,6 +205,7 @@ namespace NodaTime
             {HebrewScripturalId, () => GetHebrewCalendar(HebrewMonthNumbering.Scriptural)},
             {GregorianId, () => Gregorian},
             {CopticId, () => Coptic},
+            {WondrousId, () => Wondrous},
             {JulianId, () => Julian},
             {UmAlQuraId, () => UmAlQura},
             {GetIslamicId(IslamicLeapYearPattern.Indian, IslamicEpoch.Civil), () => GetIslamicCalendar(IslamicLeapYearPattern.Indian, IslamicEpoch.Civil)},
@@ -247,7 +253,27 @@ namespace NodaTime
             Preconditions.CheckArgumentRange(nameof(monthNumbering), (int) monthNumbering, 1, 2);
             return HebrewCalendars.ByMonthNumbering[((int) monthNumbering) - 1];
         }
-        
+
+        /// <summary>
+        /// Returns the Wondrous (Badí') calendar, as described at https://en.wikipedia.org/wiki/Badi_calendar. 
+        /// This is a purely solar calendar with years starting at the vernal equinox.
+        /// </summary>
+        /// <remarks>
+        /// <para>The Wondrous calendar was developed and defined by the founders of the Bahá'í Faith in the mid to late
+        /// 1800's A.D. The first year in the calendar coincides with 1844 A.D. Years are labeled "B.E." for Bahá'í Era.</para>
+        /// <para>A year consists of 19 months, each with 19 days. Each day starts at sunset. Years are grouped into sets
+        /// of 19 "Unities" (Váḥid) and 19 Unities make up 1 "All Things" (Kull-i-Shay’).</para>
+        /// <para>A period of days (usually 4 or 5, called Ayyám-i-Há) occurs between the 18th and 19th months. The length of this 
+        /// period of intercalary days is solely determined by the date of the following vernal equinox. The vernal equinox is 
+        /// a momentary point in time, so the "date" of the equinox is determined by the date (beginning 
+        /// at sunset) in effect in Tehran, Iran at the moment of the equinox.</para>
+        /// <para>In this Noda Time implementation, days start at midnight and lookup tables are used to determine vernal equinox dates.
+        /// Ayyám-i-Há is internally modelled as extra days added to the 18th month. As a result, a few functions will
+        /// not work as expected for Ayyám-i-Há, such as EndOfMonth.</para>
+        /// </remarks>
+        /// <returns>The Wondrous calendar system.</returns>
+        [NotNull] public static CalendarSystem Wondrous => MiscellaneousCalendars.Wondrous;
+
         /// <summary>
         /// Returns an Islamic, or Hijri, calendar system.
         /// </summary>
@@ -356,6 +382,7 @@ namespace NodaTime
         ///   <item><term>ISO</term><description><see cref="CalendarSystem.Iso"/></description></item>
         ///   <item><term>Gregorian</term><description><see cref="CalendarSystem.Gregorian"/></description></item>
         ///   <item><term>Coptic</term><description><see cref="CalendarSystem.Coptic"/></description></item>
+        ///   <item><term>Wondrous</term><description><see cref="CalendarSystem.Wondrous"/></description></item>
         ///   <item><term>Julian</term><description><see cref="CalendarSystem.Julian"/></description></item>
         ///   <item><term>Hijri Civil-Indian</term><description><see cref="CalendarSystem.GetIslamicCalendar"/>(IslamicLeapYearPattern.Indian, IslamicEpoch.Civil)</description></item>
         ///   <item><term>Hijri Civil-Base15</term><description><see cref="CalendarSystem.GetIslamicCalendar"/>(IslamicLeapYearPattern.Base15, IslamicEpoch.Civil)</description></item>
@@ -794,6 +821,8 @@ namespace NodaTime
                 new CalendarSystem(CalendarOrdinal.Coptic, CopticId, CopticName, new CopticYearMonthDayCalculator(), Era.AnnoMartyrum);
             internal static readonly CalendarSystem UmAlQura =
                 new CalendarSystem(CalendarOrdinal.UmAlQura, UmAlQuraId, UmAlQuraName, new UmAlQuraYearMonthDayCalculator(), Era.AnnoHegirae);
+            internal static readonly CalendarSystem Wondrous =
+                new CalendarSystem(CalendarOrdinal.Wondrous, WondrousId, WondrousName, new WondrousYearMonthDayCalculator(), Era.BahaiEra);
 
             // Static constructor to enforce laziness. This used to be important to avoid a Heisenbug.
             // I don't believe it's strictly required now, but it does no harm and I don't want to go
--- a/src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
+++ b/src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
@@ -1,0 +1,439 @@
+﻿// Copyright 2014 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using NodaTime.Utility;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace NodaTime.Calendars
+{
+    /// <summary>
+    ///     See <see cref="CalendarSystem.Wondrous" /> for details.
+    /// </summary>
+    internal sealed class WondrousYearMonthDayCalculator : YearMonthDayCalculator
+    {
+        // named constants to avoid use of raw numbers in the code
+        private const int AverageDaysPer10Years = 3652; // Ideally 365.2425 per year...
+        private const int DaysInAyyamiHaInLeapYear = 5;
+        private const int DaysInAyyamiHaInNormalYear = 4;
+
+        private const int DaysInMonth = 19;
+        private const int FirstYearOfStandardizedCalendar = 172;
+        private const int GregorianYearOfFirstWondrousYear = 1844;
+
+        private const int Month18 = 18;
+        private const int Month19 = 19;
+        private const int MonthsInYear = 19;
+
+        private const int UnixEpochDayAtStartOfYear1 = -45941;
+        private const int WondrousMaxYear = 1000; // current lookup tables are pre-calculated for a thousand years
+        private const int WondrousMinYear = 1;
+
+        private static Dictionary<int, WondrousYearInfo> _calculatedYears;
+
+        /// <remarks>
+        ///     There are 19 months in a year. Between the 18th and 19th month are the "days of Ha" (Ayyam-i-Ha).
+        ///     Options for numbering the months:
+        ///     - treat Ayyam-i-Ha as month 0. This will cause problems if 0 is treated as unknown. When stored internally,
+        ///       NodaTime months are converted to 0-based numbers, so this would by -1 which is not supported.
+        ///     - treat Ayyam-i-Ha as month -1. This would be a problem if the base classes convert to 0 based months
+        ///     - treat Ayyam-i-Ha as month 19. This is confusing because the 19th month would be displayed as month #20.
+        ///     - treat Ayyam-i-Ha as month 20. This will cause problems if months are sorted naively.
+        ///     - treat Ayyam-i-Ha as extra days in month 18. This would require special handling to sometimes display
+        ///       days in "month 18" as being Ayyam-i-Ha, but we have no direct control over display.
+        ///       
+        /// This version will treat Ayyam-i-Ha days as extra days in month 18.
+        /// 
+        /// Users interacting with the system should use 0 as Ayyam-i-Ha.
+        /// </remarks>
+        public const int AyyamiHaMonth0 = 0;
+
+        static byte[] YearInfoRaw = Convert.FromBase64String("AgELAgIBCwICAQsCAgEBCwIBAQsCAQELAgEBCwIBAQsCAQELAgEBCwIBAQELAQEBCwEBAQsBAQELAQEB" +
+            "CwEBAQsBAQELAQEBCwEBAQEKAQEBCgEBAQsCAgILAgICCwICAgsCAgILAgICCwICAgELAgIBCwICAQsC" +
+            "AgELAgIBCwICAQsCAgELAgIBCwICAQELAgEBCwIBAQsCAQELAgEBCwIBAQsCAQELAgEBCwIBAQELAQEB" +
+            "CwEBAQsCAgIMAgICDAICAgwCAgIMAgICDAICAgILAgICCwICAgsCAgILAgICCwICAgsCAgILAgICCwIC" +
+            "AgELAgIBCwICAQsCAgELAgIBCwICAQsCAgELAgIBCwICAQELAgEBCwIBAQsCAgIMAwICDAMCAgwDAgIM" +
+            "AwICDAMCAgIMAgICDAICAgwCAgIMAgICDAICAgwCAgIMAgICDAICAgILAgICCwICAgsCAgILAgICCwIC" +
+            "AgsCAgILAgICAQsCAgELAgIBCwICAQsCAgELAgIBCwICAQsCAgELAgIBCwICAQELAgEBCwIBAQsCAQEL" +
+            "AgEBCwIBAQsCAQELAgEBCwIBAQELAQEBCwEBAQsBAQELAQEBCwEBAQsBAQELAQEBCwEBAQEKAQEBCgEB" +
+            "AQoBAQELAgICCwICAgsCAgILAgICAQsCAgELAgIBCwICAQsCAgELAgIBCwICAQsCAgELAgIBAQsCAQEL" +
+            "AgEBCwIBAQsCAQELAgEBCwIBAQsCAQELAgEBAQsBAQELAQEBCwEBAQsBAQELAgICDAICAgwCAgIMAgIC" +
+            "AgsCAgILAgICCwICAgsCAgILAgICCwICAgsCAgILAgICAQsCAgELAgIBCwICAQsCAgELAgIBCwICAQsC" +
+            "AgELAgIBAQsCAQELAgEBCwIBAQsCAQELAgICDAMCAgwDAgIMAwICAgwCAgIMAgICDAICAgwCAgIMAgIC" +
+            "DAICAgwCAgIMAgICAgsCAgILAgICCwICAgsCAgILAgICCwICAgsCAgILAgICAQsCAgELAgIBCwICAQsC" +
+            "AgELAgIBCwICAQsCAgELAgIBAQsCAQELAgEBCwIBAQsCAQELAgEBCwIBAQsCAQELAg==");
+
+        static WondrousYearMonthDayCalculator()
+        {
+            Preconditions.CheckState(FirstYearOfStandardizedCalendar + YearInfoRaw.Length == WondrousMaxYear + 1, "Invalid compressed data. Length: " + YearInfoRaw.Length);
+        }
+
+
+        internal WondrousYearMonthDayCalculator()
+            : base(WondrousMinYear,
+                WondrousMaxYear - 1,
+                AverageDaysPer10Years,
+                UnixEpochDayAtStartOfYear1)
+        { }
+
+        private int DayOfYearOfStartOfMonth19(int year)
+        {
+            var dayOfYearOfStartOfMonth19 = 1 + DaysInMonth * Month18 + GetYearInfo(year).DaysInAyyamiHa;
+            return dayOfYearOfStartOfMonth19;
+        }
+
+        static WondrousYearInfo GetYearInfo(int year)
+        {
+            // Considered preloading this knowledge from the byte array into an array of 1000 years, but in typical use, only one or two of the years may be used. 
+            // Calculated results are cached in a dictionary in case they are re-used.
+
+            Preconditions.CheckArgumentRange(nameof(year), year, WondrousMinYear, WondrousMaxYear);
+
+            if (_calculatedYears == null)
+            {
+                _calculatedYears = new Dictionary<int, WondrousYearInfo>();
+            }
+            else if (_calculatedYears.ContainsKey(year))
+            {
+                return _calculatedYears[year];
+            }
+
+            WondrousYearInfo wondrousYearInfo;
+
+            if (year < FirstYearOfStandardizedCalendar)
+            {
+                // Feb 29 is near the end of the Wondrous year in the next Gregorian year
+                wondrousYearInfo = new WondrousYearInfo(CalendarSystem.Iso.YearMonthDayCalculator.IsLeapYear(year + GregorianYearOfFirstWondrousYear)
+                                    ? DaysInAyyamiHaInLeapYear : DaysInAyyamiHaInNormalYear,
+                                    21);
+                _calculatedYears.Add(year, wondrousYearInfo);
+                return wondrousYearInfo;
+            }
+
+            /*
+            Base64 string only has information for years 172 to 1000.
+                NazRuzDate falls on March 19,20,21, or 22
+                DaysInAyymiHa can be 4,5
+
+                For each year, num = (NawRuzDate - 19) + 10 * (DaysInAyyamiHa - 4)
+                ... DaysInAyyamiHa are 4 or 5, so we add 10 only when there are 5 days
+
+                string = Convert.ToBase64String(new byte[] { num, num, ...})
+            */
+
+            int info = YearInfoRaw[year - FirstYearOfStandardizedCalendar];
+            int daysInAyyamiHa;
+            if (info > 10)
+            {
+                daysInAyyamiHa = DaysInAyyamiHaInLeapYear;
+                info -= 10;
+            }
+            else
+            {
+                daysInAyyamiHa = DaysInAyyamiHaInNormalYear;
+            }
+
+            const int dayInMarchForOffsetToNawRuz = 19;
+            wondrousYearInfo = new WondrousYearInfo(daysInAyyamiHa, dayInMarchForOffsetToNawRuz + info);
+
+            _calculatedYears.Add(year, wondrousYearInfo);
+
+            return wondrousYearInfo;
+        }
+
+        private DateTime GregorianDateOfNawRuz(int year)
+        {
+            Preconditions.CheckArgumentRange(nameof(year), year, WondrousMinYear, WondrousMaxYear);
+
+            var gregorianDateOfNawRuz = new DateTime(GregorianYearOfFirstWondrousYear - 1 + year, 3, GetYearInfo(year).NawRuzDayInMarch);
+
+            return gregorianDateOfNawRuz;
+        }
+
+        protected override int CalculateStartOfYearDays(int year)
+        {
+            Preconditions.CheckArgumentRange(nameof(year), year, WondrousMinYear, WondrousMaxYear);
+
+            // leverage Noda Time's Iso knowledge of when Jan 1 is, then add days to Naw Ruz
+            var gregorianYear = year + GregorianYearOfFirstWondrousYear - 1;
+            var gregorianStart = CalendarSystem.Iso.YearMonthDayCalculator.GetStartOfYearInDays(gregorianYear);
+
+            var nawRuz = GregorianDateOfNawRuz(year);
+            var firstDay = gregorianStart + nawRuz.DayOfYear - 1;
+
+            return firstDay;
+        }
+
+        protected override int GetDaysFromStartOfYearToStartOfMonth(int year, int month)
+        {
+            var daysFromStartOfYearToStartOfMonth = DaysInMonth * (month - 1);
+
+            if (month == Month19)
+            {
+                daysFromStartOfYearToStartOfMonth += GetYearInfo(year).DaysInAyyamiHa;
+            }
+
+            return daysFromStartOfYearToStartOfMonth;
+        }
+
+        internal override YearMonthDay AddMonths(YearMonthDay start, int months)
+        {
+            if (months == 0)
+            {
+                return start;
+            }
+
+            var movingBackwards = months < 0;
+
+            var thisMonth = start.Month;
+            var thisYear = start.Year;
+            var thisDay = start.Day;
+
+            var nextDay = thisDay;
+
+            if (IsInAyyamiHa(start))
+            {
+                nextDay = thisDay - DaysInMonth;
+
+                if (movingBackwards)
+                {
+                    thisMonth++;
+                }
+            }
+
+            var nextYear = thisYear;
+            var nextMonthNum = thisMonth + months;
+
+            if (nextMonthNum > MonthsInYear)
+            {
+                nextYear = thisYear + nextMonthNum / MonthsInYear;
+                nextMonthNum = nextMonthNum % MonthsInYear;
+            }
+            else if (nextMonthNum < 1)
+            {
+                nextMonthNum = MonthsInYear - nextMonthNum;
+                nextYear = thisYear - nextMonthNum / MonthsInYear;
+                nextMonthNum = MonthsInYear - nextMonthNum % MonthsInYear;
+            }
+
+            Preconditions.CheckArgumentRange(nameof(nextYear), nextYear, WondrousMinYear, WondrousMaxYear);
+
+            var result = new YearMonthDay(nextYear, nextMonthNum, nextDay);
+
+            return result;
+        }
+
+        internal override int GetDaysInMonth(int year, int month)
+        {
+            Preconditions.CheckArgumentRange(nameof(year), year, WondrousMinYear, WondrousMaxYear);
+
+            if (month == AyyamiHaMonth0)
+            {
+                return GetYearInfo(year).DaysInAyyamiHa;
+            }
+
+            return DaysInMonth;
+        }
+
+        internal override int GetDaysInYear(int year)
+        {
+            var daysInYear = 361 + GetYearInfo(year).DaysInAyyamiHa;
+
+            return daysInYear;
+        }
+
+        internal override int GetDaysSinceEpoch(YearMonthDay target)
+        {
+            var month = target.Month;
+            var year = target.Year;
+
+            var firstDay0OfYear = CalculateStartOfYearDays(year) - 1;
+
+            var daysSinceEpoch = firstDay0OfYear
+                 + (month - 1) * DaysInMonth
+                 + target.Day;
+
+            if (month == Month19)
+            {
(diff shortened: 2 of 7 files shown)


jskeet:
Hmm... it looks like it's still not fully merged/rebased. I'll have a look and see what I can do. I assume that almost all your changes are actually new files rather than modifications, which should make it simpler...

jskeet:
Okay, I think it's sorted now. I've rebased up to master, then rebased to a single commit and force-pushed onto your fork. I can now start the review - which I'll do as I find time, so it'll be bitty, I'm afraid.

src/NodaTime/CalendarSystem.cs
```diff
@@ -247,7 +253,28 @@ internal static CalendarSystem ForOrdinalUncached([Trusted] CalendarOrdinal ordi
...
+        /// <remarks>
+        /// <para>The Wondrous calendar was developed and defined by the founders of the Bahá'í Faith in the mid to late
+        /// 1800's A.D. The first year in the calendar coincides with 1844 A.D. Years are labeled "B.E." for Bahá'í Era.</para>
+        /// <para>A year consists of 19 months, each with 19 days. Each day starts at sunset. Years are grouped into sets
+        /// of 19 "Unities" (Váḥid) and 19 Unities make up 1 "All Things" (Kull-i-Shay’).</para>
+        /// <para>A period of days (usually 4 or 5, called Ayyám-i-Há) occurs between the 18th and 19th months. The length of this 
+        /// period of intercalary days is solely determined by the pre-calculated time of the following vernal equinox. 
+        /// The vernal equinox is a momentary point in time, so the "date" of the equinox is determined by the day (beginning 
+        /// at sunset) in effect in Tehran, Iran at the moment of the equinox. These times can be calculated, but published
+        /// tables are used in this implementation keep it simple!</para>
```
jskeet: It's probably worth moving the implementation detail comment into a comment in WondrousCalculator.

src/NodaTime/CalendarSystem.cs
```diff
@@ -247,7 +253,28 @@ internal static CalendarSystem ForOrdinalUncached([Trusted] CalendarOrdinal ordi
...
-        
+
+        /// <summary>
+        /// Returns the Wondrous (Badí') calendar, as described at https://en.wikipedia.org/wiki/Badi_calendar. 
+        /// This is a purely solar calendar with years starting at the vernal equinox.
+        /// </summary>
+        /// <remarks>
+        /// <para>The Wondrous calendar was developed and defined by the founders of the Bahá'í Faith in the mid to late
+        /// 1800's A.D. The first year in the calendar coincides with 1844 A.D. Years are labeled "B.E." for Bahá'í Era.</para>
+        /// <para>A year consists of 19 months, each with 19 days. Each day starts at sunset. Years are grouped into sets
```
jskeet: Mention that Noda Time doesn't model days starting at sunset? (There may be something appropriate for the Hebrew calendar that could be copied.)

src/NodaTime/CalendarSystem.cs
```diff
@@ -247,7 +253,28 @@ internal static CalendarSystem ForOrdinalUncached([Trusted] CalendarOrdinal ordi
...
+        /// of 19 "Unities" (Váḥid) and 19 Unities make up 1 "All Things" (Kull-i-Shay’).</para>
+        /// <para>A period of days (usually 4 or 5, called Ayyám-i-Há) occurs between the 18th and 19th months. The length of this 
+        /// period of intercalary days is solely determined by the pre-calculated time of the following vernal equinox. 
+        /// The vernal equinox is a momentary point in time, so the "date" of the equinox is determined by the day (beginning 
+        /// at sunset) in effect in Tehran, Iran at the moment of the equinox. These times can be calculated, but published
+        /// tables are used in this implementation keep it simple!</para>
+        /// </remarks>
+        /// <returns>The Wondrous calendar system.</returns>
+        [NotNull] public static CalendarSystem GetWondrousCalendar()
```
jskeet: Not sure why a method is needed here - just make it a property? (Starting from 2.0, most calendar systems are accessible via properties. We only have methods where there's parameterization.)

jskeet: (I see there's a property already, so let's just remove the method.)

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace NodaTime.Calendars.Wondrous
+{
+    /// <summary>
+    /// Helper and extension methods for the Wondrous Calendar.
+    /// </summary>
+    /// <remarks>
+    /// The Wondrous Calendar Helper is in a distinct namespace so that the Extensions are only available when this
```
jskeet: Extensions => extension methods

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+    /// namespace is included.
```
jskeet: included => imported (that's how using directives are typically described)

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+        ///       days in "month 18" as being Ayyam-i-Ha, but we have no direct control over display.
+        ///       
+        /// This version will treat Ayyam-i-Ha days as extra days in month 18.
+        /// 
+        /// Users interacting with the system should use 0 as Ayyam-i-Ha.
+        /// </remarks>
+        internal const int AyyamiHaMonth0 = 0;
+
+        /// <summary>
+        ///
```
jskeet: Please add documentation

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+        /// This version will treat Ayyam-i-Ha days as extra days in month 18.
+        /// 
+        /// Users interacting with the system should use 0 as Ayyam-i-Ha.
+        /// </remarks>
+        internal const int AyyamiHaMonth0 = 0;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="wondrousYear">Year in the Wondrous calendar</param>
```
jskeet: I'd rename these to just `year`, `month`, `date` - but *possibly* rename the method to `CreateWondrousDate` as then if the helper is imported statically, it can be called very easily.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+                Preconditions.CheckArgumentRange(nameof(wondrousDay), wondrousDay, 1, maxDay);
+
+                // move Ayyam-i-Ha days to fall after the last day of month 18.
+                wondrousMonth = Month18;
+                wondrousDay += DaysInMonth;
+            }
+            return new LocalDate(wondrousYear, wondrousMonth, wondrousDay, CalendarSystem.Wondrous);
+        }
+
+        /// <summary>
```
jskeet: Docs :)

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+        /// </summary>
+        /// <param name="wondrousYear"></param>
+        /// <param name="wondrousMonth"></param>
+        /// <param name="wondrousDay"></param>
+        /// <param name="hour"></param>
+        /// <param name="minute"></param>
+        /// <param name="second"></param>
+        /// <param name="millisecond"></param>
+        /// <returns></returns>
+        public static LocalDateTime CreateDateTime(int wondrousYear, int wondrousMonth, int wondrousDay, int hour, int minute = 0, int second = 0, int millisecond = 0)
```
jskeet: So far we've avoided using optional parameters everywhere in Noda Time (I think). Let's come back to this.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+        {
+            if (wondrousMonth == AyyamiHaMonth0)
```
jskeet: Is there any reason this method isn't implemented via `CreateDate(year, month, day) + new LocalTime(hour, minute, second,...)`?

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+            return new LocalDateTime(wondrousYear, wondrousMonth, wondrousDay, hour, minute, second, millisecond, CalendarSystem.Wondrous);
+        }
+
+        /// <summary>
+        /// Return the month of this date. If in Ayyam-i-Ha, returns 0.
+        /// </summary>
+        /// <remarks>Deals with Ayyam-i-Ha.</remarks>
+        /// <param name="input"></param>
+        /// <returns></returns>
+        public static int WondrousMonthNum(this LocalDate input)
```
jskeet: I'd suggest calling this `GetWondrousMonth` - the `Num` part is definitely non-idiomatic.
(Ditto following methods.)

Probably worth throwing around multiple ideas before you implement the change though.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+
+        //public static WondrousYearMonthDay AsWondrousYearMonthDay(this LocalDate input)
+        //{
+        //    return new WondrousYearMonthDay(
+        //        input.Year,
+        //        input.WondrousMonthNum(),
+        //        input.WondrousDayNum());
+        //}
+
+        /// <summary>
```
jskeet: I'd remove this until we have an implementation.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+            // to be enhanced...
+
+            var year = input.Year;
+            var month = input.WondrousMonthNum();
+            var day = input.WondrousDayNum();
+
+            return $"{year}-{month}-{day}";
+        }
+
+        internal static WondrousYearInfo GetYearInfo(int year)
```
jskeet: This is basically 2000 lines of data. We can do a lot better than this :)
I suggest we work out a system like UmAlQuraYearMonthDayCalculator, with a base64 string containing generated data.

Am I right in saying the first value is always 4 or 5, and the second value is always 20, 21 or 22? If so, we can easily store the data for 2 years in each byte of data.

We could potentially merge like this, then I can rewrite this part, if that would be simpler than coordinating. (Is there a machine-readable form of this data somewhere so that it's easy to write a generator?)

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
...
+                case 694:
+                    return new WondrousYearInfo(5, 20);
+                case 695:
+                    return new WondrousYearInfo(4, 21);
+                case 696:
+                    return new WondrousYearInfo(4, 21);
+                case 697:
+                    return new WondrousYearInfo(4, 20);
+                case 698:
+
```
jskeet: I'd expect all this information and the method itself to be in the WondrousCalculator. I wouldn't expect it to be needed outside that, except maybe for testing.

jskeet: Please remove this old code.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+{
+    /// <summary>
+    ///     See <see cref="CalendarSystem.GetWondrousCalendar" /> for details.
+    /// </summary>
+    internal sealed class WondrousYearMonthDayCalculator : YearMonthDayCalculator
+    {
+        private const int WondrousMinYear = 1;
+        private const int WondrousMaxYear = 1000;
+
+        private const int AverageDaysPer10Years = 3652; // Ideally 365.2425 per year...
```
jskeet: Is this correct, or was it copied from somewhere?

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+
+        private const int UnixEpochDayAtStartOfYear1 = -45941;
+
+        internal WondrousYearMonthDayCalculator()
+            : base(WondrousMinYear,
+                WondrousMaxYear - 1,
+                AverageDaysPer10Years,
+                UnixEpochDayAtStartOfYear1)
+        {
+            ConsoleWriteLine("~A Constructed");
```
jskeet: Please remove both lines.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+            //            System.Diagnostics.Debugger.Launch();
+        }
+
+
+        /// <summary>
```
jskeet: Please remove.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+
+namespace NodaTime.Calendars
+{
+    /// <summary>
+    ///     See <see cref="CalendarSystem.GetWondrousCalendar" /> for details.
+    /// </summary>
+    internal sealed class WondrousYearMonthDayCalculator : YearMonthDayCalculator
+    {
+        private const int WondrousMinYear = 1;
+        private const int WondrousMaxYear = 1000;
```
jskeet: Is this arbitrary, or inherently part of the calendar?

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+            Preconditions.CheckArgumentRange(nameof(year), year, WondrousMinYear, WondrousMaxYear);
+
+            var daysInMonth = DaysInMonth;
+
+            if (month == AyyamiHaMonth0)
+            {
+                daysInMonth = WondrousCalendarHelper.GetYearInfo(year).DaysInAyyamiHa;
+            }
+
+            ConsoleWriteLine("~E GetDaysInMonth {0} {1} ==> {2}", year, month, daysInMonth);
```
jskeet: Please remove all of these diagnostic calls.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+            Preconditions.CheckArgumentRange(nameof(nextYear), nextYear, WondrousMinYear, WondrousMaxYear);
+
+            var result = new YearMonthDay(nextYear, nextMonthNum, nextDay);
+
+            ConsoleWriteLine("~H AddMonths {0} {1} ==> {2}", start, months, result);
+
+            return result;
+        }
+
+        ///// <param name="start">M - x</param>
```
jskeet: Remove, as it's implemented below?

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+            }
+
+            ConsoleWriteLine("~B GetDayFromStartOfYearToStartOfMonth {0} {1} ==> {2}", year, month,
+                daysFromStartOfYearToStartOfMonth);
+            return daysFromStartOfYearToStartOfMonth;
+        }
+
+        /// <param name="wYear"></param>
+        /// <remarks>Rely on Gregorian calendar for "days" in unix epoch</remarks>
+        /// <remarks>Untested</remarks>
```
jskeet: Can you add tests for all the untested code?

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+        /// <returns></returns>
+        protected override int CalculateStartOfYearDays(int wYear)
+        {
+            Preconditions.CheckArgumentRange(nameof(wYear), wYear, WondrousMinYear, WondrousMaxYear);
```
jskeet: Please don't change the parameter name like this - it means the caller won't necessarily know which parameter it's talking about (in general) as the method *they* called may not have a `wYear` parameter.
(I'll see if I can add a Roslyn test to detect this...)

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,385 @@
...
+
+            var gregorianYear = wYear + 1843;
+            var gregorianStart = new GregorianYearMonthDayCalculator().GetStartOfYearInDays(gregorianYear);
```
jskeet: Please don't create a new calculator each time this is called - that could be really painful in terms of performance. (We generally assume this method is cheap, I think.) 

You can use `CalendarSystem.Iso.YearMonthDayCalculator`.

src/NodaTime/LocalDate.cs
```diff
@@ -35,6 +36,7 @@ namespace NodaTime
 #if !NETSTANDARD1_3
     [Serializable]
 #endif
+    [DebuggerDisplay("{Year}-{Month}-{Day} {Calendar.Name}")]
```
jskeet: Let's *maybe* do this in a separate PR, but please remove it from this one.

glittle:
Yes, this will keep me busy for a while... and half of it will be trying to juggle git!

Where should I start doing this work? I.e. which branch/fork, etc.

src/NodaTime.Test/Calendars/BclCalendars.cs
```diff
@@ -74,6 +74,7 @@ internal static CalendarSystem CalendarSystemForCalendar(Calendar bcl)
                 case "HebrewCalendar": return CalendarSystem.HebrewCivil;
                 case "PersianCalendar": return bcl.IsLeapYear(1) ? CalendarSystem.PersianSimple : CalendarSystem.PersianAstronomical;
                 case "UmAlQuraCalendar": return CalendarSystem.UmAlQura;
+                case "WondrousCalendar": return CalendarSystem.Wondrous;
```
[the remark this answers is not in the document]
glittle: Are all the other calendars in the BCL?  I was just trying to replicate what I was seeing being done for all the NodaTime calendars.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using System;
+using NodaTime.Calendars;
+using NodaTime.Text;
+using NUnit.Framework;
+using NodaTime.Calendars.Wondrous;
+
+namespace NodaTime.Test.Calendars
```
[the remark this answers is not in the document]
glittle: They all pass, and I tried to consider the edge cases.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+{
+    public class WondrousCalendarPeriodTest
```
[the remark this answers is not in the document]
glittle: Sure.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+
+        [Test]
+        public void BetweenLocalDates_MovingBackward_WithJustMonths()
+        {
+            Period actual = Period.Between(TestDate3_168_Ayyam_5, TestDate1_167_5_15, PeriodUnits.Months);
+            Period expected = Period.FromMonths(-32);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
```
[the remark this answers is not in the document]
glittle: I'm not sure which test is being referred to. The entire file?

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+        public void CreateDate_InAyyamiHa() {
+            var d1 = WondrousCalendarHelper.CreateDate(180, 0, 3);
+            var d3 = WondrousCalendarHelper.CreateDate(180, 18, 22);
+
+            Assert.AreEqual(d1, d3);
+        }
+
+
+        [Test]
+        public void CreateDate_Invalid()
```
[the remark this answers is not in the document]
glittle: Sure. Should I make a number of test files? None of the other calendars had more than one testing file.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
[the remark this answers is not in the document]
glittle: Agreed. That was left over from what I used as sample tests.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(180,1,-1));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(180,0,0));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(180,0,5));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(182,0,6));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(180,1,0));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(180,1,20));
+            Assert.Throws<ArgumentOutOfRangeException>(() => WondrousCalendarHelper.CreateDate(180,20,1));
+        }
+
+        //[Test]
```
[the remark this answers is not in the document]
glittle: Can remove it. I split this out into separate tests.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+
+            result = start + Period.FromDays(20);
+            Assert.AreEqual(WondrousCalendarHelper.CreateDate(182, 2, 2), result);
+        }
+
+        [Test]
+        public void Addition_DayCrossingMonthBoundaryFromAyyamiHa()
+        {
+            var start = WondrousCalendarHelper.CreateDate(182, AyyamiHaMonthNum, 3);
+            //            System.Diagnostics.Debugger.Launch();
```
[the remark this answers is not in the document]
glittle: Can do.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+            // in 182, Ayyam-i-Ha has 5 days
+            Assert.AreEqual(WondrousCalendarHelper.CreateDate(182, 19, 8), result);
+        }
+
+        [Test]
+        public void Addition_OneYearOnLeapDay()
+        {
+            LocalDate start = WondrousCalendarHelper.CreateDate(182, AyyamiHaMonthNum, 5);
+            LocalDate result = start + Period.FromYears(1);
+            // Ayyam-i-Ha 5 becomes Ayyam-i-Ha 4 ??
```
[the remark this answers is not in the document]
glittle: I can remove the ?? 
It was more a question of what should it do... what is 1 year after Feb 29? My answer is Feb 28.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
```diff
@@ -0,0 +1,255 @@
...
+            Period period = Period.FromYears(1) + Period.FromMonths(1) + Period.FromDays(2);
+            LocalDate start = WondrousCalendarHelper.CreateDate(171, 1, 19);
+            // Periods are added in order, so this becomes...
+            // Add one year: 172.1.19
+            // Add one month: 172.2.19
+            // Add two days: 172.3.2
+            LocalDate result = start + period;
+            Assert.AreEqual(WondrousCalendarHelper.CreateDate(172, 3, 2), result);
+        }
+
```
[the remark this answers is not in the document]
glittle: Can do.

src/NodaTime/CalendarSystem.cs
glittle: Sure.

src/NodaTime/CalendarSystem.cs
glittle: Sure.

src/NodaTime/CalendarSystem.cs
glittle: This is a case of monkey see, monkey do. It was trying to replicate what was done for other Calendars. Please adjust as desired.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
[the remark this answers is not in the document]
jskeet: I'd have one test file for things that are inherent to the calculator class, and another one for the helper.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Arbitrary. I'm not aware of precise guidelines for before year 1.  The embedded knowledge (below) is only defined to the year 1000, so I don't want to go beyond it for now. That is still 800 years in our future, so there should be time for an update before it becomes a serious concern!

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Likely copied.

src/NodaTime.Test/Calendars/BclCalendars.cs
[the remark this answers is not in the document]
jskeet: Yes, they are: see https://docs.microsoft.com/en-gb/dotnet/api/system.globalization?view=netframework-4.7

The purpose of this is so that we can check that our behaviour matches the BCL. Unless there's a WondrousCalendar in the BCL, we won't need it here.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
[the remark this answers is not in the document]
jskeet: Yes, that's our answer too.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: All the ConsoleWriteLine calls can be removed. I was using that before I was able to get a visual test runner to work.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Agreed.

jskeet:
I suggest you address as many of these as possible in the same branch (pull it locally, using -f if necessary due to my rebasing). When you're done, let me know and I can make modifications for the data part.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Sure. I was trying to keep track of whether I was dealing with a Gregorian year or a Wondrous year.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Can definitely try. Now that it seems that the calendar can be implemented in NodeTime, I can flush out some more tests. If we weren't going to be able to add it, there was no point in doing much more.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Sure.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Agreed.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: Sure.

src/NodaTime/LocalDate.cs
glittle: Sure. It was very confusing seeing the Wondrous dates being displayed as if they were Gregorian dates!

src/NodaTime.Test/Calendars/BclCalendars.cs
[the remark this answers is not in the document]
glittle: Sounds good...
So, who do I talk to get the Wondrous calendar added to the BCL?

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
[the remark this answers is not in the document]
glittle: Sounds good.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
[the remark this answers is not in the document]
glittle: Whew!

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Sure

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Sure

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Yes, now that this is going forward, I can be more verbose...

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Sounds good.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Sure, will be easy enough to do.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Not really, will do.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Adding "Get" sounds good. And I agree, "Num" is not needed as there is no other meaning for Month in this area.

Ideally, the "this" type would be WondrousCalendar, but that's not how the system works.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Unfortunately, something like this is needed. Because the internal representation of the date is not always what the "real" date is, the standard "ToString" can't be used.  

I didn't want to go much further down this path until I knew that the implementation would be viable.  I suspect that this will be the function most used by developers. There is not much need to "add 10 days" to the current month, etc. Most current uses will be "show me [Gregorian date] in the Wondrous calendar using this format".

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: You got it exactly.  The source I used is [http://www.bahaidate.today/table-of-dates](http://www.bahaidate.today/table-of-dates). I've converted that to an Excel file, which I can provide a bit later. That's how I generated this code.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
jskeet: In that case, it might be worth not having the extension methods. We could have the struct with either methods for converting values, or maybe have implicit conversions.

Another thing to *consider* is putting those types in a separate NuGet package. The calendar system would be in Noda Time itself (and would have to be) but the "slightly unusual" type could be in its own package which would avoid anyone accidentally using bits without wanting to.

Maybe we should get the calculator/calendar system in first, and hold off on everything else for now? (Once anything public is in, it can't be removed - that makes me nervous about any additions, basically. Caution is my watchword.)

glittle: Yes, that would make sense. Don't need to put a lot of code into the core system that could be in another package.

src/NodaTime.Test/Calendars/WondrousCalendarPeriodTest.cs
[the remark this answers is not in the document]
glittle: I have two test files. I could move all the tests from the "...PeriodTest" file into the existing "...SystemTest" file.  Is that preferable to having two files?

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
glittle: It is correct.

glittle:
@jskeet I think I've covered all the items that were identified. 

How should I proceed to make a different NuGet package to deal with formatting and using this calendar system?

jskeet:
Let's go one step at a time, getting this PR in first. We can then figure out whether the other package should be something hosted here, or potentially an entirely separate project you put on your own github repo.

I'll take a look at the changes when I get a chance - it should be some time this week.

glittle:
Thanks, @jskeet, sounds good.  

Another quick topic to discuss is around the religious holidays associated with the calendar. Just like how Easter is on a different Sunday each year in the Gregorian calendar, there is one set of holidays that Baha'is celebrate that are on different dates each year.  I've been very careful to consider only the mechanics of the calendar in this implementation, but many people using it will want to know when the Baha'i holidays are scheduled.  At most, I think this could be included in the "other package", or maybe it'll have to be in a 3rd package. As far as code is concerned, it is just another table of dates for the next 800 years.  Any ideas thought on how to balance the mechanics of a calendar with the religious use of it?

jskeet:
I think it would make sense for it to be part of the second package. Basically at that point, it's outside the Noda Time remit - and you may well want to make the name of the second package something entirely separate from Noda Time, even though it uses Noda Time under the covers.

glittle:
As for the "second package", could you point me to one or two packages that have done this well, that I might model this new one after? (No rush...)

jskeet:
Not really sure what you're looking for - it would just be a class library that has a dependency on Noda Time. https://github.com/mj1856/Dapper-NodaTime is one of those, for example.

glittle:
Sure, the Dapper-NodaTime link gives me a sense of how others have used Noda Time.  Doing a quick search in github on "using NodaTime;" gives me a few more.

src/NodaTime.Test/NodaTime.Test.csproj
```diff
@@ -9,6 +9,8 @@
     <Deterministic>True</Deterministic>
     <IsPackable>False</IsPackable>
     <DefineConstants Condition="$(TargetFramework.Contains('netcore'))">NETCORE</DefineConstants>
+    <ApplicationIcon />
```
jskeet: I'll remove these changes on this branch - there's no need for this file to change at all.

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,333 @@
...
+
+            if (month == Month19)
+            {
+                daysFromStartOfYearToStartOfMonth += WondrousCalendarHelper.GetYearInfo(year).DaysInAyyamiHa;
+            }
+
+            return daysFromStartOfYearToStartOfMonth;
+        }
+
+        /// <param name="year"></param>
```
jskeet: I wouldn't worry about XML docs for these overrides, unless there's anything specific to add. (I'll remove them.)

src/NodaTime/Calendars/WondrousYearMonthDayCalculator.cs
```diff
@@ -0,0 +1,333 @@
...
+
+            return isInAyyamiHa;
+        }
+
+        /// <param name="start"></param>
+        /// <param name="months"></param>
+        /// <remarks>
+        ///     Adding months around Ayyam-i-Ha is not well defined. Because they are "days outside of time", this
+        ///     implementation will skip over them.
+        ///     Can never get to Ayyam-i-Ha with this. From Month 18, will go to month 19.
```
jskeet: Although presumably you can get from Ayyam-i-Ha in one year to Ayyam-i-Ha in another year.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
```diff
@@ -0,0 +1,2204 @@
+using JetBrains.Annotations;
```
jskeet: I thought we were removing this class, as it would basically be moved into the second package?
I'm happy to refactor to move the parts that are needed for the calculator into that class - and to sort out the data generation. Is there any other need for this class?

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: I had no real use for creating a LocalDateTime, so have removed this.

src/NodaTime/Calendars/Wondrous/WondrousCalendarHelper.cs
glittle: Removed it entirely.

jskeet:
Just to check, I now see that WondrousCalendarHelper is gone - are you ready for me to have another look and then change the code to use a compact data representation, or are you still working on it? No hurry at all - I just don't want to do stuff before you're ready.

glittle:
Yes, it is ready for when you can get around to review it again.  I've already compacted the data :)

jskeet:
Even better :) Let's see if we can get it in in the next few days then... my evening is clear, so that should work well.

glittle:
Yup, that all sounds good.

jskeet:
Squashed into a single commit and removed the changes to NodaTime.Test.csproj - will merge on green and start a new PR for minor tweaks.

Issue #1026, opened 2017-11-13
Consider adding a CONTRIBUTING.md file


carlosschults:
...and/or updating the docs to provide a little bit more help for first-time contributors.

(The [developer guide](https://nodatime.org/developer/) has useful information, but IMHO could use a little improvement, particularly the "Building and Testing" page.)

Topics that could be covered:
- talk (briefly) about the project's structure, how things are organized, etc.
- usage and purpose of the attributes.
- tests: how to run, preferred naming conventions, etc
- general guidelines: how granular should PRs/commits be? Should every PR require an issue?, etc
- basic github tips (forking, cloning, setting an upstream, etc)

[Example of a nice contributing.md file](https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md)

jskeet:
Sorry, I didn't see this before for some reason.

Yes, the developer guide is fairly badly out of date. (One problem with more documentation is that whenever we change the structure a bit, I need to remember which files to update :)

I think the "basic github tips" may be a part I'd push back on - it feels like the kind of thing where we could guess which specific bits of git people aren't used to, but always get it wrong. Certainly something around the expectation being "fork in github, create a branch off master, create PR from that" would be worth doing though.

jskeet:
We might want to work out a way to show the same file (or most of it?) in the web site too...

carlosschults:
> One problem with more documentation is that whenever we change the structure a bit, I need to remember which files to update

I'd happily help with that :)

> Certainly something around the expectation being "fork in github, create a branch off master, create PR from that" would be worth doing though.

Yeah, that's basically what I had in mind.

So.....can I send a PR? :)

jskeet:
Of course! (If the PR deals with the dev guide as well, that would be even better - but that one may be more appropriate for me to tackle...)

carlosschults:
I guess I'll just deal with the CONTRIBUTING file for now, since it seems the easiest to quickly solve and get it out the way.

So, what do you think should definitely be in the file?

jskeet:
Your set of topics sounded reasonable to me - although I wouldn't say very much about the attributes, other than to mention that they exist. They probably merit a separate page (or just refer to the documentation).

carlosschults:
So, regarding the tests. I try to use NUnitTestAdapter (both the Visual Studio extension and the Nuget package) to run the tests and failed. 

[...]

PR #1036, opened 2017-11-29
Add contributing guidelines

carlosschults:
As discussed in #1026.

This is a first try, and I welcome suggestions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,22 +4,30 @@ Thanks for your interest in Noda Time. We appreciate all kinds of contributions,
 
 ## Basic Requirements
 
-If you want to contribute to the codebase, you must have Visual
-Studio 2017 installed - you can download the Community edition from
-[here.](https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx) (The other editions will work fine as well.)
+If you want to contribute to the codebase, you're going to need a text editor or IDE. We recommend [Visual Studio Community 2017](https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx)(Windows) or [Visual Studio Code](https://code.visualstudio.com/)(Windows/Linux/macOS).
 
 You're also going to need .NET Core SDK installed - you can download it from [here.](https://www.microsoft.com/net/download/)
 
-Please make sure you have a [git client](https://git-scm.com/) installed. If you don't already have a Github account, [please create one.](https://github.com/signup/free)
+Please make sure you have a [git client](https://git-scm.com/) installed. If you don't already have a GitHub account, [please create one.](https://github.com/signup/free)
 
 After you're all set, you can [fork the project](https://help.github.com/articles/fork-a-repo). Then you'll be able to clone your fork, so you can edit the files locally on your machine:
 
-`git clone https://github.com/YOUR-USERNAME/nodatime.git`
+```Text
+git clone https://github.com/YOUR-USERNAME/nodatime.git
+```
+
+Once you clone the repository, you'll have a [remote repository](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes) (or simply *remote*) called `origin`, that points to your forked repository on GitHub.
+
+You'll usually want to add another remote, pointing to the original repository on GitHub. It's an acepted convention to call this remote *upstream*. You can do it like this:
+
+```Text
+git remote add upstream https://github.com/nodatime/nodatime.git
+```
 
 ## How to start contributing?
 
 We have a [`help-wanted`](https://github.com/nodatime/nodatime/labels/help%20wanted)
-label on our issue tracker to indicate tasks which contributors can work on.
+label on our issue tracker to indicate tasks which new contributors can work on without much previous experience in Noda Time.
 
 If you've found something you'd like to help with, please leave a comment in the issue.
 
@@ -31,6 +39,7 @@ When you're ready to start working, create a new branch off the `master` branch:
 
 ```
 git checkout master
+git pull upstream master
 git checkout -b SOME-BRANCH-NAME
 ```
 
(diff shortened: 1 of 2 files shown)


jskeet:
Thanks - will try to review later today.

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
+# How to Contribute
+
+Thanks for your interest in Noda Time. We appreciate all kinds of contributions, from submitting issues to improving documentation; from writing tests to implementing new code. All help is welcome!
+
+## Basic Requirements
+
+If you want to contribute to the codebase, you must have Visual
```
jskeet: I think it would be more accurate to *recommend* using Visual Studio. You *need* the .NET Core SDK, but that's all - it's feasible to work on Noda Time just from the command line, or from VS Code.

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+Studio 2017 installed - you can download the Community edition from
+[here.](https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx) (The other editions will work fine as well.)
+
+You're also going to need .NET Core SDK installed - you can download it from [here.](https://www.microsoft.com/net/download/)
+
+Please make sure you have a [git client](https://git-scm.com/) installed. If you don't already have a Github account, [please create one.](https://github.com/signup/free)
```
jskeet: Nit: Github => GitHub.

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+After you're all set, you can [fork the project](https://help.github.com/articles/fork-a-repo). Then you'll be able to clone your fork, so you can edit the files locally on your machine:
+
+`git clone https://github.com/YOUR-USERNAME/nodatime.git`
```
jskeet: I'd probably use triple-backticks to show a code block. Three backticks on a line on their own (maybe with "text" following (without quotes)) to indicate that it's not code as such, then the code, then three backticks on a line on their own again. Sorry, it's slightly hard to demonstrate!

Would it be worth showing two remotes, one "upstream" and one "origin", just so we can demonstrate pulling and rebasing later?

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+## How to start contributing?
+
+We have a [`help-wanted`](https://github.com/nodatime/nodatime/labels/help%20wanted)
+label on our issue tracker to indicate tasks which contributors can work on.
```
jskeet: Maybe "to indicate tasks which contributors can work on without much previous experience in Noda Time"?

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+Additionally, feel free to open an issue if you find a bug or want to suggest a feature or enhancement.
+
+### Making Changes
+
+When you're ready to start working, create a new branch off the `master` branch:
+
+```
+git checkout master
+git checkout -b SOME-BRANCH-NAME
```
jskeet: "git pull upstream master" to make sure it's up to date?

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+Try to use a short, descriptive name for your branch, such as `add-tests-foobar-struct`.
+
+### Building
+
+To build everything under Visual Studio, simply open the src\NodaTime-All.sln solution file and build normally. To build with just the .NET Core SDK, run
+
+> dotnet restore src\NodaTime-All.sln
```
jskeet: No need for dotnet restore any more, as of the v2 SDK :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+> 
+> dotnet build src\NodaTime-All.sln
```
jskeet: Maybe use a forward-slash which will work under both Unix and Windows?

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Running Tests
+
+The tests are currently console application projects. Simply run the following commands:
+
+> cd src\NodaTime.Test
```
jskeet: Again, forward-slash?

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+git checkout master
+git checkout -b SOME-BRANCH-NAME
+```
+
+Try to use a short, descriptive name for your branch, such as `add-tests-foobar-struct`.
+
+### Building
+
+To build everything under Visual Studio, simply open the src\NodaTime-All.sln solution file and build normally. To build with just the .NET Core SDK, run
+
```
jskeet: It's put all of these commands in triple-backtick blocks too

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+> dotnet build src\NodaTime-All.sln
+
+### Running Tests
+
+The tests are currently console application projects. Simply run the following commands:
+
+> cd src\NodaTime.Test
+> dotnet run -f net451
+> dotnet run -f netcoreapp1.0
+
```
jskeet: It might be worth mentioning that a few tests are really slow. You can run the vast majority with

```
dotnet test -f netcoreapp1.0 -- --where=cat!=Slow
```

(Ditto for net451)

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+### Submitting Changes
+
+To publish your branch to your local fork, run this command from the Git Shell:
+
+`git push origin -u MY-BRANCH-NAME`
```
jskeet: I don't use -u myself, but I gather it's based on all kinds of settings, so if this works reliably, that's fine :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+When you're done, [open a pull request](https://help.github.com/articles/using-pull-requests) against your changes.
+
+If you're pull request fixes an issue, add a comment with the word "Fixes", "Resolves" or "Closes", followed by the issue's number:
```
jskeet: you're => your

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+>   Fixes #1145
+
+If you need to, feel free to add comments to the PR asking for sugestions or help.
```
jskeet: sugestions => suggestions

CONTRIBUTING.md
carlosschults: Poorly worded. Will fix, thanks.

CONTRIBUTING.md
carlosschults: Sounds great!

CONTRIBUTING.md
carlosschults: Nice :)

But in this case, the site needs an update as well, right?

CONTRIBUTING.md
carlosschults: That's useful, I didn't know about that. Will change.

CONTRIBUTING.md
carlosschults: Are you sure? Because I really meant something like "When you are done with whatever you are doing..."

CONTRIBUTING.md
jskeet: Not sure what you mean - do you mean the developer part of the web site? If so, then yes - that's quite out of date and I should really get to it at some point. (That should either pick up some text here, or possibly just refer to here...)

CONTRIBUTING.md
jskeet: It was the "If you're pull request" part that's wrong. "When you're done" is fine. Sorry for not having been clearer before!

carlosschults: Oh, I'm sorry! I had not even seen that one.

Issue #1037, opened 2017-11-29
Consider adding DateInterval.Intersection/Union etc


carlosschults:
In the same spirit of #977, maybe it would make sense to add methods such as the following to the `DateInterval` type:

- `bool Contains(DateInterval other)`
- `bool Intersects(DateInterval other)`
- `DateInterval? Intersection(DateInterval other)`
- `DateInterval Union(DateInterval other)`

jskeet:
Possibly - it should be simpler as DateInterval can't be infinite.

carlosschults:
Nice. Can I give it a try?

jskeet:
Absolutely, but it might be worth waiting until all the decisions for #977 have been taken - unless you think none of the open questions there apply to DateInterval, which is entirely feasible.

carlosschults:
> Absolutely, but it might be worth waiting until all the decisions for #977 have been taken - unless you think none of the open questions there apply to DateInterval, which is entirely feasible.

I'm not sure if the questions there  really apply here. I mean, there is no such a thing as an empty `DateInterval`, right?

> (Perhaps Union should return DateInterval? too? Otherwise non-overlapping date intervals could create problems.)

Got it.

jskeet:
Yup, it's entirely possible that none of the corner cases apply here. Implement away :)

(One thing to be careful of is different calendar systems, though. That will bring a whole different set of issues.)

malcolmr:
> I'm not sure if the questions there really apply here. I mean, there is no such a thing as an empty `DateInterval`, right?

Correct (and as Jon mentions, they're bounded), so I don't think #977 applies. I think there are still a few things to be worked out though:

- Would something like "subset-of" be a better description that "contains", for the relationship between two intervals? We already have `DateInterval.Contains(LocalDate)`; is `Contains(DateInterval)` sufficiently similar that an overload makes sense, or should we have a different method name? ("Subset" itself is also potentially confusing because it may be interpreted as "proper subset".)
- Should the union of two disjoint intervals return `DateInterval?` (as Jon suggests), or should it throw an exception?
- Do we need another method to return the smallest interval covering two (possibly-disjoint) intervals, or is that unlikely to be useful?
[...]

PR #1038, opened 2017-11-30
Add set operations to DateInterval type

carlosschults:
First try at adressing #1037.

Implemented `Contains`, `Intersects`, `Intersection`, but haven't implemented `Union` yet.

### DIFF
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -399,5 +399,83 @@ namespace NodaTime.Test
 
             Assert.True(year2017.Intersects(november2017));
         }
+
+        [Test]
+        public void Intersection_NullInterval_Throws()
+        {
+            var value = new DateInterval(new LocalDate(100), new LocalDate(200));
+            Assert.That(() => value.Intersection(null), Throws.TypeOf<ArgumentNullException>());
+        }
+
+        [Test]
+        public void Intersection_IntervalInDifferentCalendar_Throws()
+        {
+            var value = new DateInterval(
+                new LocalDate(2017, 11, 6, CalendarSystem.Gregorian),
+                new LocalDate(2017, 11, 10, CalendarSystem.Gregorian));
+
+            var other = new DateInterval(
+                new LocalDate(2017, 11, 6, CalendarSystem.Coptic),
+                new LocalDate(2017, 11, 10, CalendarSystem.Coptic));
+
+            Assert.That(() => value.Intersection(other), Throws.TypeOf<ArgumentException>());
+        }
+
+        [Test]
+        public void Intersection_EqualIntervals()
+        {
+            var value = new DateInterval(new LocalDate(100), new LocalDate(200));
+            var other = new DateInterval(new LocalDate(100), new LocalDate(200));
+
+            Assert.AreEqual(value, value.Intersection(other));
+        }
+
+        [Test]
+        public void Intersection_EarlierInterval()
+        {
+            var value = new DateInterval(new LocalDate(2014, 3, 7), new LocalDate(2014, 3, 31));
+            var other = new DateInterval(new LocalDate(2013, 6, 26), new LocalDate(2013, 7, 25));
+
+            Assert.Null(value.Intersection(other));
+        }
+
+        [Test]
+        public void Intersection_LaterInterval()
+        {
+            var value = new DateInterval(new LocalDate(2013, 6, 26), new LocalDate(2013, 7, 25));
+            var other = new DateInterval(new LocalDate(2014, 12, 7), new LocalDate(2015, 1, 1));
+
+            Assert.Null(value.Intersection(other));
+        }
+
+        [Test]
+        public void Intersection_IntersectingAtTheStart()
+        {
+            var value = new DateInterval(new LocalDate(2017, 11, 3), new LocalDate(2017, 11, 29));
+            var other = new DateInterval(new LocalDate(2017, 11, 1), new LocalDate(2017, 11, 12));
+            var expected = new DateInterval(new LocalDate(2017, 11, 3), new LocalDate(2017, 11, 12));
+
+            Assert.AreEqual(expected, value.Intersection(other));
+        }
+
+        [Test]
+        public void Intersection_IntersectingAtTheEnd()
+        {
+            var value = new DateInterval(new LocalDate(2017, 11, 3), new LocalDate(2017, 11, 25));
+            var other = new DateInterval(new LocalDate(2017, 11, 18), new LocalDate(2017, 11, 30));
+            var expected = new DateInterval(new LocalDate(2017, 11, 18), new LocalDate(2017, 11, 25));
+
+            Assert.AreEqual(expected, value.Intersection(other));
+        }
+
+        [Test]
+        public void Intersection_Subset()
+        {
+            var year2017 = new DateInterval(new LocalDate(2017, 1, 1), new LocalDate(2017, 12, 31));
+            var november2017 = new DateInterval(new LocalDate(2017, 11, 1), new LocalDate(2017, 11, 30));
+            var expected = november2017;
+
+            Assert.AreEqual(expected, year2017.Intersection(november2017));
+        }
     }
 }
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -315,5 +315,89 @@ namespace NodaTime.Test
 
             Assert.False(november2017.Contains(year2017));
         }
+
+        [Test]
+        public void Intersects_NullInterval_Throws()
+        {
+            var value = new DateInterval(new LocalDate(100), new LocalDate(200));
+            Assert.That(() => value.Intersects(null), Throws.TypeOf<ArgumentNullException>());
+        }
+
+        [Test]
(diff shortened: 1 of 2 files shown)


src/NodaTime/DateInterval.cs
```diff
@@ -134,6 +134,25 @@ public bool Contains(LocalDate date)
             return Start <= date && date <= End;
         }
 
+        /// <summary>
```
jskeet: Possibly make explicit mention that an interval contains itself?

src/NodaTime/DateInterval.cs
```diff
@@ -167,5 +186,70 @@ public void Deconstruct(out LocalDate start, out LocalDate end)
...
+        /// <returns>
+        /// A <see cref="DateInterval"/> corresponding to the intersection between the given interval and the current
+        /// instance. If there is no intersection, a null reference is returned.
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        /// The start and end dates of <paramref name="interval" /> are not in the same calendar
+        /// as the start and end date of this interval.
+        /// </exception>
+        [CanBeNull]
+        public DateInterval Intersection([NotNull]DateInterval interval)
```
jskeet: It feels like we could get rid of `Intersects` and just make this return null if there's no intersection.

(I hadn't spotted that DateInterval is a class where Interval is a struct. I can't remember offhand why we did that.)

src/NodaTime/DateInterval.cs
```diff
@@ -167,5 +186,70 @@ public void Deconstruct(out LocalDate start, out LocalDate end)
...
+        /// </summary>
+        /// <param name="interval">The <see cref="DateInterval"/> to check for intersection with this interval.</param>
+        /// <returns>
+        /// <c>true</c> if <paramref name="interval"/> has days in common with this interval; <c>false</c> otherwise.
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        /// The start and end dates of <paramref name="interval" /> are not in the same calendar
+        /// as the start and end date of this interval.
+        /// </exception>
+        public bool Intersects([NotNull]DateInterval interval)
```
jskeet: Nit - please add a space after the attribute: `[NotNull] DateInterval interval`

src/NodaTime/DateInterval.cs
```diff
@@ -167,5 +186,70 @@ public void Deconstruct(out LocalDate start, out LocalDate end)
...
+        {
+            Preconditions.CheckNotNull(interval, nameof(interval));
+
+            Preconditions.CheckArgument(
```
jskeet: It may well be worth adding:

- A private helper method to assert that an interval is non-null and in the same calendar system
- A `Calendar` property which just returns `Start.Calendar`

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -208,5 +208,274 @@ public void Deconstruction()
...
         }
+
+        [Test]
+        public void Contains_NullInterval_Throws()
+        {
+            var start = new LocalDate(2017, 11, 6);
+            var end = new LocalDate(2017, 11, 10);
+            var value = new DateInterval(start, end);
+
+            Assert.That(() => value.Contains(null), Throws.TypeOf<ArgumentNullException>());
```
jskeet: Any reason not to use `Assert.Throws<ArgumentException>(() => value.Contains(null))`? That's what we typically do... I only use `Assert.That` where it adds significant value, otherwise it feels like fluff to me.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -208,5 +208,274 @@ public void Deconstruction()
...
+        [Test]
+        public void Contains_EarlierInterval()
+        {
+            var value = new DateInterval(new LocalDate(2014, 3, 7), new LocalDate(2014, 3, 31));
+            var other = new DateInterval(new LocalDate(2013, 6, 26), new LocalDate(2013, 7, 25));
+
+            Assert.False(value.Contains(other));
+        }
+
+        [Test]
```
jskeet: We could potentially use parameterized tests for all of these. I'd probably take strings representing the start/end intervals, and the expected results. (Strings rather than lots of integers to help with readability.)

src/NodaTime.Test/DateIntervalTest.cs
carlosschults: No reason in particular, it's just the way I'm more used to, so I kind of do it this way automatically. I'm going to change it :)

src/NodaTime.Test/DateIntervalTest.cs
carlosschults: Got it.

src/NodaTime/DateInterval.cs
carlosschults: OK. Should I put this inside `remarks` or just on the main text?

src/NodaTime/DateInterval.cs
carlosschults: Yeah, I thought of doing something like that, but ended up not doing.

src/NodaTime/DateInterval.cs
carlosschults: Ok, I have a question here about nullability and the use of the attributes.

What exactly are they used for in Noda? I'm aware that there are tests that break when I don't use an attribute where I was supposed to, but is this it?

Do some of the CI tools perform statical analysis such that I get a broken build or warning in cases like the following:

```
var x = obj.SomeMethodThatCanReturnNull();
x.FooBar(); // oops, risk of NRE
```
(Sorry if this is outside the scope of the PR, but I thought it'd make sense to ask here.)

src/NodaTime/DateInterval.cs
jskeet: Either works for me :)

src/NodaTime/DateInterval.cs
jskeet: Currently, there's very little checking done. I did start writing a Roslyn analyzer, but it ended up being very, very hard to do with any precision. Resharper knows about the attributes and can warn on them though.

We use NotNull for parameters to automate exception documentation.

When C# 8 comes out, hopefully a lot of this can be more obvious from the types, and checked more by the compiler. Until then, it's mostly informational.

src/NodaTime/DateInterval.cs
```diff
@@ -167,5 +193,51 @@ public void Deconstruct(out LocalDate start, out LocalDate end)
...
+        private void ValidateInterval(DateInterval interval)
+        {
+            var msg = "The start and end dates of the interval to check " +
+                "must be in the same calendar as the start and end dates of this interval.";
+
+            Preconditions.CheckNotNull(interval, nameof(interval));
+            Preconditions.CheckArgument(interval.Calendar.Equals(Start.Calendar), nameof(interval), msg);
+        }
+
+        private bool ContainsExtreme(DateInterval interval) =>
```
jskeet: It looks like this isn't needed any more

src/NodaTime/DateInterval.cs
```diff
@@ -167,5 +193,51 @@ public void Deconstruct(out LocalDate start, out LocalDate end)
...
+        /// <exception cref="ArgumentException">
+        /// The start and end dates of <paramref name="interval" /> are not in the same calendar
+        /// as the start and end date of this interval.
+        /// </exception>
+        [CanBeNull]
+        public DateInterval Intersection([NotNull]DateInterval interval)
+        {
+            ValidateInterval(interval);
+
+            if (interval == this || Contains(interval))
```
jskeet: Nit: could we have braces around these conditions? Or possibly a big set of conditional expressions:

```csharp
return interval == this || Contains(interval) ? interval
    : interval.Contains(this) ? this
    : interval.Contains(this.Start) ? new DateInterval(interval.Start, End)
    : interval.Contains(this.End) ? new DateInterval(Start, interval.End)
    : null;
```

src/NodaTime/DateInterval.cs
```diff
@@ -134,6 +134,24 @@ public bool Contains(LocalDate date)
...
+        /// <remarks>
+        /// An interval contains another interval with same start and end dates, or itself.
+        /// </remarks>
+        /// <param name="interval">The interval to check for containment within this interval.</param>
+        /// <exception cref="ArgumentException">Start and end dates of <paramref name="interval"/> are not in the same
+        /// calendar as the start and end date of this interval.</exception>
+        /// <returns><c>true</c> if <paramref name="interval"/> is within this interval; <c>false</c> otherwise.</returns>
+        public bool Contains([NotNull] DateInterval interval)
+        {
+            ValidateInterval(interval);
```
jskeet: Possibly optimize for `interval == this`? Then you don't need to do it in Intersection

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -208,5 +218,95 @@ public void Deconstruction()
...
+        {
+            var value = ParseInterval(firstInterval);
+            var other = ParseInterval(secondInterval);
+            var expectedResult = ParseInterval(expectedInterval);
+            Assert.AreEqual(expectedResult, value.Intersection(other));
+        }
+        
+        private DateInterval ParseInterval(string textualInterval)
+        {
+            if (textualInterval == null)
```
jskeet: I'll fix this up myself after merging - I don't want to keep asking for tiny things :)

PR #1043, opened 2017-12-07
Add union dateinterval

### LINKED ISSUE #1037: Consider adding DateInterval.Intersection/Union etc

carlosschults:
Closes #1037.

### DIFF
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -318,18 +318,22 @@ namespace NodaTime.Test
             Assert.Throws<ArgumentException>(() => value.Union(other));
         }
 
-        [TestCase("2014-03-07,2014-03-10", "2014-03-12,2014-03-17", null)]
-        [TestCase("2014-03-07,2014-03-10", "2014-03-11,2014-03-17", "2014-03-07,2014-03-17")]
-        [TestCase("2014-03-07,2014-03-20", "2014-03-07,2014-03-20", "2014-03-07,2014-03-20")]
-        [TestCase("2017-01-01,2017-03-31", "2017-02-15,2017-06-26", "2017-01-01,2017-06-26")]
-        [TestCase("2017-02-15,2017-06-26", "2017-01-01,2017-03-31", "2017-01-01,2017-06-26")]
+        [TestCase("2014-03-07,2014-03-20", "2015-03-07,2015-03-20", null, Description = "Disjointed intervals")]
+        [TestCase("2014-03-07,2014-03-20", "2014-03-21,2014-03-30", "2014-03-07,2014-03-30", Description = "Abutting intervals")]
+        [TestCase("2014-03-07,2014-03-20", "2014-03-07,2014-03-20", "2014-03-07,2014-03-20", Description = "Equal intervals")]
+        [TestCase("2014-03-07,2014-03-20", "2014-03-15,2014-03-23", "2014-03-07,2014-03-23", Description = "Overlapping intervals")]
+        [TestCase("2014-03-07,2014-03-20", "2014-03-10,2014-03-15", "2014-03-07,2014-03-20", Description = "Interval completely contained in another")]
         public void Union(string first, string second, string expected)
         {
             DateInterval firstInterval = ParseInterval(first);
             DateInterval secondInterval = ParseInterval(second);
             DateInterval expectedResult = ParseInterval(expected);
 
-            Assert.AreEqual(expectedResult, firstInterval.Union(secondInterval));
+            Assert.Multiple(() =>
+            {
+                Assert.AreEqual(expectedResult, firstInterval.Union(secondInterval), "First union failed.");
+                Assert.AreEqual(expectedResult, secondInterval.Union(firstInterval), "Second union failed.");
+            });
         }
 
         private DateInterval ParseInterval(string textualInterval)
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -296,7 +296,42 @@ namespace NodaTime.Test
             var expectedResult = ParseInterval(expectedInterval);
             Assert.AreEqual(expectedResult, value.Intersection(other));
         }
-        
+
+        [Test]
+        public void Union_NullInterval_Throws()
+        {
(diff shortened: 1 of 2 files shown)


src/NodaTime/DateInterval.cs
```diff
@@ -214,6 +214,31 @@ public DateInterval Intersection([NotNull]DateInterval interval)
                 : null;
         }
 
+        /// <summary>
+        /// Returns the union between the given interval and this interval, as long as they're overlapping or contiguous.
+        /// </summary>
+        /// <param name="interval">
+        /// The specified interval to unite with this one.
```
jskeet: Not sure whether this is a little ambiguous - it sounds like it *might* mutate this one. I'm struggling to find better wording though...

(As an aside, we generally collapse simple tags like this onto a single line, although it doesn't matter.)

src/NodaTime/DateInterval.cs
```diff
@@ -214,6 +214,31 @@ public DateInterval Intersection([NotNull]DateInterval interval)
...
+        /// calendar to this date interval.</exception>
+        [CanBeNull]
+        public DateInterval Union([NotNull] DateInterval interval)
+        {
+            ValidateInterval(interval);
+
+            var start = LocalDate.Min(Start, interval.Start);
+            var end = LocalDate.Max(End, interval.End);
+
+            return Period.Between(start, end).Days >= Length + interval.Length
```
jskeet: It's try to avoid using Period.Between here - that could do quite a bit of arithmetic.

I'll have a think about the best way of detecting this.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -296,7 +296,42 @@ public void Intersection(string firstInterval, string secondInterval, string exp
...
+        [TestCase("2014-03-07,2014-03-20", "2014-03-07,2014-03-20", "2014-03-07,2014-03-20")]
+        [TestCase("2017-01-01,2017-03-31", "2017-02-15,2017-06-26", "2017-01-01,2017-06-26")]
+        [TestCase("2017-02-15,2017-06-26", "2017-01-01,2017-03-31", "2017-01-01,2017-06-26")]
+        public void Union(string first, string second, string expected)
+        {
+            DateInterval firstInterval = ParseInterval(first);
+            DateInterval secondInterval = ParseInterval(second);
+            DateInterval expectedResult = ParseInterval(expected);
+
+            Assert.AreEqual(expectedResult, firstInterval.Union(secondInterval));
```
jskeet: I'd add another test of

```csharp
Assert.AreEqual(expectedResult, secondInterval.Union(firstInterval))
```

That sort of helps to document that it's commutative, and we don't need to specify too many test cases.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -296,7 +296,42 @@ public void Intersection(string firstInterval, string secondInterval, string exp
...
+                new LocalDate(2017, 11, 10, CalendarSystem.Gregorian));
+
+            var other = new DateInterval(
+                new LocalDate(2017, 11, 6, CalendarSystem.Coptic),
+                new LocalDate(2017, 11, 10, CalendarSystem.Coptic));
+
+            Assert.Throws<ArgumentException>(() => value.Union(other));
+        }
+
+        [TestCase("2014-03-07,2014-03-10", "2014-03-12,2014-03-17", null)]
```
jskeet: I think you can use Description="..." to give descriptions to each test case. That would definitely help here :)

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -296,7 +296,42 @@ public void Intersection(string firstInterval, string secondInterval, string exp
...
+        [TestCase("2014-03-07,2014-03-10", "2014-03-11,2014-03-17", "2014-03-07,2014-03-17")]
+        [TestCase("2014-03-07,2014-03-20", "2014-03-07,2014-03-20", "2014-03-07,2014-03-20")]
+        [TestCase("2017-01-01,2017-03-31", "2017-02-15,2017-06-26", "2017-01-01,2017-06-26")]
```
jskeet: It would probably be simpler to reason about the tests if they all had the same first interval - just vary the second one.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -296,7 +296,42 @@ public void Intersection(string firstInterval, string secondInterval, string exp
...
+        [TestCase("2017-02-15,2017-06-26", "2017-01-01,2017-03-31", "2017-01-01,2017-06-26")]
```
jskeet: We can get rid of this one with my suggestion below.

src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -296,7 +296,42 @@ public void Intersection(string firstInterval, string secondInterval, string exp
...
+        public void Union(string first, string second, string expected)
```
jskeet: I think we're missing a test for "one interval within another". I'd expect tests of:

- Disjoint
- Abutting
- Equal
- Overlapping
- One entirely contained within another

*Maybe* we could have tests where they share the same start date but different end dates and vice versa, but that's probably unnecessary.

src/NodaTime/DateInterval.cs
jskeet: And actually, this will be incorrect, as it will fail if there are multiple years and months in the period - you'd need to specify PeriodUnits.Days. At that point, that's actually reasonably cheap (IIRC) - so maybe that's the right approach to take.

(We may want to write an internal Period.DaysBetween method to make this and Length even cheaper, but that's a different matter.)

jskeet: Having looked at this, feel free to use Period.Between but making sure you specify PeriodUnits.Days, and I'll fix it up to be more efficient later.

Before you change it to be right though, it would be good to write a test that shows it's broken: I suspect if you create two date intervals for the same dates but different years, then the union will be non-null at the moment, but should be null.

PR #1046, opened 2017-12-07
Tweaks to the Wondrous calendar system

### LINKED ISSUE #1044: Make the Wondrous calendar consistent about Ayyam-i-Ha
Some parts of the code still use 0 as the month for Ayyam-i-Ha.

We should make the Noda Time calendar system consistent about using 18 for Ayyam-i-Ha everywhere, then the secondary package can use 0 consistently, with no leak of the abstractions or danger to other Noda Time code.

I'm doing this now, but may not finish tonight.

jskeet:
- Remove WondrousYearInfo: we never need both parts together,
  so just use two separate methods instead.
- Remove the cache: a dictionary is going to be more involved
  than the simple array lookup anyway, and it's not thread-safe.
- Simplify CalculateStartOfYearDays - aside from anything else,
  now we don't need to use DateTime :)
- Change WondrousYearMonthDayCalculator to *only* use the
  "Ayyam-i-Ha are part of month 18" model, e.g. for GetDaysInMonth.
  The alternative model of "Ayyam-i-Ha are in month 0" is used
  extensively in WondrousCalendarSystemTest. Fixes #1044.

@glittle Please could you review? It's nice to see that the calculator is now only 276 lines :)

### DIFF
--- a/src/NodaTime.Test/Calendars/WondrousCalendarSystemTest.cs
+++ b/src/NodaTime.Test/Calendars/WondrousCalendarSystemTest.cs
@@ -1,26 +1,26 @@
-// Copyright 2011 The Noda Time Authors. All rights reserved.
+// Copyright 2017 The Noda Time Authors. All rights reserved.
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
 using NodaTime.Calendars;
+using NodaTime.Utility;
 using NUnit.Framework;
 using System;
 
 namespace NodaTime.Test.Calendars
 {
-    public static class WondrousCalendarHelper {
+    public static class WondrousCalendarHelper
+    {
         /// <summary>
         /// Get a text representation of the date.
         /// </summary>
-        /// <param name="input"></param>
-        /// <returns></returns>
         public static string AsWondrousString(this LocalDate input)
         {
             // minimal formatting
 
             var year = input.Year;
-            var month = WondrousYearMonthDayCalculator.WondrousMonth(input);
-            var day = WondrousYearMonthDayCalculator.WondrousDay(input);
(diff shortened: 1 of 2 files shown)


glittle:
Sounds good. I'll work on these improvements, hopefully soon.

jskeet:
When you say work on them, it's just reviewing - I'm not asking you to write any more code :)

glittle:
But, but... I want to improve it!  :) 

"...use two separate methods..." sounds like a good idea that I can do.

jskeet:
But it's already done - that's my point. The code is there, awaiting your review. There's no point in you doing the same thing again. If you want to *further* improve it (or suggest changes to what I've done) that makes sense, but this pull request is primarily code - the comment at the top is just a log of what it consists of.

glittle:
Ahh! Thanks! I was just getting the code now. Will review...

glittle:
It all looks good. Thanks!
I have a further improvement that I'll commit soon.

jskeet:
Great - will merge this PR. You can then pull master, then create a new branch for your PR based on that, so there are no conflicts.

glittle:
I've switched to jskeet/nodatime/master.  The updated test file is there, but the `WondrousYearMonthDayCalculator.cs` file seems to be missing. Any suggestions?

jskeet:
Don't use `jskeet/nodatime/master`. Use `nodatime/nodatime/master`. You shouldn't need to use my fork at all.

glittle:
Much better. Now I just need to update to .NET 4.7 to be able to compile it.  :(

jskeet:
You shouldn't need to update your version of .NET, although you do need a C# 7 compiler (e.g. VS2017). ValueTuple should be restored automatically, as it's a dependency of the test project.

glittle:
Yup. A full Rebuild solved it.

PR #1047, opened 2017-12-08
Benchmarks and optimizations for DateInterval

jskeet:


src/NodaTime/DateInterval.cs
```diff
@@ -158,7 +158,7 @@ public bool Contains([NotNull] DateInterval interval)
         public int Length =>
             // Period.Between will give us the exclusive result, so we need to add 1
```
carlosschults: Just a tiny nitpick, but I think you should update the method name to `DaysBetween` in the comment, right?

jskeet: Whoops, yes :) Are you happy with this apart from that?

carlosschults: Yep, sure :)

PR #1048, opened 2017-12-08
update test file

glittle:
- simplify the test file by removing the extension method
- test the helper functions


jskeet:
LGTM, thanks.

glittle:
I don't have a need for it at the moment, but if and when I want to make another change, is there any convention about the name of the branch that I create to do it? Should I keep using the same name each time or ensure that it is always different?

jskeet:
Either is fine - whichever works best for you. Personally I delete my local branches as soon as I've merged, just to keep it obvious what I'm still working on.

PR #1085, opened 2018-03-23
Documentation for OffsetTime and OffsetDate

### LINKED ISSUE #864: Consider OffsetTime and OffsetDate types

jskeet:
Minimal, but enough to specify text formats etc.

Fixes #864 - and unblocks a 2.3.0 release.

### DIFF
--- a/src/NodaTime.Web/Markdown/unstable/core-types.md
+++ b/src/NodaTime.Web/Markdown/unstable/core-types.md
@@ -1,0 +1,143 @@
+@Title="Core types quick reference"
+
+This is a companion page to the
+["core concepts"](concepts), and ["choosing between types"](type-choices)
+pages, describing the fundamental types in Noda Time very briefly, primarily for reference.
+If you're not familiar with the core concepts, read that page first.
+
+Instant
+-------
+
+An [`Instant`][Instant] is a point on notional a global time-line, regardless of calendar system and time zone.
+It's simply a number of nanoseconds since some arbitrary epoch. Noda Time always uses the Unix epoch, which corresponds to midnight on January 1st 1970 UTC.
+(This is merely one way of expressing the epoch - it would be equally valid to express it using other calendar systems
+and time zones; the epoch itself has no notion of a time zone or calendar system.)
+
+Offset
+------
+
+An [`Offset`][Offset] is used to express the difference between UTC and local time. It is always *added*
+to a UTC value to obtain a local time, or *subtracted* from a local time to obtain a UTC value. Values
+are between 18 hours before UTC and 18 hours after UTC, inclusive. `Offset` has a granularity of
+one second; any attempt to create an offset with a value including subseconds is simply truncated to
+the second.
+
+CalendarSystem
+--------------
+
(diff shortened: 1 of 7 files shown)


src/NodaTime.Web/Markdown/unstable/core-types.md
```diff
@@ -0,0 +1,143 @@
+@Title="Core types quick reference"
```
malcolmr: Sigh, git doesn't record where copies come from, so I had to clone the patch locally and run diff with `--find-copies-harder` to get a good diff.

src/NodaTime.Web.Test/NodaTime.Web.Test.csproj
```diff
@@ -14,5 +14,9 @@
     <PackageReference Include="NUnit" Version="3.6.1" />
     <PackageReference Include="NUnitLite" Version="3.6.1" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
malcolmr: (This is https://github.com/Microsoft/vstest/issues/472, and while there's a lot of discussion there about whether this makes sense, it's probably fine to leave it as-is.)

src/NodaTime.Web/NodaTime.Web.csproj
```diff
@@ -24,7 +24,6 @@
     <PackageReference Include="morelinq" Version="2.10.0" />
   
     <!-- Handling of non-code folders -->
-    <Folder Include="Markdown\" />
```
malcolmr: I assume this was redundant with the below?

src/NodaTime.Web.Test/NodaTime.Web.Test.csproj
jskeet: Ah, missed this one. I removed it again from all the rest of the projects before merging. Think I'll do so again here.

src/NodaTime.Web/NodaTime.Web.csproj
jskeet: Apparently so. I didn't do that myself, so presumably VS was just being clever...

PR #1086, opened 2018-03-23
Improve documentation for period arithmetic

### LINKED ISSUE #1084: LocalDate + Period addition is not documented as adding period components individually
new LocalDate(2016, 2, 29).Plus(Period.FromYears(1) + Period.FromMonths(1))

My initial expectation was that this would produce Mar 29, 2017. Instead, it produces Mar 28, 2017, because it adds a year first, and only then adds a month. I'd have to normalise this to a period of 13 months (with the understanding that this doesn't work in other calendar systems) to get the result I was expecting.

For `LocalDateTime`, this is sort of documented:

https://nodatime.org/2.2.x/api/NodaTime.LocalDateTime.html

> **Addition(LocalDateTime, Period)**  
> Adds a period to a local date/time. Fields are added in the order provided by the period. This is a convenience operator over the Plus(Period) method.

and

> **Plus(Period)**  
> Adds a period to this local date/time. Fields are added in the order provided by the period.

I don't see what "the order provided by the period" is, but at least it tells that there's a gotcha.

For `LocalDate`, this documentation is missing:

https://nodatime.org/2.2.x/api/NodaTime.LocalDate.html

> **Addition(LocalDate, Period)**  
> Adds the specified period to the date. 

and

> **Plus(Period)**  
> Adds the specified period to this date. Fluent alternative to operator+().

jskeet:
Fixes #1084.

### DIFF
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -291,6 +291,7 @@ namespace NodaTime
 
         /// <summary>
         /// Adds the specified period to the date.
+        /// Fields are added in descending order of significance (years first, then months, and so on).
         /// </summary>
         /// <param name="date">The date to add the period to</param>
         /// <param name="period">The period to add. Must not contain any (non-zero) time units.</param>
@@ -303,7 +304,9 @@ namespace NodaTime
         }
 
         /// <summary>
-        /// Adds the specified period to the date. Friendly alternative to <c>operator+()</c>.
+        /// Adds the specified period to the date.
+        /// Fields are added in descending order of significance (years first, then months, and so on).
+        /// Friendly alternative to <c>operator+()</c>.
         /// </summary>
         /// <param name="date">The date to add the period to</param>
         /// <param name="period">The period to add. Must not contain any (non-zero) time units.</param>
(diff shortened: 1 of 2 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -291,6 +291,7 @@ public static LocalDate FromYearMonthWeekAndDay(int year, int month, int occurre
 
         /// <summary>
         /// Adds the specified period to the date.
+        /// Fields are added in descending order of significance (years first, then months etc).
```
malcolmr: nit: surely "months, etc.)."? (comma before, period after) (and throughout)

src/NodaTime/LocalDate.cs
```diff
@@ -329,6 +334,7 @@ public static LocalDate FromYearMonthWeekAndDay(int year, int month, int occurre
 
         /// <summary>
         /// Subtracts the specified period from the date.
+        /// Fields are subtracted in descending order of significance (years first, then months etc).
```
malcolmr: Incidentally, this surprises me. I guess I would have expected that `date + period - period == date`, but that presumably isn't true, given this?

src/NodaTime/LocalDate.cs
jskeet: I'll change to "and so on" to avoid the double period.

src/NodaTime/LocalDate.cs
jskeet: That isn't always true even for a single field. Consider January 30th + 1 month - 1 month - that will always result in January 28th or January 29th. Sigh.

PR #1092, opened 2018-03-25
Automatic project changes

jskeet:
These are provided by Visual Studio. I believe they should be
removable in VS15.7, but until then it's simpler to have a single
commit that we can revert later than to let the changes accidentally
end up in other commits.


glittle:
I was getting some of those added in my local code and avoided committing them. It [appears](https://stackoverflow.com/a/22917902/32429) to be related to test projects.

PR #1094, opened 2018-03-28
Update NUnit

jskeet:
3 stages:

- Update the project files
- Fix a CultureSaver issue under netcoreapp2.0 with recent NUnit
- Fix a FakeDateTimeZoneSource bug and the test

### DIFF
--- a/src/NodaTime.Test/Testing/TimeZones/FakeDateTimeZoneSourceTest.cs
+++ b/src/NodaTime.Test/Testing/TimeZones/FakeDateTimeZoneSourceTest.cs
@@ -129,15 +129,17 @@ namespace NodaTime.Test.Testing.TimeZones
         // Unfortuntely NUnit requires the exact exception type :(
         private static void AssertBuildFails(FakeDateTimeZoneSource.Builder builder)
         {
+            bool success = false;
             try
             {
                 builder.Build();
-                Assert.Fail("Expected exception");
+                success = true;
             }
             catch (Exception)
             {
                 // Expected
             }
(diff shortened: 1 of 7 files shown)


src/NodaTime.Test/Testing/TimeZones/FakeDateTimeZoneSourceTest.cs
```diff
@@ -129,15 +129,17 @@ public void GetEnumerator()
         // Unfortuntely NUnit requires the exact exception type :(
         private static void AssertBuildFails(FakeDateTimeZoneSource.Builder builder)
         {
+            bool success = false;
             try
             {
                 builder.Build();
-                Assert.Fail("Expected exception");
```
malcolmr: Nice fix. But how is this related to the NUnit upgrade? (I don't mind either way, just curious.)

jskeet: While we were still catching the exception in the latest NUnit, it was being reported as a failure - I suspect that there's some out-of-band way that it records errors, even if an exception isn't thrown.

jskeet: (I haven't dug into it though - that's just a guess.)

malcolmr: Doh, and you even said as much in the commit message, which I did read.

Thanks for explaining (again!)

PR #1100, opened 2018-04-23
Created FileLoader utility class

oriash93:
Contains LoadFileOrUrl method, solves code duplication

### DIFF
--- a/src/NodaTime/Utility/FileLoader.cs
+++ b/src/NodaTime/Utility/FileLoader.cs
@@ -1,0 +1,35 @@
+﻿// Copyright 2011 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using System.IO;
+using System.Net.Http;
+
+namespace NodaTime.Utility
+{
+    /// <summary>
+    /// Provides method to help with loading file or URL.
+    /// </summary>
+    public static class FileLoader
+    {
+        /// <summary>
+        /// Loads a given file or URL, and returns its contents as a byte array.
+        /// </summary>
+        /// <param name="source">Path to the source</param>
+        /// <returns></returns>
+        public static byte[] LoadFileOrUrl(string source)
+        {
+            if (source.StartsWith("http://") || source.StartsWith("https://") || source.StartsWith("ftp://"))
+            {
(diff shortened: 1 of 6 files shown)


jskeet:
The CI failures are due to missing annotations. I can probably fix those up in a separate commit after merging. Will review this when I get a chance.

src/NodaTime/Utility/FileLoader.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.IO;
+using System.Net.Http;
+
+namespace NodaTime.Utility
+{
+    /// <summary>
+    /// Provides method to help with loading file or URL.
+    /// </summary>
+    public static class FileLoader
```
jskeet: I'm reluctant to put this into NodaTime itself. It's not used within NodaTime - you've only included it here because it's a common dependency, right?

src/NodaTime/Utility/FileLoader.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <param name="source">Path to the source</param>
+        /// <returns></returns>
+        public static byte[] LoadFileOrUrl(string source)
+        {
+            if (source.StartsWith("http://") || source.StartsWith("https://") || source.StartsWith("ftp://"))
+            {
+                using (var client = new HttpClient())
+                {
+                    // I know using .Result is nasty, but we're in a console app, and nothing is
```
jskeet: And this is the kind of reason I don't want to include it in the public API of the main library :) We don't know we in a console app any more.

src/NodaTime/Utility/FileLoader.cs
oriash93: Yes. Where else would you include it?

src/NodaTime/Utility/FileLoader.cs
oriash93: Can you think of a different implementation?

jskeet: I would rather just not put it into the NodaTime assembly. I'll create a new project as a common one for tools, and use that. As such, it's probably worth closing this PR - but thanks for the nudge.

oriash93: You're welcome. Feel free to close it

PR #1101, opened 2018-04-24
Using C# 6 String Interpolation feature

oriash93:
Also clearing some whitespace


lib/commandline/CommandLineText.cs
```diff
@@ -56,39 +56,39 @@ public static BaseSentenceBuilder CreateBuiltIn ()
         {
```
jskeet: I'm avoiding changing this file as far as I can. It was another library (which has now moved on in a way which isn't helpful to us) and I'd like to keep it as close to the original source as possible.

oriash93:
Reverted that file.

PR #1108, opened 2018-04-30
Adjust web site styling

jskeet:
- Use blue/yellow/white for side-bar nav for user guide
- Move side-bar nav to left hand side
- Use side-bar rather than accordian for the user guide
- Remove the max-width set by Foundation

### DIFF
--- a/src/NodaTime.Web/Views/Documentation/Docs.cshtml
+++ b/src/NodaTime.Web/Views/Documentation/Docs.cshtml
@@ -2,32 +2,26 @@
 
 @{ ViewBag.Title = Model.Title; }
 
-<section class="body">
-    <div class="row">
-        <div class="large-9 columns">
-            <h1>@Model.Title</h1>
-            @Model.Content
-        </div>
-
-        <div class="large-3 columns">
-            <ul class="accordion" data-accordion>
-                @foreach (var category in Model.Bundle.Categories)
-                {
-                    <li class="accordion-navigation">@category.Title</li>
-                    <div class="content">
-                    <ul class="side-nav">
-                    @foreach (var contentPage in category.Pages)
-                    {
-                        <li><a href="@contentPage.Id">@contentPage.Title</a></li>
-                    }
-                    </ul>
-                    </div>
-                }
-            </ul>
-            @if (Model.Bundle.Name != "developer")
+<div class="row">
+    <div class="large-3 columns">
+        <ul class="side-nav">
+            @foreach (var category in Model.Bundle.Categories)
             {
-                <footer>Version @Model.Bundle.Name</footer>
+                <li class="heading">@category.Title</li>
+                @foreach (var contentPage in category.Pages)
+                {
+                    <li class="content"><a href="@contentPage.Id">@contentPage.Title</a></li>
+                }
             }
-        </div>
+        </ul>
+        @if (Model.Bundle.Name != "developer")
+        {
+            <footer class="version">Version @Model.Bundle.Name</footer>
+        }
     </div>
-</section>
+
+    <div class="large-9 columns">
+        <h1>@Model.Title</h1>
+        @Model.Content
+    </div>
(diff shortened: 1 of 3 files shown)


jskeet:
Merging - we can roll back if it looks horrible :)

src/NodaTime.Web/wwwroot/css/site.css
```diff
@@ -260,29 +265,27 @@ footer .design {
...
-  padding: 10px 16px;
-  font-size: 12px;
-  border: 0px none;
+.side-nav {
+  padding-bottom: 0;
 }
 
-.accordion-navigation {
+.side-nav > .heading {
   background-color: #FFE76B;
```
malcolmr: It's good practice to specify `color` and `background-color` together unless you really want to inherit one. This is picking up a default (#008cba) from Foundation, from what I can see.

Also, that combination isn't super readable to my eyes (and it's got a contrast ratio < the recommended for small text). Perhaps specify #006b8f here instead? It's the same hue, but higher contrast.

src/NodaTime.Web/wwwroot/css/site.css
```diff
@@ -260,29 +265,27 @@ footer .design {
...
 
-.accordion .side-nav > li > a,
-.accordion .side-nav > li > a:visited,
-.accordion .side-nav > li > a:hover,
-.accordion .side-nav > li > a:focus {
-  color: #FFF;
-  padding-top: 0;
-  padding-bottom: 0;
+.side-nav > li > a,
+.side-nav > li > a:not(.button),
```
malcolmr: I think this may need to be e.g. `a:not(.button):focus` (etc) in order to have the same priority as the Foundation-default rules it's overriding. At present, we get a light-blue text colour.

src/NodaTime.Web/Views/Shared/_Layout.cshtml
```diff
@@ -129,7 +129,6 @@
     <script src="/js/vendor/modernizr.js"></script>
     <script src="/js/vendor/jquery.js"></script>
     <script src="/js/foundation/foundation.js"></script>
-    <script src="/js/foundation/foundation.accordion.js"></script>
```
malcolmr: Make the corresponding change in `src/NodaTime.Web/bundleconfig.json` so that we remove it from the minified JS (site.min.js) too?

src/NodaTime.Web/Views/Documentation/Docs.cshtml
```diff
@@ -2,32 +2,26 @@
 
 @{ ViewBag.Title = Model.Title; }
 
-<section class="body">
```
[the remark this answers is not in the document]
jskeet: It looked like it was pointless, in terms of having a section as well as a div. Let's have a look and decide whether we actively want it.

src/NodaTime.Web/Views/Shared/_Layout.cshtml
jskeet: Sure, will do.

src/NodaTime.Web/wwwroot/css/site.css
jskeet: Happy to.

src/NodaTime.Web/wwwroot/css/site.css
jskeet: This was to override just `a:not(.button)` - without that, it looked awful, though I can't remember in what way. I'm happy with the light blue for the value we're hovering over, unless that doesn't have enough contrast.

src/NodaTime.Web/Views/Documentation/Docs.cshtml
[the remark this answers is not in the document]
jskeet: Just looked again, and it's already in the layout, so we were getting one nested within another. I've removed another copy of it as well now :)

PR #1115, opened 2018-05-04
Make DateInterval iterable

### LINKED ISSUE #1111: Could DateInterval implement IEnumerable<LocalDate>?
I'm currently working on something where I want to foreach over all the dates between a start and end.

It would be useful if I could use the DateInterval class to do this, rather than having to write a helper function to do it. The benefit of this is that DateInterval already ensures the dates are in the correct order (start <= end) and in the same calendar. It seems (to me, anyway) to be a natural operation on this object.

This could be done using a property on the class:

`IEnumerable<LocalDate> Dates { get; }`

or by making the class implement IEnumerable<LocalDate>:

`class DateInterval : IEnumerable<LocalDate>`

jskeet:
Fixes #1111.

### DIFF
--- a/src/NodaTime.Test/DateIntervalTest.cs
+++ b/src/NodaTime.Test/DateIntervalTest.cs
@@ -4,6 +4,7 @@
 using NodaTime.Text;
 using NUnit.Framework;
 using System;
+using System.Linq;
 
 namespace NodaTime.Test
 {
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/DateIntervalTest.cs
```diff
@@ -331,6 +332,30 @@ public void Union(string first, string second, string expected)
             });
         }
 
+        [Test]
```
malcolmr: Add a test to verify we can iterate over a minimal (one-day) `LocalDateInterval` too?

jskeet: Good call. Rewritten to be a single data-driven test :)

PR #1117, opened 2018-05-06
Fix issues when TimeZoneInfo.Local is null

jskeet:
This was found when trying out C# 8 nullable references.
TimeZoneInfo.Local has been observed to return null in some Mono
installations. We should handle that as reasonably as we can. We
already did for BclDateTimeZoneSource, but now we do in
TzdbDateTimeZoneSource too. Added tests for everywhere we need to
use TimeZoneInfo.Local (via the interceptor).

### DIFF
--- a/src/NodaTime.Test/TimeZones/BclDateTimeZoneSourceTest.cs
+++ b/src/NodaTime.Test/TimeZones/BclDateTimeZoneSourceTest.cs
@@ -116,9 +116,7 @@ namespace NodaTime.Test.TimeZones
             using (TimeZoneInfoReplacer.Replace(null, systemZone))
             {
                 var source = new BclDateTimeZoneSource();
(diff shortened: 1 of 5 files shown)


src/NodaTime.Test/TimeZones/BclDateTimeZoneSourceTest.cs
```diff
@@ -116,9 +116,7 @@ public void LocalZoneIsNull()
             using (TimeZoneInfoReplacer.Replace(null, systemZone))
             {
                 var source = new BclDateTimeZoneSource();
-                CollectionAssert.AreEqual(new[] { systemZone.Id }, source.GetIds().ToList());
```
jskeet: (This bit isn't strictly needed for the PR, but was just tidying up.)

PR #1118, opened 2018-05-08
Annual date pattern

### LINKED ISSUE #1114: Annual Date Patterns
It doesn't seem like there are any patterns for `AnnualDate`s and most of the general parsing logic seems to be hidden inside internal methods. Are there any plans to add any `AnnualDate` patterns, and if not, what is the easiest way to implement it outside the library?

jskeet:
Best to review one commit at a time.

Fixes #1114.

Happy to reconsider the character to use for the single standard pattern.


src/NodaTime/Text/AnnualDatePattern.cs
```diff
@@ -0,0 +1,213 @@
...
+        internal static readonly AnnualDate DefaultTemplateValue = new AnnualDate(1, 1);
+
+        private const string DefaultFormatPattern = "G"; // General, ISO-like
+
+        internal static readonly PatternBclSupport<AnnualDate> BclSupport =
+            new PatternBclSupport<AnnualDate>(DefaultFormatPattern, fi => fi.AnnualDatePatternParser);
+
+        /// <summary>
+        /// Gets an invariant local date pattern which is compatible with the month/day part of ISO-8601.
+        /// This corresponds to the text pattern "MM'-'dd".
```
malcolmr: Note that `javax.time.MonthDay` uses `--MM-dd` (for the debugging-oriented `toString()`), presumably to avoid confusion with their `YearMonth`. We don't have the latter, and even if we did, I don't think we need need our default patterns to be non-overlapping.

(Surprisingly, ISO 8601 appears to have nothing to say about formatting a month and day, even in the [imo ridiculous] recent update draft.)

src/NodaTime/Text/AnnualDatePattern.cs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        internal static readonly AnnualDate DefaultTemplateValue = new AnnualDate(1, 1);
+
+        private const string DefaultFormatPattern = "G"; // General, ISO-like
+
+        internal static readonly PatternBclSupport<AnnualDate> BclSupport =
+            new PatternBclSupport<AnnualDate>(DefaultFormatPattern, fi => fi.AnnualDatePatternParser);
+
+        /// <summary>
+        /// Gets an invariant local date pattern which is compatible with the month/day part of ISO-8601.
```
malcolmr: Not a "local date pattern"? See also a number of similar occurrences below.

src/NodaTime/Text/AnnualDatePattern.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// the argument being null are wrapped in a parse result.
+        /// </remarks>
+        /// <param name="text">The text value to parse.</param>
+        /// <returns>The result of parsing, which may be successful or unsuccessful.</returns>
+        [NotNull] public ParseResult<AnnualDate> Parse([SpecialNullHandling] string text) => UnderlyingPattern.Parse(text);
+
+        /// <summary>
+        /// Formats the given local date as text according to the rules of this pattern.
+        /// </summary>
+        /// <param name="value">The local date to format.</param>
```
malcolmr: e.g.

src/NodaTime/Text/AnnualDatePattern.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// <summary>
+        /// Creates a pattern for the given pattern text, culture, and template value.
+        /// </summary>
+        /// <remarks>
+        /// See the user guide for the available pattern text options.
+        /// </remarks>
+        /// <param name="patternText">Pattern text to create the pattern for</param>
+        /// <param name="cultureInfo">The culture to use in the pattern</param>
+        /// <param name="templateValue">Template value to use for unspecified fields</param>
+        /// <returns>A pattern for parsing and formatting local dates.</returns>
```
malcolmr: Also e.g.

jskeet:
Thanks for all the comments - will address then merge. (Will probably add another commit to add this to the version history too.)

PR #1124, opened 2018-05-15
Fix docfx

### LINKED ISSUE #1123: Namespace is System.Dynamic.ExpandoObject for all types in docs
Quite possibly a docfx bug; will need to check what's going on.

jskeet:
Should fix #1123.

Admittedly I haven't tested locally, which makes me nervous, but we can roll back if it breaks anything...


build/docfx/template/partials/class.tmpl.partial
```diff
@@ -174,12 +163,6 @@
 {{#example}}
 {{{.}}}
 {{/example}}
-{{#snippet}}
```
malcolmr: Did you mean to remove this one?

jskeet: Whoops, nope. Will fix.

PR #1129, opened 2018-05-19
Initial changes for 3.0

jskeet:
Best reviewed one commit at a time.

There may well be more work required to get everything stable in CI and on Bagpuss, e.g. changing the scheduled task to run benchmarks.


jskeet:
(I'll fix things up until CI is happy, then leave you to review :)

jskeet:
Okay, it looks like we need to do some work on .NET Core handling of TimeZoneInfo, but that can be done separately. (Given how early this is in the 3.0 timeline, I'm okay with CI not passing for a little while.)

Ready for review.

malcolmr:
Looks reasonable from here.

(Are we removing XML serialization too?)

jskeet:
The current plan is to keep XML serialization, as it's significantly less dangerous than binary serialization, or at least it *can* be used safely.

Issue #1033, opened 2017-11-16
Investigate C# 7.2 benefits


jskeet:
It's not clear to me what the versioning implications are yet, but this could be good.

https://docs.microsoft.com/en-gb/dotnet/csharp/reference-semantics-with-value-types

oriash93:
I'm not sure if C# 7 features are used, but I made a PR for inline out variables (#1098).
Also, in lack of a better issue to mention this, I also applied Remove and Sort Usings throught the solution, so another PR (#1099). 😃

jskeet:
Thanks for the PRs. We can definitely use C# 7 features - the SDK is currently pinned to 2.1.4, but I can update that if necessary.

I'll try to get to the PRs in the next couple of days - sorry not to be able to review them immediately.

oriash93:
@jskeet All good. If you're already at it, take a look at #1100.
I saw a TODO comment about a method being rewritten exactly the same 3-4 times, so I extracted it to a helper class.
(Sorry again for the lack of a better place to mention this 😀)

oriash93:
I swear 😆 #1104

jskeet:
Readonly structs have immediately improved the performance of some benchmarks that were using static readonly fields. We may want to use `in` parameters within the implementation (but not in the API, as I view them as dangerous) but that requires further investigation.

jskeet:
Just tried using `in` parameters for formatting, and it made things worse - but it's possible that I didn't do it hard enough...

jskeet:
After trying harder, it doesn't make things worse any more - but there's almost no benefit, and the code is much more complex. Closing as "we've done all we need to for C# 7.2 benefits for now."

oriash93:
>Closing as "we've done all we need to for C# 7.2 benefits for now."

What about C# 7.3?

jskeet:
@oriash93: I don't think there's anything in C# 7.3 that particularly affects Noda Time.

PR #1130, opened 2018-05-19
Make structs readonly

jskeet:
This is the first pass at #1033.

Once #1129 is merged, this is a single commit.

IXmlSerializable is fundamentally broken for readonly structs, but we can cast away the readonly-ness using the System.Runtime.CompilerServices.Unsafe package. It feels like the benefits of this are greater than the downsides.


malcolmr:
LGTM. Are there any threading or 'undefined behaviour'-style issues with the use of `Unsafe.AsRef()`?

jskeet:
Basically we're breaking the readonly contract, with all the fun that could entail. If this is used by anyone outside regular XML serialization, I'd expect problems. Within XML serialization, it should be fine - this it all called on new objects.

I'm going to wait until we've got the first benchmarks working for v3 before merging.

PR #1136, opened 2018-05-20
Handle running on GCP transparently

jskeet:


src/NodaTime.Web/Providers/GoogleCredentialProvider.cs
```diff
@@ -16,14 +16,25 @@ public class GoogleCredentialProvider
     {
         public static GoogleCredential FetchCredential(IConfiguration configuration)
         {
+            string foo = configuration["foo"];
```
malcolmr: Um?

jskeet: Whoops. That's left over as a test that the indexer returns null if the key's not found. Will remove :)

PR #1137, opened 2018-05-21
Remove test variable and add a TODO

jskeet:


src/NodaTime.Web/Providers/GoogleCredentialProvider.cs
```diff
@@ -30,6 +29,7 @@ public static GoogleCredential FetchCredential(IConfiguration configuration)
 
             // But if none of the environment variables has been specified, expect that
             // we're running on Google Cloud Platform with implicit credentials.
+            // TODO: Autodetect if we're on GCP and only use the default credentials then.
```
malcolmr: I'm not sure this is true: running locally is a use case where we generally also want to use the default credentials.

(That's why we introduced `GOOGLE_APPLICATION_CREDENTIALS`. Now that the Azure configuration is optional, we could also remove L19-24 above, and use this fallback. Unless there are cases where you want to use the default credentials where the Azure variables are _also_ available?)

jskeet: Sorry, yes - this was meant to be "if we don't have the environment variable set, check that we really are on GCP". Basically four options:

- GOOGLE_APPLICATION_CREDENTIALS defined: use that
- Settings for Azure Key Value: use those
- No environment variables, but running on GCP: default credentials
- None of the above? Go bang with a decent exception

malcolmr: That's alright, that's how I read it.

What I was thinking of was the simpler (but perhaps more opaque):

* Settings for Azure KeyVault: use those
* otherwise use default credentials

... but your plan is good too :)

PR #1150, opened 2018-05-31
Update .NET Core SDK to 2.1.300

jskeet:
(Aka .NET Core 2.1)

This should improve build times, and let us benchmark .NET Core 2.0 vs .NET Core 2.1 easily.

(This may well fail on AppVeyor initially. Will need to decide whether to merge anyway to unblock benchmarks...)


jskeet:
Looks like we need to wait for Travis scripts to be updated too. Will have another look in a week...

.travis.yml
```diff
@@ -3,7 +3,7 @@
 # TODO: Test against Mono as well, using matrix.include
 language: csharp
 mono: none
-dotnet: 2.1.101
+dotnet: 2.1
```
malcolmr: According to https://docs.travis-ci.com/user/languages/csharp/#.NET-Core, Travis actually wants the SDK version here. Maybe try `2.1.300`?

jskeet: Tried that earlier. There are problems, basically. https://github.com/travis-ci/travis-ci/issues/9684 is the relevant Travis issue - I'm sure it'll be resolved swiftly :)

jskeet:
Travis looks like it's now working. AppVeyor isn't yet; I'll keep an eye on https://github.com/appveyor/ci/issues/2373

I may well merge when Travis goes green though.

PR #1160, opened 2018-06-05
(Experimental) A couple of optimizations

jskeet:
Advice welcome on the second commit. (The first is trivial.)

- Can you think of useful benchmarks that might show this to be a good thing?
- If we can't detect any benefit but suspect there should be one, does it make sense to merge?

I was planning on merging to see whether the benchmarks *do* show anything.


jskeet:
Ack. Will merge and we can have a look at the results tomorrow.

PR #1163, opened 2018-06-10
Reformat the layout file slightly

jskeet:


jskeet:
Note that there's a downside here, in that when the HTML is rendered it will be "double-indented" for the bits that are per-environment. I suspect we've got a lot of that effect anyway in Razor - let's focus on usability when writing the code rather than the HTML.

PR #1172, opened 2018-06-12
Add Try .NET to the Recipes page

jskeet:


src/NodaTime.Web/Views/Shared/_Layout.cshtml
```diff
@@ -132,6 +132,7 @@
         <script src="/js/foundation/foundation.topbar.js"></script>
         <script src="/js/highlight.pack.js"></script>
         <script src="/js/nuget-button.js"></script>
+        <script src="/js/trydotnet.js"></script>
```
malcolmr: I think you'll also have to add it somewhere so that it gets built into the minified version, won't you?

PR #1179, opened 2018-06-16
Experimental change to OffsetDateTime

jskeet:
This used to affect performance, but that was before readonly
structs etc. The code is definitely simpler using LocalDate.

Next step would be to try to use OffsetTime instead of the long, but
that's more complicated. Let's see what the performance impact of
this is first.

### DIFF
--- a/src/NodaTime/Instant.cs
+++ b/src/NodaTime/Instant.cs
@@ -670,8 +670,7 @@ namespace NodaTime
         public ZonedDateTime InUtc()
         {
             // Bypass any determination of offset and arithmetic, as we know the offset is zero.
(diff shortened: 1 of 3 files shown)


jskeet:
For .NET Core, there's no significant difference with this PR.
For .NET 4.6.1, the "before" code is much faster for Year/Month/Day etc, but slower for Date. But the difference is between (say) 0.3ns and 0.01ns... so it's pretty fast either way. (On .NET Core it's about 1.5ns.)

Optimizing for .NET 4.6.1's inlining (which I assume this is) makes sense where there's no readability benefit either way, but I think the change here is worth the loss in performance for clarity. We might want to see where else uses YearMonthDayCalendar and could use LocalDate instead.

I'll start looking at using OffsetTime for the other part of OffsetDateTime, at which point it would be a much simpler struct.

PR #1181, opened 2018-06-17
Manual fix for is-null

jskeet:
This change was created with find/replace. Ideally I'd like to use a Roslyn analyzer in the future to find these...


jskeet:
(Won't merge this until tomorrow, to avoid it interfering with the benchmark results of #1180.)

PR #1191, opened 2018-08-16
Document the TZDB update policy

### LINKED ISSUE #1042: Determine and document support lifetime policy
In particular, we *may* need the following to be handled differently:

- Minor (in impact) bugfixes
- Major (in impact) bugfixes
- TZDB updates

Initial proposal:

We expect users within a major version to update within a year of a new minor version becoming available. After that time, the previous minor version is unsupported. For example, 1.4.0 was released on July 8th 2017, so from July 8th 2018 we won't supply new patch versions of 1.3.x.

We can probably say that minor bugfixes will only be applied to the current minor version in each major version, whereas major bugfixes and TZDB updates will be applied to all the supported minor versions.

The point at which we decide not to support 1.x at all is a whole other decision.

jskeet:
Fixes #1042

### DIFF
--- a/src/NodaTime.Web/Markdown/root/versions.md
+++ b/src/NodaTime.Web/Markdown/root/versions.md
@@ -4,6 +4,9 @@ User-visible changes from 1.0.0-beta1 onwards. See the
 [project repository](https://github.com/nodatime/nodatime) for more
 details.
 
+See the [end of this page](#tzdb-updates) for the policy on which
+versions receive patch updates for TZDB releases.
+
 ## 2.3.0, released 2018-05-19 with tzdb 2018e
 
 New features:
@@ -708,3 +711,16 @@ Bug fixes:
 ## 1.0.0-beta1, released 2012-04-12 with tzdb 2012c
 
 - Initial beta release
+
+# <a id="tzdb-updates"></a>Update policy for time zone data
+
+New time zone data is released by IANA as required. The latest data
+is always available via the Noda Time web site as described in the
+[user guide](/userguide/tzdb). Additionally, NuGet packages are
+updated with a patch release to include the latest data.
+
+The latest minor release for each major release is always updated.
+Additionally, if the latest minor release is less than 6 months old, the
+previous minor release is also updated. In other words, if you check
+once every 6 months to make sure you're on the latest minor release,
+you should always get patch releases containing TZDB updates.


src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -708,3 +711,16 @@ Bug fixes:
 ## 1.0.0-beta1, released 2012-04-12 with tzdb 2012c
 
 - Initial beta release
+
+# <a name="tzdb-updates"></a>Update policy for time zone data
```
malcolmr: prefer `<a id="">` to `<a name="">`.

(Also, I assume this actually DTRT in the rendered version?)

src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -708,3 +711,16 @@ Bug fixes:
...
+
+New time zone data is released by IANA as required. The latest data
+is always available via the Noda Time web site as described in the
+[user guide](/userguide/tzdb). Additionally, NuGet packages are
```
malcolmr: Maybe:

```
in the [user guide]().

Additionally, released versions of the NodaTime NuGet package are also updated...

(next para)
```

src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -708,3 +711,16 @@ Bug fixes:
...
+updated with a patch release to include the latest data.
+
+The latest minor release for each major release is always updated.
```
malcolmr: That's just a complicated way of saying "the latest release", isn't it? :)

src/NodaTime.Web/Markdown/root/versions.md
```diff
@@ -708,3 +711,16 @@ Bug fixes:
...
+Additionally, if a minor release is less than 6 months old, the
```
malcolmr: ", if the latest minor release is..."?

src/NodaTime.Web/Markdown/root/versions.md
jskeet: Didn't know about this. Will check both forms.

src/NodaTime.Web/Markdown/root/versions.md
jskeet: I think I'd rather not have two single-line paragraphs.

src/NodaTime.Web/Markdown/root/versions.md
jskeet: Nope - "1.4" isn't the latest release, but it's the latest minor release of a major release.

src/NodaTime.Web/Markdown/root/versions.md
jskeet: Yup

src/NodaTime.Web/Markdown/root/versions.md
malcolmr: I'm apparently unable to read. Thanks for the clarification :)

PR #1216, opened 2018-10-28
Order time zones by ID when guessing the system zone

### LINKED ISSUE #1207: Order time zones by ID before guessing the local time zone
In [this Stack Overflow question](https://stackoverflow.com/questions/52608545) the user receives an unexpected IANA ID. That's fine: their local time zone is ambiguous when guessing by transitions. However, it's also non-deterministic: it relies on the order in which a dictionary returns its value. If we order the time zones before guessing, the result becomes deterministic until the input data (which includes the current year in UTC) changes.

jskeet:
Fixes #1207

(No unit tests as the test would be far more involved than the change, and it's pretty risk-free. Let me know if you disagree :)


jskeet:
Ooh, will check that failure, thanks

PR #1232, opened 2018-12-08
Update to using the C# 8 preview

jskeet:
Travis and AppVeyor may fail at this point. Bagpuss certainly will when it comes to building docs, but that'll be easy to fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 # TODO: Test against Mono as well, using matrix.include
 language: csharp
 mono: none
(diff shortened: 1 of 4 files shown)


jskeet:
Merging anyway. I would generally strongly recommend against merging anything that breaks CI, but:

- We won't be releasing anything from master for a long time
- I have an ulterior motive in getting NodaTime onto C# 8, so I can properly investigate the new features for talks :)
- Keeping things in separate branches is feasible but extra work
- I don't think it will affect other contributors

Issue #154, opened 2015-03-15
Provide TypeConverters for Duration and Instant


GoogleCodeExporter:
```
*What situation does this request make simpler?*

    This would make porting existing code that makes use of TypeConverters for System.DateTime and System.TimeSpan over to NodaTime simpler (generic string conversions for example).

*Can you provide a straw-man example of what you'd want the API to look
like?*

    No direct API is necessary. I believe it's a matter of implementing the TypeConverters in NodaTime, and decorating the Instant and Timespan classes with the TypeConverter attribute. 

What I would like to see is the following:

```

TypeConverter converter = TypeDescriptor.GetConverter(typeof(Instant));
Assert.IsTrue(converter.CanConvertFrom(typeof(string)));
Assert.IsTrue(converter.CanConvertTo(typeof(string)));
Assert.IsTrue(converter.CanConvertFrom(typeof(System.DateTimeOffset)));
Assert.IsTrue(converter.CanConvertTo(typeof(System.DateTimeOffset)));
string dateTimeString = "..."; // <- this would be any string accepted by 
DateTimeOffset.Parse()
Instant instant = converter.ConvertFrom(dateTimeString);

```

For durations, the same thing as above would apply, but the string would be any 
valid System.Timespan string.
```

Original issue reported on code.google.com by `bryanjam...@gmail.com` on 29 Nov 2012 at 9:48

austindrenski:
@jskeet Is anyone currently working on this?

jskeet:
I'm not - I don't know enough about TypeConverters to make the right decisions. If you do (or want to research until you do) that would be great.

austindrenski:
I'll take a look and see if I can put together a PR in the next few days.

For some context: 

Initially, I'm looking to support a `TypeConverter` for `LocalDate`. The @npgsql project integrates NodaTime through a plugin model for ADO.NET and EF Core. PostgreSQL range types (e.g. `daterange`) are mapped to CLR range types (e.g. `NpgsqlRange<LocalDate>`). 

The `TypeConverter` for `NpgsqlRange<T>` assumes that a `TypeConverter` for `T` will be available for bound parsing, and when one isn't available, the operation throws. (This impacts various aspects of the MVC framework, model binding, etc.)

sfmskywalker:
Any updates on this?

austindrenski:
@sfmskywalker I still need to write some unit tests, but you can take a look at #1237 for the general idea.

[...]

PR #1237, opened 2018-12-11
Add type converters for ISO pattern parsers

austindrenski:
Closes: #154

## Background

@austindrenski [wrote](https://github.com/nodatime/nodatime/issues/154#issuecomment-420920072):
> Initially, I'm looking to support a `TypeConverter` for `LocalDate`. The @npgsql project integrates NodaTime through a plugin model for ADO.NET and EF Core. PostgreSQL range types (e.g. `daterange`) are mapped to CLR range types (e.g. `NpgsqlRange<LocalDate>`).
> 
> The `TypeConverter` for `NpgsqlRange<T>` assumes that a `TypeConverter` for `T` will be available for bound parsing, and when one isn't available, the operation throws. (This impacts various aspects of the MVC framework, model binding, etc.)

## Goals

The goal is to add type converters where they make sense, such as for types that have ISO patterns that are (reasonably) safe to use for roundtrip serialization. 

No effort is made to handle non-ISO calendars on write, because there would be no way to detect and handle them on read.

## Status

~~I still need to add some unit tests, but wanted to open this up for some feedback.~~

## Related
- https://docs.microsoft.com/dotnet/api/system.componentmodel.typeconverter#notes-to-inheritors


src/NodaTime/Text/AnnualDateTypeConverter.cs
```diff
@@ -0,0 +1,44 @@
+using System;
```
jskeet: Please add a copyright notice at the top of each new source file.

austindrenski:
> Will do a full review later when I get a chance, but at a glance it looks like there's a fair amount of repeated code here, that could be avoided using an abstract base class accepting a pattern in the constructor.

I'll take care of the copyright notices and refactoring with an abstract base later today, so feel free to wait on the full review.

> We _could_ backport this as a Noda Time 2.5 release, but I'd _prefer_ not to.

That's works for me. I don't think there's anything here to justify a backport. Plus, anyone who needs these converters in the meantime should be able to add them at runtime with a bit of work.

jskeet:
> That's works for me. I don't think there's anything here to justify a backport. Plus, anyone who needs these converters in the meantime should be able to add them at runtime with a bit of work.

I'd love to add some documentation for that somewhere. I'm happy to do most of the writing, but I don't know enough about how type converters work to get it started - how do you add type converters at execution time?

austindrenski:
It should be something along the lines of:

```c#
TypeDescriptor.AddAttributes(
    typeof(LocalDate),
    new TypeConverterAttribute(typeof(LocalDateTypeConverter)));
```

The [docs](https://docs.microsoft.com/dotnet/api/system.componentmodel.typedescriptor.addattributes#System_ComponentModel_TypeDescriptor_AddAttributes_System_Type_System_Attribute___) have some helpful information in the remarks section, and there's also an implementation in [`NpgsqlRange<T>`](https://github.com/npgsql/npgsql/blob/dev/src/Npgsql/NpgsqlTypes/NpgsqlRange.cs) that adds/looks up type converters at runtime.

jskeet:
Great - that's probably enough information for me to get going when I have time (which won't be for a while, unfortunately).

knumat:
> It should be something along the lines of:
> 
> ```cs
> TypeDescriptor.AddAttributes(
>     typeof(LocalDate),
>     new TypeConverterAttribute(typeof(LocalDateTypeConverter)));
> ```

I tested this out with ASP.NET Core 2.1 and NodaTime 2.4.2 and it worked great.  Specifically, I imported the TypeConverters from the iso-type-converters branch and called TypeDescriptor.AddAttributes from my Startup constructor for each type that I needed (LocalDate, LocalTime, and LocalDateTime).  After doing that, my Web API Controller was able to parse both query string parameters and route parameters directly into NodaTime objects.

austindrenski:
I've pushed ad892c7c22705d52479db89e33035f7a35e00dda to add the copyright notices and refactor with a generic base class.

@knumat wrote:
> and called `TypeDescriptor.AddAttributes` from my `Startup` constructor 

That should be fine, since the constructor for `Startup` should only run once per app lifetime. For good measure, you could move those calls into the static constructor.

src/NodaTime/Duration.cs
```diff
@@ -53,11 +54,12 @@ namespace NodaTime
     /// Various operations accept or return a <see cref="Double"/>, in-keeping with durations often being natural lengths
     /// of time which are imprecisely measured anyway. The implementation of these operations should never result in a not-a-number
     /// or infinite value, nor do any operations accept not-a-number or infinite values. Additionally, operations involving
-    /// <c>Double</c> have initially been implemented fairly na�vely; it's possible that future releases will improve the accuracy
+    /// <c>Double</c> have initially been implemented fairly naïvely; it's possible that future releases will improve the accuracy
```
austindrenski: It looks like this file was checked into `nodatime/master` with a windows-1252 diacritic symbol, causing that character to break when checked out as UTF-8. 

Assuming that this wasn't intentional, I've replaced the broken character with the UTF-8 version.

src/NodaTime.Test/Text/AnnualDateTypeConverterTest.cs
```diff
@@ -0,0 +1,20 @@
...
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using JetBrains.Annotations;
+using NodaTime.Text;
+
+namespace NodaTime.Test.Text
+{
+    [UsedImplicitly]
+    public class AnnualDateTypeConverterTest : TypeConverterBaseTestBase<AnnualDate>
```
austindrenski: The test suite appears to use a style of type-under-test-per-file, so I've tried to follow suit with these. That said, I could move them all into a single file (e.g. `TypeConverterTest.cs`) with each derived type as a separate method.

src/NodaTime.Test/Text/AnnualDateTypeConverterTest.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        [NotNull] [UsedImplicitly] static readonly string[] RoundtripData =
+        {
+            AnnualDatePattern.Iso.Format(default),
```
jskeet: For these tests, I wonder whether the literal values would be clearer, and possibly rather than *just* round-tripping, provide the actual value as well.
So a test would be of the form:

```csharp
new AnnualDate(1, 1), "01-01"
```

(Or parameterized as `[TestCase(1, 1, "01-01")]`)

Otherwise if the type converter just formatted everything as "" and used a default value, the tests might pass. (I haven't double checked that.)

src/NodaTime/Text/AnnualDateTypeConverter.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2018 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+namespace NodaTime.Text
```
jskeet: Given the simplicity of all of these converters, I'd be happy to have a single `TypeConverters.cs` file that has them all in. We don't even need XML doc for them, given their triviality.

src/NodaTime/Text/TypeConverterBase.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <inheritdoc />
+        [Pure]
+        public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
+            => sourceType == typeof(string);
+
+        /// <inheritdoc />
+        [Pure]
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
+        {
+            if (value is string instantString)
```
jskeet: Rename to "text"? It may not be an instant at all...

austindrenski:
> Finally had a look at this, and I'm basically happy, but have asked for a bunch of nits and organizational changes. Would you like to do that yourself, or would you prefer me to merge then change? I'm easy either way.

Thanks for the review. I've got some time today, so I can make the changes you've requested. I just need to refactor the tests, and then I'll push a commit.

austindrenski:
@jskeet Update and ready for re-review.

roji:
@jskeet regarding your comment on nullable refs and .NET Core 3:

> Note that it's unlikely that there'll be a release containing this code for a while, as the master branch is for 3.0. I won't be releasing it at least until C# 8 hits GA, as one of the aims is to be nullable reference type compliant.

If the only C# 8 feature used is nullable refs, then it seems at this stage that it's a compiler-only feature, meaning that you can target any version of .NET Framework/Core/Standard... It's simply about adding attributes on methods/parameters, and even those special compiler-recognized attributes are themselves generated by the (C# 8) compiler to avoid taking a dependency on any newer version of the platform.

So unless I'm totally misunderstand things (or things change), you should be fine to go with nullable refs and release before 3.0 or C# 8 RTM.

jskeet:
@roji: It's more than that: it's a matter of what we advertise in terms of nullable reference safety. Given that we *have* taken a major version bump, I want to wait until C# 8 is out so that the division is very clear: 1.x and 2.x won't be annotated to be nullable-reference-safe, 3.x and onwards will be.

That's not going to be the only change though: I'm hoping to use Span-based parsing and formatting too.

roji:
@jskeet sure thing, just wanted to draw your attention in case this point was somehow missed. We're also discussing nullable refs in Npgsql vNext, but it seems a bit early to go into that.

Anyway, looking forward to NodaTime 3.0!

CumpsD:
I've copied the converters, and used this in my Startup:

```csharp
            TypeDescriptor.AddAttributes(
                typeof(Instant),
                new TypeConverterAttribute(typeof(InstantTypeConverter)));
```

My request looks like:

```csharp
   public class ImportUserRequest
    {
        [Required]
        public Instant CreatedAt { get; }

        [Required]
        public Instant LastChangedAt { get; }
```

And I am getting it into :

```csharp
     public async Task<IActionResult> ImportUser(
            [FromBody] ImportUserRequest request)
```

However, my Instants are always empty (1970)

I tried changing my request to:

```csharp
   public class ImportUserRequest
    {
        [Required]
        [TypeConverter(typeof(InstantTypeConverter))]
        public Instant CreatedAt { get; }

        [Required]
        [TypeConverter(typeof(InstantTypeConverter))]
        public Instant LastChangedAt { get; }
```

But no change.

My payload is pretty simple:

```js
{
    createdAt: "2011-07-14T17:49:34Z",
    lastChangedAt: "2014-01-12T12:42:12Z"
}
```

Any ideas what could be wrong?

CumpsD:
Guess what.....

It might help if I added a `set;` to those properties 🤦‍♂

PR #1238, opened 2018-12-17
Disable code coverage temporarily

jskeet:
(This should fix appveyor, but not Travis, which still doesn't have the .NET Core 3.0 preview bits available.)


jskeet:
It turns out it doesn't actually fix AppVeyor, but it gets closer. Merging, and will then address the web app issue.

PR #1240, opened 2018-12-17
Support C# 8 nullable reference types

jskeet:
This PR is now "complete" in terms of applying nullable reference types to all the projects in NodaTime-All.sln.

Further work:

- Remove the existing nullability annotations and tests for them (probably, anyway)
- Review complete API after merging; that will be simpler than looking at the diff
- Update web-based projects too

It's expected that AppVeyor fails right now, as it [doesn't support VS2019 preview 2 just yet](https://github.com/appveyor/ci/issues/2834). I'm happy to merge based on Travis and local builds. (It's not like this code will be going GA any time soon.)

### DIFF
--- a/lib/commandline/CommandLine.cs
+++ b/lib/commandline/CommandLine.cs
@@ -43,20 +43,22 @@ namespace CommandLine
     /// </summary>
     public abstract class BaseOptionAttribute : Attribute
     {
-        private string _shortName;
-        private object _defaultValue;
+        private string? _shortName;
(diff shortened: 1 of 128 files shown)


jskeet:
Nullable reference types currently confuse docfx in terms of generated metadata. I'll see if I can fix this up with tooling. I'll do that as a separate PR, but it'll basically block this for now.

jskeet:
Workaround tooling is in #1272, so we can now merge this (then that).

PR #1264, opened 2019-01-29
Use switch expressions from C# 8

jskeet:


src/NodaTime.Test/Calendars/BclCalendars.cs
```diff
@@ -62,23 +62,18 @@ private static Calendar GetCalendar(string name)
...
-                case "PersianCalendar": return bcl.IsLeapYear(1) ? CalendarSystem.PersianSimple : CalendarSystem.PersianAstronomical;
-                case "UmAlQuraCalendar": return CalendarSystem.UmAlQura;
-                case "JulianCalendar":
-                    return CalendarSystem.Julian;
-                default:
-                    // No idea - we can't test with this calendar...
-                    return null;
-            }
-        }
+                "GregorianCalendar" => CalendarSystem.Iso,
```
mungojam: Could you switch based on the actual class type now, instead of having to use `.GetType().Name()`, i.e.:

```cs
bcl switch
{
    GregorianCalendar _  => CalendarSystem.Iso,
    HijriCalendar _ => CalendarSystem.IslamicBcl,
//etc.
}
```

Not sure if the `_` is actually needed or can be left out altogether

jskeet: We couldn't do so before, not due to switch expressions but because we couldn't refer to the specific calendar system types in more limited environments.
Now we only support netstandard2.0, this might be feasible - thanks for highlighting it. Will have a look...

PR #1282, opened 2019-02-14
Fixes from API review:

### LINKED ISSUE #1275: Review nullability of all reference type occurrences in the API
(We may want tooling to make this simpler, e.g. listing all methods/properties which mention any reference types anywhere.)

jskeet:
- Equals(object) should accept null
- (Implementation detail) use pattern matching in Equals
- Add documentation for ParseResult.ConvertTo type parameter
- Improve formatting of documentation for ZoneLocalMapping (docfx bug?)

Fixes #1275

### DIFF
--- a/src/NodaTime/LocalInstant.cs
+++ b/src/NodaTime/LocalInstant.cs
@@ -187,7 +187,7 @@ namespace NodaTime
         /// <c>true</c> if the specified <see cref="System.Object"/> is equal to this instance;
         /// otherwise, <c>false</c>.
         /// </returns>
-        public override bool Equals(object obj) => obj is LocalInstant && Equals((LocalInstant)obj);
+        public override bool Equals(object? obj) => obj is LocalInstant other && Equals(other);
 
         /// <summary>
         /// Returns a hash code for this instance.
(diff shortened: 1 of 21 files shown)


src/NodaTime/LocalInstant.cs
```diff
@@ -187,7 +187,7 @@ internal Instant SafeMinus(Offset offset)
         /// <c>true</c> if the specified <see cref="System.Object"/> is equal to this instance;
         /// otherwise, <c>false</c>.
         /// </returns>
-        public override bool Equals(object obj) => obj is LocalInstant && Equals((LocalInstant)obj);
+        public override bool Equals(object? obj) => obj is LocalInstant other && Equals(other);
```
malcolmr: Not that it matters too much, but we seem to be writing `Equals(object?)` three different ways:

`=> obj is LocalInstant other && Equals(other);`
`=> obj is LocalInstant other && this == other;`
`=> Equals(obj as LocalInstant)`

Is there any reason to prefer one over the others? If not, can we at least make them consistent?

jskeet: The first is the most common one for value types.

The second is a very slight (perhaps negligible) optimization if I happen to spot that `Equals` calls `==` directly.

The third is the simplest way of doing the first but for reference types.

So I don't mind 1 and 3 existing. Perhaps we should standardize between 1 and 2 though.

PR #1283, opened 2019-02-15
Move to NUnit3TestAdapter, so dotnet test works

### LINKED ISSUE #967: Move to dotnet test for NUnit
The NUnit3TestAdapter nuget package is the relevant one here.

Build scripts etc need changing, of course.

jskeet:
Fixes #967.

We probably don't want to actually merge this until 3.13.0 is out, as that will fix the slow test running with filters.


jskeet:
Actually I might as well merge now, given that it's still only a few minutes for the CI build. We can just improve it when the newer NUnit3TestAdapter comes out.

PR #1288, opened 2019-02-18
Allow local benchmarks to be loaded for development

jskeet:


jskeet:
Current benchmarks: https://drive.google.com/open?id=1cob-Bko1lAPsqsOB9-H3KEl8EoGXqmjA
Download and unzip into `src/NodaTime.Web` (the zip file already contains a `benchmarks` directory)

PR #1291, opened 2019-02-23
Add a new roundtrip pattern for durations

### LINKED ISSUE #1290: Include standard Duration pattern that matches the one used for JSON
The standard Duration pattern described in the docs:
https://nodatime.org/2.3.x/userguide/duration-patterns
is `-D:hh:mm:ss.FFFFFFFFF`
whereas the `JsonConverter` in 
https://github.com/nodatime/nodatime.serialization/blob/edf88349e54ab63f06266a9c864edfa1ba6503cb/src/NodaTime.Serialization.JsonNet/NodaConverters.cs#L102

has a pattern `-H:mm:ss.FFFFFFFFF`.
Does this serve any purpose, or it can simply be `DurationPattern.Roundtrip` there?

jskeet:
This matches the pattern that NodaTime.Serialization.JsonNet uses,
which just has a number of hours (and smaller) rather than the days
and hours of the regular roundtrip pattern.

Fixes #1290.

**Reviewers**

Thoughts on the choice of "j" as the standard pattern are welcome, along with the documentation. There's nothing JSON-specific about this format, but it happens to be the reason we're creating it.

### DIFF
--- a/src/NodaTime/Text/DurationPattern.cs
+++ b/src/NodaTime/Text/DurationPattern.cs
@@ -25,10 +25,18 @@ namespace NodaTime.Text
     {
         /// <summary>
         /// Gets the general pattern for durations using the invariant culture, with a format string of "-D:hh:mm:ss.FFFFFFFFF".
-        /// This pattern round-trips.
+        /// This pattern round-trips. This corresponds to the "o" standard pattern.
         /// </summary>
         /// <value>The general pattern for durations using the invariant culture.</value>
         public static DurationPattern Roundtrip => Patterns.RoundtripPatternImpl;
+
+        /// <summary>
+        /// Gets a pattern for durations using the invariant culture, with a format string of "-H:mm:ss.FFFFFFFFF".
+        /// This pattern round-trips, and is the pattern that NodaTime.Serialization.JsonNet uses by default.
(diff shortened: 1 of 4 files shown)


src/NodaTime/Text/DurationPattern.cs
```diff
@@ -25,17 +25,26 @@ public sealed class DurationPattern : IPattern<Duration>
...
         /// </summary>
         /// <value>The general pattern for durations using the invariant culture.</value>
         public static DurationPattern Roundtrip => Patterns.RoundtripPatternImpl;
 
+        /// <summary>
+        /// Gets a pattern for durations using the invariant culture, with a format string of "-H:mm:ss.FFFFFFFFF".
+        /// This pattern round-trips, and is the pattern that NodaTime.Serialization.JsonNet uses by default.
+        /// This corresponds to the "j" standard pattern.
+        /// </summary>
+        /// <value>The pattern for durations using the invariant culture. This is the pattern that NodaTime.Serialization.JsonNet used by default.</value>
```
malcolmr: "uses by default"?

jskeet: Humbug - I fixed the first one of those and force-pushed, but missed the second :) Will fix and merge. Thanks!

PR #1299, opened 2019-03-02
Fixes #937

### LINKED ISSUE #937: Improve benchmark navigation and display
The benchmark display we currently have is reasonably functional, but far from pretty.

Currently, the data model is all in-memory, which should make it fairly easy to do all kinds of cool things without worrying about accessing too much data. If/when we move to something like Spanner, we'll need to revisit all the visualizations, but that can wait...

I'm very happy to go over the data model in more detail with interested parties. The code is just ASP.NET Core, in a pretty simple way. Ideas for other styles of navigation and visualization welcome - pull requests even more welcome!

shaddu:
Changes done under this : 

- Added Breadcrumb on benchmark related screen for proper navigations 
- Moved the view hyperlink on right side of table
- Fixed a hieght issue , where footer was coming in between screen and not sticking in bottom

### DIFF
--- a/src/NodaTime.Web/Views/Benchmarks/CompareTypesByCommit.cshtml
+++ b/src/NodaTime.Web/Views/Benchmarks/CompareTypesByCommit.cshtml
@@ -6,6 +6,13 @@
 }
 
 <div class="row">
+	<ul class="breadcrumbs">
+		<li><a href="/">Home</a></li>
+		<li><a href="/benchmarks">Benchmarks</a></li>
+		<li><a href="/benchmarks/environments/@Model.Environment.BenchmarkEnvironmentId">Environment (@Model.Environment.Machine, @Model.Environment.TargetFramework )</a></li>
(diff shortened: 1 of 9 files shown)


jskeet:
Thanks, will pull and play with this tomorrow.

jskeet:
Just looking at this now. I like the breadcrumbs - but I don't like the way the details end up on the right hand side of the screen. I preferred the layout of "Summary/context at the top, followed by the details below."

How easy would it be to keep the rest of the PR but change that part of the layout? (I haven't looked at the change in terms of implementation at all yet.)

shaddu:
Those changes were done on two pages only. I Thought that it utilize space better than to scroll down for detail. I will revert the changes for the two pages and push changes

shaddu:
it is happening for last namespace type only

jskeet:
Right. I'm happy for you to look into that, or decide not to as you wish :) (It's definitely not high priority!)

shaddu:
@jskeet I can check it, but I am more interested in issues related to nodatime functionality.

jskeet:
No problem. I'll look if I get an idle time :)

PR #1322, opened 2019-03-19
Add Stackdriver integration

jskeet:
This isn't as clean as I'd like it to be, but it looks like it works.


jskeet:
Not merging yet - I need to check that debug logs really are filtered as I expect them to be.

jskeet:
Okay, validated the logging - and fixed the commit. Will merge on green.

PR #1335, opened 2019-04-06
Fix for phrasing: "notional global timeline"

SeanKilleen:
Was reading the docs and saw the phrase (emphasis mine)

> An Instant is a point on **notional a** global time-line

And I believe you may have meant:

> An Instant is a point on **a notional** global time-line

So I figured I'd do a quick PR to suggest it.


jskeet:
Well spotted, thanks :)

PR #1337, opened 2019-04-07
Add TimeInterval struct

rodion-m:
In our case TimeInterval seemed to be very useful (flexible working schedule for employees). I guess many of developers will find it useful too.
Logic for this struct is based on `Interval` source code.

### DIFF
--- a/src/NodaTime.Test/TimeIntervalTest.cs
+++ b/src/NodaTime.Test/TimeIntervalTest.cs
@@ -17,34 +17,34 @@ namespace NodaTime.Test
         private static readonly LocalTime SampleEndInverted = new LocalTime(1, 00);
 
         [Test]
-        public void Duration_Success()
+        public void Duration_True()
         {
             var interval = new TimeInterval(SampleStart, SampleEnd);
             Assert.AreEqual(Duration.FromHours(7), interval.Duration);
         }
         [Test]
-        public void Period_Success()
+        public void Period_True()
         {
             var interval = new TimeInterval(SampleStart, SampleEnd);
             Assert.AreEqual(Period.FromHours(7), interval.Period);
         }
         
         [Test]
-        public void InvertedTimeInterval_Duration_Success()
+        public void InvertedTimeInterval_Duration_True()
         {
             var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
             Assert.AreEqual(Duration.FromHours(2), interval.Duration);
         }
         
         [Test]
-        public void InvertedTimeInterval_Period_Success()
+        public void InvertedTimeInterval_Period_True()
         {
             var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
             Assert.AreEqual(Period.FromHours(2), interval.Period);
         }
         
         [Test]
-        public void InvertedTimeInterval_Contains_Success()
+        public void InvertedTimeInterval_Contains_True()
         {
             var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
             Assert.IsTrue(interval.Contains(new LocalTime(23, 00)));
@@ -55,7 +55,7 @@ namespace NodaTime.Test
         }
         
         [Test]
-        public void InvertedTimeInterval_NotContains_Success()
+        public void InvertedTimeInterval_Contains_False()
         {
             var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
             Assert.IsFalse(interval.Contains(new LocalTime(1, 00)));
@@ -63,6 +63,43 @@ namespace NodaTime.Test
             Assert.IsFalse(interval.Contains(new LocalTime(13, 00)));
             Assert.IsFalse(interval.Contains(new LocalTime(22, 30)));
             Assert.IsFalse(interval.Contains(new LocalTime(22, 59)));
+        }
+
+
+        [Test]
+        public void ToString_True()
+        {
+            var sampleStart = SampleStart.PlusMilliseconds(230).PlusNanoseconds(139);
+            var interval = new TimeInterval(sampleStart, SampleEnd);
+            Assert.AreEqual("10:00:00.230000139/17:00:00", interval.ToString());
+        }
+        [Test]
+        public void Parse_True()
+        {
+            var sampleStart = SampleStart.PlusMilliseconds(230).PlusNanoseconds(139);
+            var interval = new TimeInterval(sampleStart, SampleEnd);
+            Assert.AreEqual(interval, TimeInterval.Parse("10:00:00.230000139/17:00:00"));
+        }
+        
+        [Test]
+        public void Parse_False()
+        {
+            var interval = new TimeInterval(SampleStart, SampleEnd);
+            Assert.AreNotEqual(interval, TimeInterval.Parse("10:00:00.230000139/17:00:00"));
+        }
+        
+        [Test]
+        public void Parse_ThrowsException()
+        {
+            Assert.Throws<FormatException>(() => TimeInterval.Parse("10:00:0017:00:00"));
+        }
+        
+        [Test]
+        public void ToStringAndParse_True()
+        {
+            var sampleStart = SampleStart.PlusMilliseconds(230).PlusNanoseconds(139);
+            var interval = new TimeInterval(sampleStart, SampleEnd);
+            Assert.AreEqual(interval, TimeInterval.Parse(interval.ToString()));
         }
 
     }
--- a/src/NodaTime.Test/TimeIntervalTest.cs
+++ b/src/NodaTime.Test/TimeIntervalTest.cs
@@ -1,0 +1,69 @@
+// Copyright 2011 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using System;
+using NodaTime.Text;
+using NUnit.Framework;
+
+namespace NodaTime.Test
+{
+    public class TimeIntervalTest
+    {
+        private static readonly LocalTime SampleStart = new LocalTime(10, 00);
+        private static readonly LocalTime SampleEnd = new LocalTime(17, 00);
+        
+        private static readonly LocalTime SampleStartInverted = new LocalTime(23, 00);
+        private static readonly LocalTime SampleEndInverted = new LocalTime(1, 00);
+
+        [Test]
+        public void Duration_Success()
+        {
+            var interval = new TimeInterval(SampleStart, SampleEnd);
+            Assert.AreEqual(Duration.FromHours(7), interval.Duration);
+        }
+        [Test]
+        public void Period_Success()
+        {
+            var interval = new TimeInterval(SampleStart, SampleEnd);
+            Assert.AreEqual(Period.FromHours(7), interval.Period);
+        }
+        
+        [Test]
+        public void InvertedTimeInterval_Duration_Success()
+        {
+            var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
+            Assert.AreEqual(Duration.FromHours(2), interval.Duration);
+        }
+        
+        [Test]
+        public void InvertedTimeInterval_Period_Success()
+        {
+            var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
+            Assert.AreEqual(Period.FromHours(2), interval.Period);
+        }
+        
+        [Test]
+        public void InvertedTimeInterval_Contains_Success()
+        {
+            var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
+            Assert.IsTrue(interval.Contains(new LocalTime(23, 00)));
+            Assert.IsTrue(interval.Contains(new LocalTime(23, 30)));
+            Assert.IsTrue(interval.Contains(new LocalTime(00, 00)));
+            Assert.IsTrue(interval.Contains(new LocalTime(00, 30)));
+            Assert.IsTrue(interval.Contains(new LocalTime(00, 59)));
+        }
+        
+        [Test]
+        public void InvertedTimeInterval_NotContains_Success()
+        {
+            var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
+            Assert.IsFalse(interval.Contains(new LocalTime(1, 00)));
+            Assert.IsFalse(interval.Contains(new LocalTime(1, 01)));
+            Assert.IsFalse(interval.Contains(new LocalTime(13, 00)));
+            Assert.IsFalse(interval.Contains(new LocalTime(22, 30)));
+            Assert.IsFalse(interval.Contains(new LocalTime(22, 59)));
+        }
+
+    }
+}
--- a/src/NodaTime/TimeInterval.cs
+++ b/src/NodaTime/TimeInterval.cs
@@ -9,276 +9,276 @@ using NodaTime.Utility;
 
 namespace NodaTime
 {
-	/// <summary>
-	/// An interval between two local time values (start and end).
-	/// </summary>
-	/// <remarks>
-	/// <para>
-	/// The TimeInterval includes the start LocalTime and excludes the end LocalTime. However, an TimeInterval
-	/// may be missing its start or end, in which case the TimeInterval is deemed to be infinite in that
-	/// direction.
-	/// </para>
-	/// <para>
-	/// The end may equal the start (resulting in an empty TimeInterval),
-	/// but it can also be before the start: in this case property IsEndInNextDay will be true
-	/// </para>
-	/// </remarks>
-	/// <threadsafety>This type is an immutable value type. See the thread safety section of the user guide for more information.</threadsafety>
-	public readonly struct TimeInterval : IEquatable<TimeInterval>
-	{
-		/// <summary>The start of the interval.</summary>
-		private readonly LocalTime? start;
+    /// <summary>
+    /// An interval between two local time values (start and end).
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The TimeInterval includes the start LocalTime and excludes the end LocalTime. However, an TimeInterval
+    /// may be missing its start or end, in which case the TimeInterval is deemed to be infinite in that
+    /// direction.
+    /// </para>
+    /// <para>
+    /// The end may equal the start (resulting in an empty TimeInterval),
+    /// but it can also be before the start: in this case property IsEndInNextDay will be true
+    /// </para>
+    /// </remarks>
+    /// <threadsafety>This type is an immutable value type. See the thread safety section of the user guide for more information.</threadsafety>
+    public readonly struct TimeInterval : IEquatable<TimeInterval>
+    {
+        /// <summary>The start of the interval.</summary>
+        private readonly LocalTime? start;
 
-		/// <summary>The end of the interval. It can be less than start.</summary>
-		private readonly LocalTime? end;
-		
-		/// <summary>
-		/// Initializes a new instance of the <see cref="TimeInterval"/> struct.
-		/// The TimeInterval includes the <paramref name="start"/> LocalTime and excludes the
-		/// <paramref name="end"/> LocalTime. The end may equal the start (resulting in an empty TimeInterval).
-		/// </summary>
-		/// <param name="start">The start <see cref="LocalTime"/>.</param>
-		/// <param name="end">The end <see cref="LocalTime"/>.</param>
-		public TimeInterval(LocalTime start, LocalTime end)
-		{
-			this.start = start;
-			this.end = end;
-			IsEndInNextDay = end.NanosecondOfDay - start.NanosecondOfDay < 0;
-		}
+        /// <summary>The end of the interval. It can be less than start.</summary>
+        private readonly LocalTime? end;
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeInterval"/> struct.
+        /// The TimeInterval includes the <paramref name="start"/> LocalTime and excludes the
+        /// <paramref name="end"/> LocalTime. The end may equal the start (resulting in an empty TimeInterval).
+        /// </summary>
+        /// <param name="start">The start <see cref="LocalTime"/>.</param>
+        /// <param name="end">The end <see cref="LocalTime"/>.</param>
+        public TimeInterval(LocalTime start, LocalTime end)
+        {
+            this.start = start;
+            this.end = end;
+            IsEndInNextDay = end.NanosecondOfDay - start.NanosecondOfDay < 0;
+        }
 
-		/// <summary>
-		/// Initializes a new instance of the <see cref="TimeInterval"/> struct from two nullable <see cref="LocalTime"/>
-		/// values.
-		/// </summary>
-		/// <remarks>
-		/// If the start is null, the TimeInterval is deemed to stretch to the start of time. If the end is null,
-		/// the TimeInterval is deemed to stretch to the end of time.
-		/// </remarks>
-		/// <param name="start">The start <see cref="LocalTime"/> or null.</param>
-		/// <param name="end">The end <see cref="LocalTime"/> or null.</param>
-		public TimeInterval(LocalTime? start, LocalTime? end)
-		{
-			this.start = start;
-			this.end = end;
-			if (start.HasValue && end.HasValue)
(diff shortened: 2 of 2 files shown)


src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+// as found in the LICENSE.txt file.
+
+using JetBrains.Annotations;
+using NodaTime.Text;
+using System;
+using NodaTime.Utility;
+
+namespace NodaTime
+{
+	/// <summary>
```
jskeet: Quick drive-by comment before I do the real review: please format the code with spaces instead of tabs, at a four space indent.

jskeet (commented):
One comment now, but I'll try to review the rest tomorrow. (I'm afraid I'm busy all of today.)

src/NodaTime/TimeInterval.cs
Rody66: > 
> 
> Quick drive-by comment before I do the real review: please format the code with spaces instead of tabs, at a four space indent.

It fixed.

src/NodaTime.Test/TimeIntervalTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        
+        [Test]
+        public void InvertedTimeInterval_Period_True()
+        {
+            var interval = new TimeInterval(SampleStartInverted, SampleEndInverted);
+            Assert.AreEqual(Period.FromHours(2), interval.Period);
+        }
+        
+        [Test]
+        public void InvertedTimeInterval_Contains_True()
```
jskeet: Please could you convert all the tests that are data-driven to use TestCase? For example, I'd expect:

```csharp
[Test]
[TestCase(23, 0, true)]
... // Other true cases
[TestCase(1, 0, false)]
... // Other false cases
public void Contains_InvertedTimeInterval(int hour, int minute, bool expectedResult)
```

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+using NodaTime.Utility;
+
+namespace NodaTime
+{
+    /// <summary>
+    /// An interval between two local time values (start and end).
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The TimeInterval includes the start LocalTime and excludes the end LocalTime. However, an TimeInterval
```
jskeet: The part about a missing start time feels slightly odd to me, because it's exactly equivalent to having a start time of 00:00:00. (This isn't true for Interval.) So we can make sure there's *always* a start time... we can't do the same for the end time, as the end is exclusive and we don't support 24:00 as a `LocalTime` value.

With start being nullable, we can have two time intervals (one starting at 00:00:00 and one having no start) which are functionally equivalent, but not equal to each other. I'd quite like some more input about this.

@malcolmr @lundmikkel @mj1856 any thoughts?

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// <summary>The start of the interval.</summary>
+        private readonly LocalTime? start;
+
+        /// <summary>The end of the interval. It can be less than start.</summary>
+        private readonly LocalTime? end;
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeInterval"/> struct.
+        /// The TimeInterval includes the <paramref name="start"/> LocalTime and excludes the
+        /// <paramref name="end"/> LocalTime. The end may equal the start (resulting in an empty TimeInterval).
```
jskeet: We should indicate the behaviour if start is later than end - in particular, that it's valid, but represents an inverted interval. That could be surprising in some cases. (I think it's *probably* the right behavior, but worth calling out.)

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+            this.end = end;
+            IsEndInNextDay = start > end;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeInterval"/> struct from two nullable <see cref="LocalTime"/>
+        /// values.
+        /// </summary>
+        /// <remarks>
+        /// If the start is null, the TimeInterval is deemed to stretch to the start of time. If the end is null,
```
jskeet: I'd use "start of the day" and "end of the day" rather than "start of time" and "end of time".

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// the TimeInterval is deemed to stretch to the end of time.
+        /// </remarks>
+        /// <param name="start">The start <see cref="LocalTime"/> or null.</param>
+        /// <param name="end">The end <see cref="LocalTime"/> or null.</param>
+        public TimeInterval(LocalTime? start, LocalTime? end)
+        {
+            this.start = start;
+            this.end = end;
+            if (start.HasValue && end.HasValue)
```
jskeet: Please either use braces, or just

```csharp
IsEndInNextDay = start.HasValue && end.HasValue && start > end;
```

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// Gets the start LocalTime - the inclusive lower bound of the TimeInterval.
+        /// </summary>
+        /// <value>The start <see cref="LocalTime"/>.</value>
+        /// <exception cref="InvalidOperationException">The TimeInterval extends to the start of time.</exception>
+        /// <seealso cref="HasStart"/>
+        public LocalTime Start
+        {
+            get
+            {
+                Preconditions.CheckState(start.HasValue, "TimeInterval extends to start of time");
```
jskeet: One advantage of not having a nullable start is that we don't need to worry about this.

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// extends to the start of time.</value>
+        public bool HasStart => start != null;
+
+        /// <summary>
+        /// Gets the end LocalTime - the exclusive upper bound of the TimeInterval.
+        /// </summary>
+        /// <value>The end <see cref="LocalTime"/>.</value>
+        /// <exception cref="InvalidOperationException">The TimeInterval extends to the end of time.</exception>
+        /// <seealso cref="HasEnd"/>
+        public LocalTime End
```
jskeet: It's at least worth *considering* just making this nullable. I couldn't do that in Interval because we only made use of nullable instants after 1.0 had shipped.
I'm not sure of the right thing to do in this particular case.

src/NodaTime/TimeInterval.cs
jskeet: (Or remove entirely - see later.)

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// </summary>
+        /// <value><c>true</c> if this TimeInterval has a fixed end point, or <c>false</c> if it
+        /// extends to the end of time.</value>
+        public bool HasEnd => end.HasValue;
+
+
+        /// <summary>
+        /// Returns <c>true</c> if Start of TimeInterval is greater than End (for intervals like 23:00/1:00)
+        /// </summary>
+        public bool IsEndInNextDay { get; }
```
jskeet: I'd like us to brainstorm some alternative names here, e.g.

- WrapsMidnight
- IsInverted
- ???

Additionally, I'm not sure it's worth keeping this as a field. It's cheap to compute, and entirely determined by start/end. We could evaluate it on each access.

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+
+        /// <summary>
```
jskeet: I'm not sure this worth having as a property. Periods are *usually* around calendrical calculations.
What we should probably have is just the Duration property, and then Duration.ToPeriod (which I'm slightly surprised we don't already have). Then at least the functionality is simple to achieve.

We can always add this property later if necessary.

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        /// <summary>
+        /// Returns the Duration of the TimeInterval.
+        /// </summary>
+        /// <remarks>
+        /// This will always be a non-negative Duration, though it may be zero.
+        /// </remarks>
+        /// <value>The Duration of the TimeInterval.</value>
+        /// <exception cref="InvalidOperationException">The TimeInterval extends to the start or end of time.</exception>
```
jskeet: I'm not sure about this. It feels like a TimeInterval with no start and no end should be 24 hours long, and a single-end interval should have a duration based on the start/end of the day, e.g. 10pm- has a duration of 2 hours, and -8am has a duration of 8 hours.

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+
+        /// <summary>
+        /// Returns whether or not this TimeInterval contains the given LocalTime.
+        /// </summary>
+        /// <param name="check">LocalTime to test.</param>
+        /// <returns>True if this TimeInterval contains the given LocalTime; false otherwise.</returns>
+        [Pure]
+        public bool Contains(LocalTime check)
+        {
+            if (IsEndInNextDay)
```
jskeet: I suspect we can simplify the logic here, but we can do that later.

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+            return check >= Start && check < End;
+        }
+
+        /// <summary>
+        /// Deconstruct this value into its components.
+        /// </summary>
+        /// <param name="_start">The start of the TimeInterval.</param>
+        /// <param name="_end">The end of the TimeInterval.</param>
+        [Pure]
+        public void Deconstruct(out LocalTime? _start, out LocalTime? _end)
```
jskeet: Please rename these parameters to `start` and `end`.

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        {
+            _start = start;
+            _end = end;
+        }
+
+        #region Implementation of IEquatable<TimeInterval>
```
jskeet: Let's get rid of all these regions before they even appear. They're *somewhat* helpful for more complex types, but I don't think there's value here. There are plenty I should probably get rid of in the existing code but haven't got round to...

src/NodaTime/TimeInterval.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// <summary>
+        /// Parses the given text value according to the rule LocalTimePattern.ExtendedIso,
+        /// using / char as separator or two LocalTimes
+        /// </summary>
+        /// <param name="text">The text value to parse.</param>
+        /// <returns>The result of parsing.</returns>
+        /// <example>3:44:59/4:00:00</example>
+        /// <example>18:00:00.239/23:00:00</example>
+        /// <exception cref="FormatException"></exception>
+        public static TimeInterval Parse(string text)
```
jskeet: I think I'd prefer to at least make this internal for the moment. We may or may not want a TimeIntervalPattern in the future - it feels odd to have a Parse method on this type but nowhere else. It's easy to make internal methods public later, but we can't go the other way.

Rody66: @jskeet Thank you for the detailed review! I'll fix everything that in my power.
About Parse - I understand that it's not idiomatic way in NodaTime, but we don't have any solution now to parse TimeInterval (I use this method for deserialization in my project). Perhaps we can leave it public but decorate with attribute `[Obsolete("This method is temporary and will be replaced with TimeIntervalPattern.")]`?

jskeet: No, I don't want to have anything "obsolete from the start" in the master - it's too easy to accidentally leave it in there.

The Parse method doesn't rely on anything internal, so you can just have that in your project's code for now, but still use TimeInterval itself in NodaTime.

Rody66: Ok, thank you

src/NodaTime/TimeInterval.cs
Rody66: Actually I cannot imagine any case when the TimeInterval may not have a start or end. I want to suggest this logic: `if(start == null) start = LocalTime.MinValue`, `if(end == null) end = LocalTime.MaxValue`. Or maybe we should add `LocalTime.AfterMaxValue` to include the last nanosecond (because comparison at the moment is exclusively)? Or just make `end` comparison inclusively?
Please consider it.

jskeet: Not having a start is equivalent to midnight. That's easy. But it's entirely reasonable to create an interval of "Any time from 7pm onwards" but we can't represent that without something like LocalTime.AfterMaxValue, and I don't want to create that *and expose it*. (We've got that for Instant because it makes many things a lot simpler, but we avoid ever exposing it.)

src/NodaTime/TimeInterval.cs
Rody66: Do you mean something like this?
```csharp
        public TimeInterval(LocalTime start, LocalTime end, bool endCanBeInNextDay = false)
        {
            this.start = start;
            this.end = end;
            IsEndInNextDay = start > end;
            if(IsEndInNextDay && !endCanBeInNextDay) 
                throw new ArgumentOutOfRangeException(nameof(end), 
                    "The end parameter must be equal to or later than the start parameter. Enable endCanBeInNextDay if you want to allow end in next day.");
        }
```

jskeet: No, I mean adding more information about the behaviour in the comments.

src/NodaTime/TimeInterval.cs
Rody66: So, just delete region IEquatable, its interface inheritance and move Equals logic into object overridden method?

src/NodaTime/TimeInterval.cs
Rody66: Ok, should I delete the code or just make Period method internal?

src/NodaTime/TimeInterval.cs
Rody66: WrapsMidnight is good, but EndsInNextDay seems to be more concrete for me.

src/NodaTime/TimeInterval.cs
Rody66: Then I suggest to make `start` not nullable, but leave `end` nullable.
So, Contains became:
```csharp
            if (EndsInNextDay)
            {
                if (check < Start) return check < End;
                return true;
            }
            return check >= Start && (!end.HasValue || check < End);
```
And Duration became:
```csharp
                var endNanoseconds = end.HasValue ? End.NanosecondOfDay : NodaConstants.NanosecondsPerDay;
                var nanoseconds = endNanoseconds - Start.NanosecondOfDay;
                if (EndsInNextDay) nanoseconds += NodaConstants.NanosecondsPerDay;
                return Duration.FromNanoseconds(nanoseconds);
```

src/NodaTime/TimeInterval.cs
jskeet: I'm just suggesting removing all of the `#region` and `#endregion` lines - no need for anything to be moved or removed.

src/NodaTime/TimeInterval.cs
jskeet: I'd just remove it entirely.

src/NodaTime/TimeInterval.cs
jskeet: Something like that, yes. But I think it's worth focusing on the API design before worrying too much about the implementation changes.

Rody66: Only one problem with this approach - TimeInterval of 22:00/00:00 and 22:00/null will not be equal, but internally they are. So, if we will use LocalTime.MinValue instead of null for end - it'll not work properly too, because of 00:00/00:00 TimeInterval will be interval with 0 Duration, not 24 hour as we except using null as end. What do you think about it? Should we modify Equals method too to make equal intervals like 22:00/00:00, 22:00/null?

src/NodaTime/TimeInterval.cs
Rody66: I'll try to write it, but don't be surprised if it'll looks slightly ridiculous :)

src/NodaTime.Test/TimeIntervalTest.cs
Rody66: Done. Please tell if I should change some other tests.

src/NodaTime/TimeInterval.cs
Rody66: Made as property.

src/NodaTime/TimeInterval.cs
Rody66: Leaved as is. We should think more about it.

src/NodaTime/TimeInterval.cs
Rody66: Removed.

src/NodaTime/TimeInterval.cs
Rody66: Renamed.

src/NodaTime/TimeInterval.cs
Rody66: Regions removed.

src/NodaTime/TimeInterval.cs
Rody66: Removed.

src/NodaTime/TimeInterval.cs
jskeet: I hadn't thought that of course an end of midnight is "everything before midnight".

If it weren't for the fact that we couldn't represent "all day" as well as "no time at all" it would be perfect just to have Start and End as LocalTime.

Humbug. I think we really need to think carefully about how important it is to represent all the appropriate situations - and being able to construct them directly. (If we didn't mind "All day" or "Empty" being impossible to construct directly, I'm sure we could work out some cunning plan.)

(This is were designing an API for the library is much more time-consuming for designing a type that's useful within a single application, of course.)

I'll see whether java.time has anything like this, and what their approach was, if so.

Rody66: I don't see any LocalTime's intervals in java. Anyways, we can create 00:00/LocalTime.MaxValue and in fact it'll "all day". So, I suggest to make `end` non-nullable and just use MinValue for null-end. In comments we can note that null/null interval will became 00:00/00:00 interval.
I hope another developers will come up some additional use-cases to help us take a decision what logic to choose. 
Should I commit code changes now or lets solve this problem first?

jskeet: No, 00:00-LocalTime.MaxValue wouldn't mean "all day" - it would mean "all day except for the last nanosecond". Fundamentally we need to decide whether X-X (for any value of X) means "all day" or "empty". If we definitely need to support both of those use cases, we won't be able to *just* store two `LocalTime` values.

I'm genuinely not sure which of those use cases is more useful, or whether both are required.

If we're moving away from nullable ends though (which I'd generally like) then `LocalTime?` shouldn't appear anywhere in the API - we wouldn't need to note anything about a "null/null interval" because there'd be no such concept.

Rody66: So you suggest just remove all nullable types from TimeInterval? Including a nullable constructor? `public TimeInterval(LocalTime? start, LocalTime? end)`

jskeet: Yes, with the downside that we then can't represent both "empty" and "all day" - we need to decide which of those is the meaning when start and end are the same.

lundmikkel:
I would love to participate in this discussion, since I've done a lot of work with local time intervals professionally. I won't have time until the weekend though.

One thing I've noticed missing, are methods like `Overlaps()` and `Contains()` which aren't trivial to implement. Having them, will make it a lot easier for a lot of people.

jskeet:
> I won't have time until the weekend though.

I doubt that there's any rush - I'd definitely rather get it right than get it in quickly but badly :) Your feedback will be very valuable whenever you have time to provide it.

Rody66:
> 
> 
> I would love to participate in this discussion, since I've done a lot of work with local time intervals professionally. I won't have time until the weekend though.
> 
> One thing I've noticed missing, are methods like `Overlaps()` and `Contains()` which aren't trivial to implement. Having them, will make it a lot easier for a lot of people.
Hello @lundmikkel ! Thank you for participation in this matter. Do I get right that `Overlaps()` should check if the main interval overlapped the checked onr? Like this:
 ```csharp
var mainInterval = new TimeInterval(new LocalTime(13, 00), new LocalTime(19, 00));
var checkInterval = new TimeInterval(new LocalTime(15, 00), new LocalTime(17, 00));
mainInterval.Overlaps(checkInterval).Shoud().BeTrue(); // correct?
```
And we already have `Contains` method, is it incorrect? 
Please answer when you'll have time.

**Comment by lundmikkel** [comment]
Hi guys!

Sorry for the long delay. I've been collaborating with [@Planday](https://github.com/planday) (my old employer) to open source the NodaTime-specific code I've written for them. That has finally resolved in them releasing the general code [on GitHub](https://github.com/planday/nodatime-extensions).

Though the code is far from being ready to merge into the NodaTime codebase, there are some things of interest to this discussion. Have a look at the [`TimeInterval`](https://github.com/planday/nodatime-extensions/blob/master/NodaTime.Future/TimeInterval.cs) and the [`DateTimeInterval`](https://github.com/planday/nodatime-extensions/blob/master/NodaTime.Future/DateTimeInterval.cs).

A few notes:
* Notice [how converting an `NodaTime.Interval` to a `DateTimeInterval`](https://github.com/planday/nodatime-extensions/blob/master/NodaTime.Future/Extensions/IntervalExtensions.cs#L16-L36) can cause problems if the instants fall inside the DST transition periods.
* The intervals are used by Planday to model events that last at most a day (~24 hours) and are never empty. I.e. a `TimeInterval` cannot be empty, e.g. [14:00 - 14:00) wraps around, crossing midnight.
* There are a buttload of tests for both [`TimeInterval`](https://github.com/planday/nodatime-extensions/blob/master/NodaTime.Tests/TimeIntervalTests.cs) and [`DateTimeInterval`](https://github.com/planday/nodatime-extensions/blob/master/NodaTime.Tests/DateTimeIntervalTests.cs) in the test project.

I'll be going on vacation now (for three weeks), but please let me know, once you have looked through the code.

Finally a big thanks to [Planday](https://www.planday.com/) for releasing the code. This will be a great contribution to open source software!

**Comment by NeoXtreem** [comment]
@jskeet I was just looking for a way to represent a time period where the start and end could be different dates and times, and the `DateTimeInterval` I think would be an awesome addition. Here's my use case: I want to provide a time period between two date-times for the purpose of analysing price data on a currency chart within that time period. I think this would be a common use case in finance.

Is there a possibility this could be brought into NodaTime as @lundmikkel alludes to?

**Comment by jskeet** [comment]
There's a possibility - but I'll need to look at it very carefully. (When I can find time to do so...) It's unfortunately easy to add a feature that works very well for one use case, but then find it doesn't generalize appropriately (or whatever).

One option is to have a separate package for the moment, of more experimental features... I'm not sure whether that's a good idea either though.

File: src/NodaTime/TimeInterval.cs (hunk 3 above)
Comment (a-tsymbal) [reply src/NodaTime/TimeInterval.cs]: Hello, I would like to propose another variant of the constructor that might be more useful:
```C#
public TimeInterval(LocalTime start, Duration duration)
{
    this.start = start;
    this.end = start.Plus(Period.FromNanoseconds((long)duration.TotalNanoseconds));
    IsEndInNextDay = start > end || (start == end && duration > Duration.Zero);
}
```
So when duration is zero then this interval is empty.

jskeet: That's certainly an interesting option. I think at that point I'd probably store this as a Start and Duration though, and expose that via properties (as well as an End still, of course). We could probably store the duration just as a number of nanoseconds, for efficiency.

Evengard:
I wonder if that PR was lost in limbo =)

jskeet:
@Evengard: We had a good conversation about it, but the fact that it's gone four years without having lots more attention and feature requests makes me wonder about its utility. In particular, there seem to be quite a few different ways you might want to design it for different use cases... if there were a single natural design, I'd be more inclined to go for it.

KhaledMabroukTbs (approved):

jskeet:
I'm going to close this for now, given how long it's been open and that we haven't found a natural design. I'd be happy for this to be reopened though, or a new PR created.

PR #1345, opened 2019-04-19
Template values for instant patterns

jskeet:


jskeet:
This was originally from quite a while ago - I've added tests and fixed a bug in it, but I think it's okay now. Will merge on green.

Issue #827, opened 2017-05-21
Missing Time Zone Info


SoftwareGuy88:
I have noticed that you are missing a couple entries in the database file. I have found where they should go in the WindowsZone-31.xml file. I have added them and attached the file as a txt file. I am not sure if i can just do a pull request with the update or not but i just wanted to make sure that you were aware. Otherwise you have a great product. 

Line 98   -Need to add "America/Indiana/Indianapolis"

Need to add a new time zone entries under the "(UTC+05:30) Chennai, Kolkata, Mumbai, New Delhi" section.
   --Asia/Kolkata

Update FIle.
[windowsZones-31.txt](https://github.com/nodatime/nodatime/files/1017147/windowsZones-31.txt)

jskeet:
The file was extracted from CLDR 31.0 - I've just checked, and CLDR 31.0.1 is exactly the same.

What makes you think the entries are "missing"? I'm not going to start changing the files that are directly from CLDR. There's a facility in TzdbCompiler to use an extra file for any entries we want to include, but that's not the same as modifying the original CLDR data.

SoftwareGuy88:
Well those 2 Olson TimeZones that are passed in, don't return anything. I am pulling these Olson TimeZones right from the browser. So someone is missing, just not sure who need to add them in. I have put code checks in there for those to get me buy but not sure who else is needing those time zones converted and they are getting nothing back. So i just wanted to let you know about it.

jskeet:
Yes, it's known that not every IANA time zone is represented in CLDR's Windows mapping. You'll always need some kind of fallback, as it's entirely possible that a time zone is added between CLDR releases, for example.

In the case of Asia/Kolkata, I think this may well be due to that now being the IANA zone ID, but Calcutta being the stable ID used by CLDR.

In the case of America/Indiana/Indianapolis, there's a link from America/Indianapolis to America/Indiana/Indianapolis in IANA, and America/Indianapolis is in CLDR as "US Eastern Standard Time" - not "Central Standard Time" as per your patch.

It sounds like possibly the way forward is to make it easy to check whether there's an alias name in CLDR, so you could look up "Asia/Kolkata", Noda Time would know that Asia/Calcutta is an alias for that, and that it exists in CLDR, and return you the right result. No data changes would be need.

@mj1856 any thoughts around this?

SoftwareGuy88:
[...]

PR #1348, opened 2019-04-19
Simplify mapping between Windows and TZDB time zone IDs

### LINKED ISSUE #274: Simplify Windows to/from IANA conversion methods
```
What situation does this request make simpler?
Since Noda Time contains IANA, Windows, and CLDR data, it can be used to 
convert BCL time zones to or from TZDB time zones.  The data is exposed in the 
API, but sensible conversion methods aren't included.  This could be more 
convenient, such as illustrated here:
http://stackoverflow.com/a/17348822/634824

Can you provide a straw-man example of what you'd want the API to look
like?

string bclZoneId = 
TzdbDateTimeZoneSource.Default.WindowsMapping.TranslateToBcl(tzdbZoneId);

string tzdbZoneId = 
TzdbDateTimeZoneSource.Default.WindowsMapping.TranslateToTzdb(bclZoneId);


We could also consider (in addition) a new method to DateTimeZone with the 
signature:

public DateTimeZone TranslateFor(IDateTimeZoneProvider provider)

Which would give some convenience for those not used to calling into a the 
TzdbDateTimeSource, for example:

DateTimeZone bclZone = 
DateTimeZoneProviders.Tzdb[tzdbZoneId].TranslateFor(DateTimeZoneProviders.Bcl);

DateTimeZone tzdbZone = 
DateTimeZoneProviders.Bcl[bclZoneId].TranslateFor(DateTimeZoneProviders.Tzdb);




```

Original issue reported on code.google.com by `mj1856` on 26 Mar 2014 at 6:08

jskeet:
Fixes #274.

I'd be open to the idea of adding convenience methods to TzdbDateTimeZoneSource as well, that just pass through to WindowsZones, if that were considered useful.

(Matt, are you happy to review? This was your original issue, many years ago :)

### DIFF
--- a/src/NodaTime.Test/TimeZones/Cldr/WindowsZonesTest.cs
+++ b/src/NodaTime.Test/TimeZones/Cldr/WindowsZonesTest.cs
@@ -5,6 +5,8 @@
 using NodaTime.TimeZones.Cldr;
 using NodaTime.TimeZones.IO;
 using NUnit.Framework;
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace NodaTime.Test.TimeZones.Cldr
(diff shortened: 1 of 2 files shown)


jskeet:
Note that this doesn't include anything for TZDB aliases, as described in #827.

jskeet:
If we wanted to handle TZDB aliases, it would probably make sense to *only* put the methods on `TzdbDateTimeZoneSource`, and always return canonical TZDB IDs. Maybe :)

mj1856:
I'll take a look and get back to you.  Thanks.

jskeet:
@mj1856: Thanks. The more I think about it, it more I suspect that just putting methods on TzdbDateTimeZoneSource is the way forward, to avoid the mess of having to think about CLDR-canonical vs TZDB-canonical etc.

jskeet:
I'm going to start a new branch for the "TzdbDateTimeZoneSource methods only" version.

jskeet:
Abandoning this in favour of #1393

PR #1352, opened 2019-04-21
Fix/remove some TODOs

jskeet:


jskeet:
Note: the period arithmetic changes made minimal difference (but beneficial) for most cases, but quite significant in `LocalDateTime + Period` in cases where we had to roll over to another date due to the time changes.

PR #1354, opened 2019-04-21
Duration benchmarks

jskeet:


src/NodaTime/Duration.cs
```diff
@@ -1019,15 +1019,26 @@ public static Duration FromTicks(double ticks)
...
-            long nanoOfDay = nanoseconds >= long.MinValue + NanosecondsPerDay
-                ? nanoseconds - days * NanosecondsPerDay
-                : nanoseconds - (days + 1) * NanosecondsPerDay + NanosecondsPerDay; // Avoid multiplication overflow
-            return new Duration(days, nanoOfDay);
+            unchecked
+            {
+                if (nanoseconds >= 0)
+                {
+                    // DivRem benchmarked against netcoreapp2.1 on 2019-04-21; definite win.
+                    long days = Math.DivRem(nanoseconds, NanosecondsPerDay, out var nanoOfDay);
```
malcolmr: Any reason not to `DivRem` unconditionally, instead adjusting if days is negative afterwards?

(Or are there complications with `Int64.MinValue` I'm not thinking about?)

jskeet: Only laziness, basically - I haven't thought about it carefully enough in the negative case. I may have a look later :)

PR #1364, opened 2019-04-29
More standard patterns for LocalDate and LocalTime

jskeet:


jskeet:
Reduced coverage isn't great - will see if I can improve that.

jskeet:
(BTW, do you prefer to be assigned as a reviewer, or an assignee? We tend to use assignee in google-cloud-dotnet, but we may be outliers.)

malcolmr:
(I think reviewer probably makes more sense? Assignee strikes me as the person who'll actually be merging the PR.)

PR #1367, opened 2019-05-01
Add LocalTime.FromHoursSinceMidnight and FromMinutesSinceMidnight

### LINKED ISSUE #1366: Add LocalTime.FromMinutesSinceMidnight
`LocalTime` has methods the following factory methods:

- `FromSecondsSinceMidnight()`
- `FromMillisecondsSinceMidnight()`
- `FromTicksSinceMidnight()`

Could a `FromMinutesSinceMidnight()` method be added? It would be useful for something I am trying to do (we currently store a local time in minutes since midnight in our DB, so this method would make the conversion from DB types into NodaTime types easier).

A `FromHoursSinceMidnight()` method would complete the set, but I can't see it being as useful because you can just call the hour/minute ctor with `0` minutes.

jskeet:
Fixes #1366

(FromHoursSinceMidnight is a little less useful, but I've included it for consistency/completeness.)

### DIFF
--- a/src/NodaTime/LocalTime.cs
+++ b/src/NodaTime/LocalTime.cs
@@ -303,6 +303,36 @@ namespace NodaTime
         }
 
         /// <summary>
+        /// Factory method for creating a local time from the number of minutes which have elapsed since midnight.
+        /// </summary>
+        /// <param name="minutes">The number of minutes, in the range [0, 1439]</param>
+        /// <returns>The resulting time.</returns>
+        public static LocalTime FromMinutesSinceMidnight(int minutes)
+        {
+            // Avoid the method calls which give a decent exception unless we're actually going to fail.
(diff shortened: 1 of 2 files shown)


src/NodaTime/LocalTime.cs
```diff
@@ -302,6 +302,36 @@ public static LocalTime FromSecondsSinceMidnight(int seconds)
             return new LocalTime(unchecked(seconds * NanosecondsPerSecond));
         }
 
+        /// <summary>
+        /// Factory method for creating a local time from the number of minutes which have elapsed since midnight.
+        /// </summary>
+        /// <param name="minutes">The number of seconds, in the range [0, 1439]</param>
```
malcolmr: `The number of minutes` :)

jskeet: Humbug, I thought I was being so careful! Thanks for noticing.

PR #1368, opened 2019-05-03
Add snippet for YearMonth type

carlosschults:
Some demo tests, covering mainly construction of YearMonth instances.


src/NodaTime.Demo/YearMonthDemo.cs
```diff
@@ -0,0 +1,68 @@
...
+// as found in the LICENSE.txt file.
+
+using NUnit.Framework;
+using System;
+
+namespace NodaTime.Demo
+{
+    public class YearMonthDemo
+    {
+        [Test]
```
jskeet: There's no documented constructor for this to attach to. I'd suggest removing it to avoid confusion.

carlosschults: Got it.

PR #1374, opened 2019-05-04
Introduce a date adjuster to add a period when invoked

jskeet:
I'm not fixed on the name; we could potentially omit "Period" from the name, and change "Add" to "Addition", or "Adder" etc.

I looked at introducing a similar adjuster for LocalTime, but that
could cause confusion:

```csharp
// If this were valid...
TimeAdjuster adjuster = TimeAdjuster.AddPeriod(Period.FromHours(2));
LocalDateTime start = new LocalDateTime(2019, 5, 4, 23, 0); // 11pm on May 4th

// Expectation: result is 1am on May 5th.
// Reality: result is 1am on May 4th, as the adjuster is only applied to time.
LocalDateTime result = start.With(Adjuster);
```

Fixing that would require a "LocalDateTimeAdjuster" which we don't currently have, and I'm not sure it's worth it.

### DIFF
--- a/src/NodaTime.Test/DateAdjustersTest.cs
+++ b/src/NodaTime.Test/DateAdjustersTest.cs
@@ -120,5 +120,23 @@ namespace NodaTime.Test
             Assert.Throws<ArgumentOutOfRangeException>(() => DateAdjusters.Previous(invalid));
             Assert.Throws<ArgumentOutOfRangeException>(() => DateAdjusters.PreviousOrSame(invalid));
         }
+
+        [Test]
+        public void AddPeriod_Valid()
+        {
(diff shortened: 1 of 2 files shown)


jskeet:
> This does looks reasonable, but do you have any background or rationale as to why this is useful?

It was somewhat inspired by https://stackoverflow.com/questions/55854640 - the answer that built a custom adjuster felt like it should be simpler :)

`LocalDateTime.With(Func<LocalDateTime, LocalDateTime>)` isn't terribly useful in itself, but I can see that introducing that *and* including it in `OffsetDateTime` would be useful. I can introduce those in a separate PR. Then we can have `DateTimeAdjusters.AddPeriod` which will do the expected thing.

PR #1375, opened 2019-05-04
Add some snippets for the 'OffsetDateTime' type

carlosschults:
I intend to add some more later.
[UPDATE] Added some more snippets, and fixed some mistakes that I had made in the previous ones. Squashed and force-pushed.

### DIFF
--- a/src/NodaTime.Demo/OffsetDateTimeDemo.cs
+++ b/src/NodaTime.Demo/OffsetDateTimeDemo.cs
@@ -1,0 +1,78 @@
+﻿// Copyright 2019 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using NUnit.Framework;
+
+namespace NodaTime.Demo
+{    
+    public class OffsetDateTimeDemo
+    {
+        [Test]
+        public void Construction()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18);
+            Offset offset = Offset.FromHours(-5);
+            OffsetDateTime offsetDateTime = Snippet.For(new OffsetDateTime(localDateTime, offset));
+
+            Assert.AreEqual(localDateTime, offsetDateTime.LocalDateTime);
+            Assert.AreEqual(offset, offsetDateTime.Offset);
+        }
+
+        [Test]
+        public void WithCalendar()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18, CalendarSystem.Iso);
+            Offset offset = Offset.FromHours(-5);
+            OffsetDateTime original = new OffsetDateTime(localDateTime, offset);
+            OffsetDateTime updated = Snippet.For(original.WithCalendar(CalendarSystem.Julian));
+
+            Assert.AreEqual(
+                new LocalDateTime(1985, 10, 13, 1, 18, CalendarSystem.Julian).ToString("r", null),
+                updated.LocalDateTime.ToString("r", null));
+
+            Assert.AreEqual(original.Offset, updated.Offset);
+        }
+
+        [Test]
+        public void WithOffset()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18);
+            Offset offset = Offset.FromHours(-3);
+            OffsetDateTime original = new OffsetDateTime(localDateTime, offset);
+            OffsetDateTime updated = Snippet.For(original.WithOffset(Offset.FromHours(-2)));
+
+            Assert.AreEqual(new LocalDateTime(1985, 10, 26, 2, 18), updated.LocalDateTime);
+            Assert.AreEqual(Offset.FromHours(-2), updated.Offset);
+        }
+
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/OffsetDateTimeDemo.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        [Test]
+        public void WithCalendar()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18, CalendarSystem.Gregorian);
+            Offset offset = Offset.FromHours(-5);
+            OffsetDateTime original = new OffsetDateTime(localDateTime, offset);
+            OffsetDateTime updated = Snippet.For(original.WithCalendar(CalendarSystem.Iso));
+
+            Assert.AreNotEqual(original.LocalDateTime, updated.LocalDateTime);
```
jskeet: This won't be shown clearly in the snippet. (We have a special case for AreEqual, but not AreNotEqual). I suspect it's worth constructing the expected date instead. Then use

```csharp
Assert.AreEqual(new LocalDate(...).ToString('r'), updated.LocalDateTime.ToString('r'));
```

That will show the full round trip format, including the calendar. You could then remove the last line that compares the calendar explicitly.

src/NodaTime.Demo/OffsetDateTimeDemo.cs
```diff
@@ -0,0 +1,75 @@
...
+            OffsetDateTime offsetDateTime = Snippet.For(new OffsetDateTime(localDateTime, offset));
+
+            Assert.AreEqual(localDateTime, offsetDateTime.LocalDateTime);
+            Assert.AreEqual(offset, offsetDateTime.Offset);
+        }
+
+        [Test]
+        public void WithCalendar()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18, CalendarSystem.Gregorian);
```
jskeet: It's probably worth using calendar systems where the date actually changes - ISO and Gregorian are effectively synonyms. I'd probably use ISO to start, then convert to Julian.

src/NodaTime.Demo/OffsetDateTimeDemo.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        [Test]
+        public void WithOffset()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18);
+            Offset offset = Offset.FromHours(-5);
+            OffsetDateTime original = new OffsetDateTime(localDateTime, offset);
+            OffsetDateTime updated = Snippet.For(original.WithOffset(Offset.FromHours(-3)));
+
+            Assert.AreNotEqual(original.LocalDateTime, updated.LocalDateTime);
```
jskeet: Again, construct the expected date/time and use Assert.AreEqual. You probably don't need to call ToString explicitly this time though, as the difference will be obvious enough.

src/NodaTime.Demo/OffsetDateTimeDemo.cs
```diff
@@ -0,0 +1,75 @@
...
+            Assert.AreEqual(Offset.FromHours(-3), updated.Offset);
+        }
+
+        [Test]
+        public void WithDateAdjuster()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18);
+            Offset offset = Offset.FromHours(-5);
+            OffsetDateTime original = new OffsetDateTime(localDateTime, offset);
+            OffsetDateTime updated = Snippet.For(original.With(x => x.PlusYears(30)));
```
jskeet: As of yesterday, you could use

```csharp
var dateAdjuster = DateAdjusters.AddPeriod(Period.Years(30));
OffsetDateTime updated = Snippet.For(original.With(dateAdjuster));
```

src/NodaTime.Demo/OffsetDateTimeDemo.cs
```diff
@@ -0,0 +1,75 @@
...
+            Assert.AreEqual(original.Offset, updated.Offset);
+        }
+
+        [Test]
+        public void WithTimeAdjuster()
+        {
+            LocalDateTime localDateTime = new LocalDateTime(1985, 10, 26, 1, 18);
+            Offset offset = Offset.FromHours(-5);
+            OffsetDateTime original = new OffsetDateTime(localDateTime, offset);
+            OffsetDateTime updated = Snippet.For(original.With(x => x.PlusHours(1)));
```
jskeet: This is probably not a good adjuster to use - it suggests it adds an hour to the LocalDateTime, when it doesn't really. (If you set the time to 23:30 you'll see what I mean.)

I'd use the `TimeAdjusters.AtMidnight` adjuster

carlosschults: I can't find this adjuster. It doesn't look like it's located at the TimeAdjusters class at all. Is it in some special namespace or something?

jskeet: Whoops - my mistake. Try TruncateToHour or similar.

carlosschults:
@jskeet Anything else required on my part for this one?

PR #1377, opened 2019-05-07
Add snippets for the 'Interval' type

carlosschults:


src/NodaTime.Demo/IntervalDemo.cs
```diff
@@ -0,0 +1,56 @@
...
+            Assert.AreEqual(start, interval.Start);
+            Assert.AreEqual(end, interval.End);
+        }
+
+        [Test]
+        public void Construction_WithNoBeginning()
+        {
+            Instant end = Instant.FromUtc(2019, 1, 1, 16, 25, 48);
+            Interval interval = Snippet.For(new Interval(null, end));
+            Assert.False(interval.HasStart);
```
jskeet: I don't currently support Assert.False. I can add that though.

src/NodaTime.Demo/IntervalDemo.cs
```diff
@@ -0,0 +1,56 @@
...
+            Assert.AreEqual(end, interval.End);
+        }
+
+        [Test]
+        public void Construction_WithNoEnd()
+        {
+            Instant start = Instant.FromUtc(2019, 1, 1, 16, 25, 48);
+            Interval interval = Snippet.For(new Interval(start, null));
+            Assert.False(interval.HasEnd);
```
jskeet: I think it would be better to keep this in start/end order.

src/NodaTime.Demo/IntervalDemo.cs
```diff
@@ -0,0 +1,56 @@
...
+            Assert.AreEqual(start, interval.Start);
+        }
+
+        [Test]
+        public void Construction_EternalInterval()
```
jskeet: We only support a single snippet per member at the moment. So we need two snippets - one for the `Interval(Instant, Instant)` constructor, and one for `Interval(Instant?, Instant?)`. I'd suggest joining these three examples together just with comments.

src/NodaTime.Demo/IntervalDemo.cs
carlosschults: Well, I guess I could just use `Assert.True` and negate the expression. Would that work?

src/NodaTime.Demo/IntervalDemo.cs
carlosschults: Ok, will do.

src/NodaTime.Demo/IntervalDemo.cs
jskeet: No need - I've added Assert.False support. That'll make it simpler - otherwise it would end up with `Console.WriteLine(!interval.HasStart)` which would be weird to read.

It might be worth you getting set up to generate the snippets locally and look at them before submitting a PR. That involves cloning the nodatime.org repo as a peer of the nodatime repo, and then running the snippet generator - I'm not sure whether I've got a script to just generate the snippets at the moment, but I could write one.

PR #1383, opened 2019-05-25
Additional tests to improve coverage

jskeet:


jskeet:
Merging now, although there are still a few further avenues I want to explore.

It's annoying that the bulk of the missing coverage is now in BclDateTimeZone.BclAdjustmentRule.FromUnixAdjustmentRule, which is somewhat hard to test on Windows :(

PR #1393, opened 2019-05-26
Provide dictionaries from TZDB ID to Windows ID and vice versa

### LINKED ISSUE #274: Simplify Windows to/from IANA conversion methods

jskeet:
These are canonicalized, and handle cases such as an alias being present but not the canonical ID, or vice versa.

Fixes #274

### DIFF
--- a/src/NodaTime.Test/TimeZones/TzdbDateTimeZoneSourceTest.cs
+++ b/src/NodaTime.Test/TimeZones/TzdbDateTimeZoneSourceTest.cs
@@ -490,5 +490,149 @@ namespace NodaTime.Test.TimeZones
             var source = new TzdbDateTimeZoneSource(streamData);
             Assert.Throws<InvalidNodaDataException>(source.Validate);
         }
+        
+        [Test]
+        public void WindowsToTzdbIds()
+        {
+            var source = CreateComplexMappingSource();
+            var actual = source.WindowsToTzdbIds;
+            var expected = new Dictionary<string, string>
+            {
+                { "win1", "zone1" },
+                { "win2", "zone2" },
+                { "win3", "zone3" },
+                { "win4", "zone4" },
+            };
+            CollectionAssert.AreEquivalent(expected, actual);
+        }
+
+        [Test]
+        public void TzbdToWindowsIds()
+        {
+            var source = CreateComplexMappingSource();
+            var actual = source.TzdbToWindowsIds;
+            var expected = new Dictionary<string, string>
+            {
+                { "zone1", "win1" },
+                { "link1", "win1" },                
+                { "zone2", "win2" },
+                { "link2", "win2" },
+                // No explicit zone3 mapping; link3a and link3b map to win3 and win4 respectively;
+                // link3a "wins" as it's earlier lexicographically.
+                { "zone3", "win3" },
+                { "link3a", "win3" },
+                // Explicit mapping to win4 from link3b.
+                // This is unusual, but we handle it
+                { "link3b", "win4" },
+                // link3c isn't mentioned at all, so we use the canonical ID
+                { "link3c", "win3" },
+                { "zone4", "win4" },
+                { "zone5", "win4" }
+                // zone6 isn't mapped at all
+            };
+            CollectionAssert.AreEquivalent(expected, actual);
+        }
+
+        // We'll need to adjust the documentation if either of these tests starts to fail.
+        [Test]
+        public void TzdbToWindowsIdDocumentationExamples()
+        {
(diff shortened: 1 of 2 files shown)


jskeet:
This replaces #1348 - providing two dictionaries feels simpler to me that providing specific methods.

jskeet:
@mj1856 Ping for feedback?

mj1856:
Sorry for the delayed response.

This looks pretty good.  However, you might want to also pull in aliases from CLDR that are in `/common/bcp47/timezone.xml`.  (They're not all present in `/common/supplemental/windowsZone.xml`.)

I do a few extra things in TimeZoneConverter though that may be worth handling:

- Ensure that the Windows `UTC` zone is mapped to `Etc/UTC` instead of `Etc/GMT`
- Handle both TZDB links and CLDR aliases.  For example, `Etc/GMT+0` links to `Etc/GMT` in TZDB so should map to Windows `UTC`, even though not present in CLDR. (I *think* you're doing that already, but want to verify.)
- Maintain "overrides" of mappings for zones not yet in CLDR.  I think this is enabled by #473, but not sure if you're actually including the data or not yet? [Here are the ones I'm currently tracking in TimeZoneConverter.](https://github.com/mj1856/TimeZoneConverter/blob/371cef691f6d974b6c5062674f6bee9517e8c3e1/src/TimeZoneConverter.DataBuilder/Program.cs#L49-L67)
- Add some [artificial aliases](](https://github.com/mj1856/TimeZoneConverter/blob/371cef691f6d974b6c5062674f6bee9517e8c3e1/src/TimeZoneConverter.DataBuilder/Program.cs#L49-L67)) for a few of the abbreviation-style zones that are defined in TZDB as zones and not linked anywhere in TZDB or CLDR.
- Define mappings for the [two deprecated Windows zones](https://github.com/mj1856/TimeZoneConverter/blob/371cef691f6d974b6c5062674f6bee9517e8c3e1/src/TimeZoneConverter.DataBuilder/Program.cs#L86-L88), but only in the Windows to IANA direction.

I also do some stuff with mapping for specific country codes in the Windows to IANA direction.  For example, `Eastern Standard Time` with country code `CA` returning `America/Toronto`.  Though that would require changing the API here beyond a simple dictionary.

jskeet:
Will have a look at the BCP47 part, although I think that would effectively be a separate field in the `TzdbDateTimeZoneSource`. Probably worth looking at it before merging this change though, in case it affects us.

As you've noted, we don't currently have any overrides. I'm in two minds about including them in the regular release - I'd prefer to keep it to more "official" data. Will have to mull over that.

My gut feeling is that it's worth keeping this API simple and easy to predict, but potentially have a more comprehensive mapper separately. Will spend more time looking at each of those bullet points to see... thanks very much for the info.

jskeet:
Going through the bullets one at a time:

- UTC is mapped to Etc/GMT, because that's what the XML says and Etc/GMT *is* a canonical ID. (I can see why you're doing things differently in TimeZoneConverter, but I don't think I want to invent a special rule just for UTC.
- Yes, Etc/GMT+0 maps to UTC.
- No overrides yet, but they'd just be present as if they were in the XML.
- No artificial alias support - that would probably be an override
- No deprecated Windows zone support

I've added a test for the UTC cases.

I'm pretty sure that I don't want to mix up these Windows zones with any BCP47 support, so I'll merge this with the extra test.

PR #1396, opened 2019-06-28
Provide Zone/Clock/Calendar properties on ZonedClock

### LINKED ISSUE #1362: Add Zone property to ZonedClock that returns DateTimeZone of the clock
Hi there. It came as a surprise to me that ZonedClock don't have its DateTimeZone public property.
Sometimes we need to get know the timezone of given clock, please consider to add this property:
```csharp
public DateTimeZone Zone => zone;
```

jskeet:
I hope that most uses of ZonedClock won't need them, but they don't do active harm.

Fixes #1362

### DIFF
--- a/src/NodaTime/ZonedClock.cs
+++ b/src/NodaTime/ZonedClock.cs
@@ -15,9 +15,17 @@ namespace NodaTime
     [Immutable]
     public sealed class ZonedClock : IClock
     {
-        private readonly IClock clock;
-        private readonly DateTimeZone zone;
-        private readonly CalendarSystem calendar;
+        /// <summary>Gets the clock used to provide the current instant.</summary>
+        /// <value>The time zone associated with this zoned clock.</value>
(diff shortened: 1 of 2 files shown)


src/NodaTime/ZonedClock.cs
```diff
@@ -15,9 +15,17 @@ namespace NodaTime
     [Immutable]
     public sealed class ZonedClock : IClock
     {
-        private readonly IClock clock;
-        private readonly DateTimeZone zone;
-        private readonly CalendarSystem calendar;
+        /// <summary>Gets the clock used to provide the current instant.</summary>
+        /// <value>The time zone associated with this zoned clock.</value>
```
malcolmr: The clock associated with...?

jskeet: Whoops - thanks, will fix tomorrow.

PR #1402, opened 2019-08-08
Update NUnit test adapter

jskeet:
This should make the filtered tests run much quicker.


jskeet:
Hmm. Except it doesn't. I'll need to look into that... it may be that we need to opt into something. Will merge this anyway, as updated dependencies are generally a good thing.

Issue #1403, opened 2019-08-21
Support nullable values for Min/Max


space-alien:
Just an idea - it might be nice if Min/Max supported comparing nullable dates.

If it sounds like a good addition I could have a crack at a PR.

jskeet:
Sorry, Min/Max for what, exactly? I'm not sure what you're suggesting. If you could sketch the API change in a first commit, we could iterate on that and then implement/test afterwards in other commits in the same PR.

space-alien:
Thanks Jon, I've created a PR with an example.

jskeet:
Ah, I see now. We'd presumably want to do that for all the other types with min/max values.

I'm not sure about this, to be honest. It's not something I've seen elsewhere - in the BCL, for example. I can imagine there are cases where having a *single* overload means that a method group conversion could work where it might fail with this. Given that they're static methods, it's easy enough to have them in a separate library. Could I suggest that you do that locally, and see how much you find you actually use them?

space-alien:
Yes, of course! And thanks for NodaTime :)

jskeet:
Just to capture the feature request clearly here, the suggestion is that the following method signature is added to `LocalDate`:

```csharp
public static LocalDate? Min(LocalDate? x, LocalDate? y)
```

I'll close the PR for now, but leave this feature request open with no target milestone, to see if it attracts more support. If by this time in 12 months we don't think it provides huge value, we could close it - but if before then we see more demand, we can implement it. Hope that works for you :)

PR #1404, opened 2019-08-21
LocalDate.Min() that accepts nullable values, as an example.

space-alien:
Example for #1403


jskeet:
Closing for now, while we wait to decide about #1403.

PR #1413, opened 2019-09-18
Fix nullability warnings and treat them as errors

jskeet:


src/NodaTime/Text/Patterns/PatternBclSupport.cs
```diff
@@ -29,7 +29,8 @@ internal string Format(T value, string? patternText, IFormatProvider? formatProv
                 patternText = defaultFormatPattern;
             }
             NodaFormatInfo formatInfo = NodaFormatInfo.GetInstance(formatProvider);
-            IPattern<T> pattern = patternParser(formatInfo).ParsePattern(patternText);
+            // TODO(nullable): Work out why Roslyn doesn't know about string.IsNullOrEmpty any more.
```
jnyrup: Sounds like this https://github.com/dotnet/roslyn/issues/37995

jskeet: Yup. I may end up writing my own StringIsNullOrEmpty that just calls the real one, but has the annotations we want...

PR #1415, opened 2019-09-20
More general package description

benmccallum:
... since it can be used outside the full .NET Framework now.


jskeet:
I'm not sure it's *particularly* necessary, but I have no objection. Will merge now.

benmccallum:
Yea it's picky. I just noticed it in the nuget package manager and thought, why not. Can go out whenever.

jskeet:
> Can go out whenever.

The project description in master will next go out when we do another v3 beta... it won't be part of any 2.x or 1.x releases. While we *could* modify the versions in other branches, I propose that it's not really worth doing that.

PR #1418, opened 2019-10-02
Expanded snippets

DeanBecker:
Hi @jskeet 

In reference to #918, I've started looking at contributing to your snippets library. This is a small PR just to get the green light from yourself that I've understood the requirements for your snippet generation.

Feel free to comment with a thumbs up that I'm on the right track, or merge and I'll create a new PR with any new, fresh snippets.

PR #1419, opened 2019-10-02
Add AtMidnight Snippet for LocalDate

neiesc:
Hi @jskeet 
In reference to #918.


jskeet:
Any reason for this to be separate to #1418? We can just keep updating that until it's ready to merge.

neiesc:
Sorry, we could close that.

PR #1420, opened 2019-10-02
Add snippet for Max and Min for LocalDate

neiesc:
In reference to #918.

### DIFF
--- a/src/NodaTime.Demo/LocalDateDemo.cs
+++ b/src/NodaTime.Demo/LocalDateDemo.cs
@@ -51,5 +51,23 @@ namespace NodaTime.Demo
             LocalDateTime dateTime = Snippet.For(date.AtMidnight());
             Assert.AreEqual(new LocalDateTime(2010, 6, 16, 0, 0, 0), dateTime);
         }
+
+        [Test]
+        public void Max()
+        {
+            LocalDate date = new LocalDate(2010, 6, 16);
+            LocalDate date_other = new LocalDate(2010, 6, 10);
+            LocalDate dateTime = Snippet.For(LocalDate.Max(date, date_other));
+            Assert.AreEqual(new LocalDate(2010, 6, 16), dateTime);
+        }
+
+        [Test]
+        public void Min()
+        {
+            LocalDate date = new LocalDate(2010, 6, 16);
+            LocalDate date_other = new LocalDate(2010, 6, 10);
+            LocalDate dateTime = Snippet.For(LocalDate.Min(date, date_other));
+            Assert.AreEqual(new LocalDate(2010, 6, 10), dateTime);
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -51,5 +51,23 @@ public void AtMidnight()
             LocalDateTime dateTime = Snippet.For(date.AtMidnight());
             Assert.AreEqual(new LocalDateTime(2010, 6, 16, 0, 0, 0), dateTime);
         }
+
+        [Test]
+        public void Max()
+        {
+            LocalDate date = new LocalDate(2010, 6, 16);
```
jskeet: I'd suggest renaming these to something like `earlyJune` and `lateJune` (possibly making them the 5th and 25th of the month), and save the result into `max` or `result`.

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -51,5 +51,23 @@ public void AtMidnight()
...
+            LocalDate date_other = new LocalDate(2010, 6, 10);
+            LocalDate dateTime = Snippet.For(LocalDate.Max(date, date_other));
+            Assert.AreEqual(new LocalDate(2010, 6, 16), dateTime);
+        }
+
+        [Test]
+        public void Min()
```
jskeet: Ditto previous comments.

src/NodaTime.Demo/LocalDateDemo.cs
neiesc: Any point to be in June?

neiesc: Like this?
```csharp
[Test]
public void Max()
{
    LocalDate earlyJune = new LocalDate(2010, 6, 5);
    LocalDate lateJune = new LocalDate(2010, 6, 10);
    var max = LocalDate.Max(earlyJune, lateJune);
    LocalDate dateTime = Snippet.For(max);
    Assert.AreEqual(new LocalDate(2010, 6, 16), dateTime);
}
```

jskeet: Not quite - the idea is to use `max` instead of `dateTime`. Currently that's not going to create a snippet for anything - the expression in `Snippet.For` has to be the method that we're trying to document. I meant:

```csharp
[Test]
public void Max()
{
    LocalDate earlyJune = new LocalDate(2010, 6, 5);
    LocalDate lateJune = new LocalDate(2010, 6, 25);
    var max = Snippet.For(LocalDate.Max(earlyJune, lateJune));
    Assert.AreEqual(new LocalDate(2010, 6, 16), max);
}
```

Note that I've updated `lateJune` to be June 25th, as June 10th isn't very late in June :)
(There's nothing specific about June here - I'm just making the name reflect the actual value in a simple way.)

neiesc: about june 10th i forgot to update, thanks for help!

PR #1421, opened 2019-10-02
Update Makefile

vinamrbajaj:
removed unnecessary blank line


jskeet:
Thanks for the pull request, but at this point I think it will be best to just delete the `Makefile`. It wouldn't work with the tests anyway, and it's now fine to just use `dotnet build` anyway. I'll delete the Makefile in a separate commit.

PR #1423, opened 2019-10-05
Add snippet for Deconstruct

HenrikWM:
Adds snippet for `Deconstruct` in `DateInterval`.

`Snippet.For()` doesn't have support for void-methods, so any ideas for how to support that with a new signature for `Snippet.For()` would be appreciated :) I can add that into this PR.

### DIFF
--- a/src/NodaTime.Demo/DateIntervalDemo.cs
+++ b/src/NodaTime.Demo/DateIntervalDemo.cs
@@ -71,7 +71,7 @@ namespace NodaTime.Demo
         }
 
         [Test]
-        public void Deconstruct()
+        public void Deconstruction()
         {
             var calendar = CalendarSystem.Gregorian;
             LocalDate start = new LocalDate(2017, 1, 1);
--- a/src/NodaTime.Demo/DateIntervalDemo.cs
+++ b/src/NodaTime.Demo/DateIntervalDemo.cs
@@ -84,5 +84,25 @@ namespace NodaTime.Demo
             Assert.AreEqual(start, actualStart);
             Assert.AreEqual(end, actualEnd);
         }
+
+        [Test]
+        public void Union()
+        {
+            DateInterval firstInterval = new DateInterval(
+                new LocalDate(2014, 3, 7),
+                new LocalDate(2014, 3, 20));
+
+            DateInterval secondInterval = new DateInterval(
+                new LocalDate(2014, 3, 15),
+                new LocalDate(2014, 3, 23));
+
+            DateInterval? overlappingInterval = Snippet.For(firstInterval.Union(secondInterval));
+
+            Assert.AreEqual(
+                new DateInterval(
+                    new LocalDate(2014, 3, 7),
+                    new LocalDate(2014, 3, 23)),
+                overlappingInterval);
+        }
     }
 }
--- a/src/NodaTime.Demo/DateIntervalDemo.cs
+++ b/src/NodaTime.Demo/DateIntervalDemo.cs
@@ -73,12 +73,12 @@ namespace NodaTime.Demo
         [Test]
         public void Deconstruction()
         {
-            var calendar = CalendarSystem.Gregorian;
             LocalDate start = new LocalDate(2017, 1, 1);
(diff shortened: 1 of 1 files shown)


jskeet:
Will leave this unmerged for now, as a reminder to get Snippet.For working with void methods (which I think should be easy enough, just taking an `Action`). Will try to do it in the next week or so.

src/NodaTime.Demo/DateIntervalDemo.cs
```diff
@@ -69,5 +69,23 @@ public void Contains_Interval()
...
+        {
+            var calendar = CalendarSystem.Gregorian;
+            LocalDate start = new LocalDate(2017, 1, 1);
+            LocalDate end = new LocalDate(2017, 12, 31);
+
+            DateInterval value = new DateInterval(start, end);
+
+            value.Deconstruct(out LocalDate actualStart, out LocalDate actualEnd);
+
+            Assert.Multiple(() =>
```
jskeet: Please don't use Assert.Multiple in snippets. The snippet converter doesn't "know" how to convert it into non-test code, and it doesn't make sense to add that into the code - just multiple top-level calls to Assert.AreEqual is fine. They'll end up as multiple Console.WriteLine calls.

HenrikWM: Sure :) It was a bit hard to understand how the snippets will look once generated. Is there a tool for previewing the generated docs locally? Or will https://nodatime.org/unstable/api/NodaTime.DateInterval.html be updated once a release is ready? I can create a README in https://github.com/nodatime/nodatime/tree/master/src/NodaTime.Demo to clarify, if you like.

jskeet: You can clone https://github.com/nodatime/nodatime.org as a peer of this repo, then run `build/testsnippets.sh` - that's probably the best way to go for now.

The release process is a box I have here locally running the build once every half hour, but only during certain times of day. That goes against master, so the date adjusters code is now live, for example: https://nodatime.org/unstable/api/NodaTime.DateAdjusters.html

src/NodaTime.Demo/DateIntervalDemo.cs
```diff
@@ -69,5 +69,20 @@ public void Contains_Interval()
...
+        [Test]
+        public void Deconstruction()
+        {
+            var calendar = CalendarSystem.Gregorian;
+            LocalDate start = new LocalDate(2017, 1, 1);
+            LocalDate end = new LocalDate(2017, 12, 31);
+
+            DateInterval value = new DateInterval(start, end);
+
+            value.Deconstruct(out LocalDate actualStart, out LocalDate actualEnd);
```
jskeet: Okay, I've now merged the new Snippet code. If you *add*

```csharp
Snippet.SilentForAction(() => value.Deconstruct(out var _, out var _));
```

... it should be fine. The idea is that it's "silent" because the whole statement is removed. We can't just remove the `Snippet.For` part, because of the way definite assignment works. Let me know if you need more details.

src/NodaTime.Demo/DateIntervalDemo.cs
HenrikWM: Great, thanks! I checked `DateAdjusters` and the snippets didn't use the declared variable `end`, so I've inlined `end` instead in the asserts. The snippets should not include `end` in the next release.

src/NodaTime.Demo/DateIntervalDemo.cs
HenrikWM: Great, thanks! Generated the snippet locally and this is the new generated snippet:

```csharp
using NodaTime;
using System;

LocalDate start = new LocalDate(2017, 1, 1);
LocalDate end = new LocalDate(2017, 12, 31);

DateInterval value = new DateInterval(start, end);
value.Deconstruct(out LocalDate actualStart, out LocalDate actualEnd);

Console.WriteLine(actualStart);
Console.WriteLine(actualEnd);
```

I've pushed the changes now.

PR #1424, opened 2019-10-05
Adds snippets to Period

HenrikWM:
Adds snippets for `Normalize()` and `ToDuration()` in `PeriodDemo` as per #918.

### DIFF
--- a/src/NodaTime.Demo/PeriodDemo.cs
+++ b/src/NodaTime.Demo/PeriodDemo.cs
@@ -182,12 +182,8 @@ namespace NodaTime.Demo
         public void Normalize()
         {
             var original = new PeriodBuilder { Weeks = 2, Days = 5 }.Build();
-
             var actual = Snippet.For(original.Normalize());
-
-            Assert.AreEqual(
-                new PeriodBuilder { Days = 19 }.Build(), 
-                actual);
+            Assert.AreEqual(new PeriodBuilder { Days = 19 }.Build(), actual);
         }
 
         [Test]
--- a/src/NodaTime.Demo/PeriodDemo.cs
+++ b/src/NodaTime.Demo/PeriodDemo.cs
@@ -182,23 +182,23 @@ namespace NodaTime.Demo
         public void Normalize()
         {
             var original = new PeriodBuilder { Weeks = 2, Days = 5 }.Build();
-            var expected = new PeriodBuilder { Days = 19 }.Build();
 
             var actual = Snippet.For(original.Normalize());
 
-            Assert.AreEqual(expected, actual);
+            Assert.AreEqual(
+                new PeriodBuilder { Days = 19 }.Build(), 
+                actual);
         }
 
         [Test]
         public void ToDuration()
         {
             Period oneDayAsPeriod = Period.FromDays(1);
-            Duration oneDayAsDuration = Duration.FromDays(1);
 
             var actual = Snippet.For(oneDayAsPeriod.ToDuration());
 
             Assert.IsFalse(oneDayAsPeriod.HasTimeComponent);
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -177,5 +177,28 @@ public void Between_LocalDateTimesWithGivenUnits()
             Assert.AreEqual(23, period.Hours);
             Assert.AreEqual("P2Y264DT23H", period.ToString());
         }
+
+        [Test]
+        public void Normalize()
+        {
+            var original = new PeriodBuilder { Weeks = 2, Days = 5 }.Build();
+            var expected = new PeriodBuilder { Days = 19 }.Build();
```
jskeet: Just thinking about this and your other snippets for `DateAdjuster`, we may well not see any use of `expected` given the conversion from `Assert.AreEqual` to `Console.WriteLine`. I'll hold off merging this until we've seen what those snippets look like - it may well be worth inlining the expected value into `Assert.AreEqual`.

HenrikWM: Proved to be correct decision. I have inlined them now.

src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -177,5 +177,28 @@ public void Between_LocalDateTimesWithGivenUnits()
...
         }
+
+        [Test]
+        public void Normalize()
+        {
+            var original = new PeriodBuilder { Weeks = 2, Days = 5 }.Build();
+
+            var actual = Snippet.For(original.Normalize());
+
+            Assert.AreEqual(
```
jskeet: Can we unwrap this to a single line, and possibly remove the blank lines before? Just thinking about how it'll look in the docs.

src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -177,5 +177,28 @@ public void Between_LocalDateTimesWithGivenUnits()
...
+        }
+
+        [Test]
+        public void ToDuration()
+        {
+            Period oneDayAsPeriod = Period.FromDays(1);
+
+            var actual = Snippet.For(oneDayAsPeriod.ToDuration());
+
+            Assert.IsFalse(oneDayAsPeriod.HasTimeComponent);
```
jskeet: Could you make this Assert.False instead of Assert.IsFalse? It looks like I have special handling for the former, but not the latter. I can add IsFalse/IsTrue handling later if we want.

src/NodaTime.Demo/PeriodDemo.cs
```diff
@@ -177,5 +177,28 @@ public void Between_LocalDateTimesWithGivenUnits()
...
+
+            Assert.AreEqual(
+                new PeriodBuilder { Days = 19 }.Build(), 
+                actual);
+        }
+
+        [Test]
+        public void ToDuration()
+        {
+            Period oneDayAsPeriod = Period.FromDays(1);
```
jskeet: Again, let's remove some of this whitespace to create a nice simple snippet.

src/NodaTime.Demo/PeriodDemo.cs
HenrikWM: Sure, pushed now

src/NodaTime.Demo/PeriodDemo.cs
HenrikWM: Fixed.

src/NodaTime.Demo/PeriodDemo.cs
HenrikWM: Fixed.

PR #1428, opened 2019-10-11
Add snippets for LocalDate method Add and CompareTo

neiesc:
In reference to #918.

### DIFF
--- a/src/NodaTime.Demo/LocalDateDemo.cs
+++ b/src/NodaTime.Demo/LocalDateDemo.cs
@@ -36,6 +36,14 @@ namespace NodaTime.Demo
         }
 
         [Test]
+        public void Add()
+        {
+            LocalDate date = new LocalDate(2010, 6, 16);
+            LocalDate dateTime = Snippet.For(LocalDate.Add(date, Period.FromDays(3)));
+            Assert.AreEqual(new LocalDate(2010, 6, 19), dateTime);
+        }
+
+        [Test]
         public void At()
         {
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -35,6 +35,14 @@ public void CombineWithTime()
             Assert.AreEqual(new LocalDateTime(2010, 6, 16, 16, 20, 0), dateTime);
         }
 
+        [Test]
+        public void Add()
+        {
+            LocalDate date = new LocalDate(2010, 6, 16);
+            LocalDate dateTime = Snippet.For(LocalDate.Add(date, Period.FromDays(3)));
```
jskeet: This isn't a `dateTime` - could you rename the variables to maybe `startDate` and `result`?

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -52,6 +60,15 @@ public void AtMidnight()
             Assert.AreEqual(new LocalDateTime(2010, 6, 16, 0, 0, 0), dateTime);
         }
 
+        [Test]
+        public void CompareTo()
+        {
+            LocalDate date = new LocalDate(2010, 6, 16);
+            LocalDate date_equal = new LocalDate(2010, 6, 16);
```
jskeet: I'd probably use `equalDate`, or just call them `date1` and `date2`.

PR #1434, opened 2019-10-22
Added snippets for [Interval] Deconstruct, Duration and ToString()

RexTremendae:
Added snippets for Interval.Deconstruct(out Instant, out Instant), Interval.Duration and Interval.ToString()


jskeet:
Thanks :)

PR #1446, opened 2019-11-30
Update "yyyy" to "uuuu" in tests/benchmarks/demos

### LINKED ISSUE #1445: Tests, demo code and benchmarks use yyyy a lot
As of 2.0, "y" is used for "year of era" and "u" is used for "absolute year". Most of the time, "u" is intended - and this is the specifier used in standard formats.
However, we have a lot of tests etc that still use "y". These should be fixed.

jskeet:
Fixes #1445


jskeet:
Merging even though AppVeyor is still failing - the tests are passing, it's just Chocolatey that's failing, which I expect to be a transient failure.

PR #1454, opened 2019-12-29
Avoid using http whenever possible

Youssef1313:


jskeet:
Currently this is 16 commits, each of which has a description that only says which file is being modified. Please rebase this to a *single* commit which has a description of the change ("Use https in links where possible" would be fine)

PR #1456, opened 2019-12-29
Fix typo

Youssef1313:


src/NodaTime/Calendars/HebrewScripturalCalculator.cs
```diff
@@ -171,17 +171,11 @@ internal static int GetDaysFromStartOfYearToStartOfMonth(int year, int month)
...
-            4 => 29,
-            6 => 29,
-            10 => 29,
-            13 => 29,
+            var m when m == 2 || m == 4 || m == 6 || m == 10 || m == 13 => 29,
             8 => IsHeshvanLong(year) ? 30 : 29,
             9 => IsKislevShort(year) ? 29 : 30,
             12 => IsLeapYear(year) ? 30 : 29,
-            _ => 30 // 1, 3, 5, 7, 11, 13
+            _ => 30 // 1, 3, 5, 7, 11
```
Youssef1313: This was possibly a comment typo ?
the `13` case were already handled directly.

jskeet: Yes, this was a typo. I'm fine for there to be a typo fix.

src/NodaTime/Calendars/HebrewScripturalCalculator.cs
```diff
@@ -171,17 +171,11 @@ internal static int GetDaysFromStartOfYearToStartOfMonth(int year, int month)
...
         internal static int DaysInMonth(int year, int month) => month switch
         {
-            // FIXME: How do we express multiple cases in a switch expression?
-            // We want: (2, 4, 6, 10, 13) => 29,
-            2 => 29,
-            4 => 29,
-            6 => 29,
-            10 => 29,
-            13 => 29,
+            var m when m == 2 || m == 4 || m == 6 || m == 10 || m == 13 => 29,
```
jskeet: I'd really rather not do this. I think it's less clear, and may well be less efficient too.

Youssef1313: Well. I'll revert it.
A cleaner way will be added to C# 9.0.

jskeet:
Please rebase this into a single change, with an appropriate description ("Remove typo from comment" or similar.)

I'm very keen on having a commit history which is clear and easy to read.

Youssef1313:
> 
> 
> Please rebase this into a single change, with an appropriate description ("Remove typo from comment" or similar.)
> 
> I'm very keen on having a commit history which is clear and easy to read.

Ok, I'll clone it locally first because GitHub doesn't allow to do so.
But I'm just wondering why not use the "squash" option when merging. It produces a single commit.

jskeet:
Yes, I *could* squash instead - but I think it's generally better for the person contributing the change to make it as clean as possible, minimizing the work for the maintainer to do. There's work that only the maintainer can do - and useful input that only the maintainer can provide, such as reasons for doing things a particular way. The more effort a contributor makes in terms of presenting their contribution cleanly, the more time the maintainer has for that.

PR #1459, opened 2019-12-29
Use string interpolation when possible

Youssef1313:


src/NodaTime.TzdbCompiler/Tzdb/ParserHelper.cs
```diff
@@ -73,7 +73,7 @@ internal static long ConvertSecondsWithFractionalToTicks(string text)
                                          CultureInfo.InvariantCulture);
             if (number < 0.0 || number >= 60.0)
             {
-                throw new FormatException("seconds out of valid range of [0, 60): " + number);
+                throw new FormatException($"seconds out of valid range of [0, 60]: {number}");
```
Youssef1313: There was a small typo here.

jskeet:
Please be aware that I'm not going to get a chance to review these multiple PRs probably for another week. I'll get to them eventually, but over the holidays my family takes priority.

src/NodaTime.TzdbCompiler/Tzdb/ParserHelper.cs
malcolmr: That's not a typo: `[a, b)` means a value _x_ such that _a_ ≤ _x_ < _b_. See e.g. https://en.wikipedia.org/wiki/Interval_(mathematics)#Terminology.

In this case, 0 is a valid number of seconds, but 60 isn't (but 59.9 would be).

Youssef1313: @malcolmr. Thanks for correction. I'm going to revert my bad change.

src/NodaTime/TimeZones/IO/DateTimeZoneReader.cs
```diff
@@ -152,7 +152,7 @@ public int ReadMilliseconds()
                             millis = (firstData << 24) + (ReadByte() << 16) + (ReadInt16() & 0xffff);
                             break;
                         default:
-                            throw new InvalidNodaDataException("Invalid flag in offset: " + flag.ToString("x2"));
+                            throw new InvalidNodaDataException($"Invalid flag in offset: {flag.ToString("x2")}");
```
jskeet: This would be better as `$"Invalid flag in offset: {flag:x2}"`

PR #1473, opened 2020-02-02
Add standard "month/day" pattern.

### LINKED ISSUE #1408: Missing the month/day formatting pattern
Hi all,

We recently started building a new mobile app using Noda Time and one of the first things we're working of requires us to format a date and time into a string. Now, part of this string should have a format including day (non-padded) and month name. As such, this is exactly what the [standard month/day](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#MonthDay) `DateTime` formatting [pattern "M"](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings) does.

Obviously, this format can only be used by itself, as otherwise it would cause confusion with the month number format specifier. And, true, I could exit Noda Time to revert to an unspecified `DateTime` to apply this formatting. But to me this seems like a missing feature that would be a great addition to Noda Time.

All the best,
Alexander.

jskeet:
Fixes #1408.

Another commit will add documentation, but that needs to be in the nodatime.org repo.

### DIFF
--- a/src/NodaTime.Test/Text/LocalDatePatternTest.cs
+++ b/src/NodaTime.Test/Text/LocalDatePatternTest.cs
@@ -134,6 +134,8 @@ namespace NodaTime.Test.Text
             // Invariant culture uses the crazy MM/dd/yyyy format. Blech.
             new Data(2011, 10, 20) { Pattern = "d", Text = "10/20/2011" },
             new Data(2011, 10, 20) { Pattern = "D", Text = "Thursday, 20 October 2011" },
+            // Year comes from the template
+            new Data(2000, 10, 20) { Pattern = "M", Text = "October 20" },
             // ISO pattern uses a sensible format
             new Data(2011, 10, 20) { StandardPattern = LocalDatePattern.Iso, Pattern = "R", Text = "2011-10-20" },
             // Round trip with calendar system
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/Text/LocalDatePatternTest.cs
```diff
@@ -134,6 +134,8 @@ public class LocalDatePatternTest : PatternTestBase<LocalDate>
             // Invariant culture uses the crazy MM/dd/yyyy format. Blech.
             new Data(2011, 10, 20) { Pattern = "d", Text = "10/20/2011" },
             new Data(2011, 10, 20) { Pattern = "D", Text = "Thursday, 20 October 2011" },
+            // Year comes from the template
+            new Data(2000, 10, 20) { Pattern = "M", Text = "October 20" },
```
malcolmr: I assume that a pattern of `"M"` before this is an error rather than falling back to the pattern `"%M"`?

malcolmr: Is there any benefit of adding a `LocalDateTime` test using something like `ld<M>`, or does that all just happen automatically?

jskeet: Yes.

jskeet: That happens automatically, fortunately :)

PR #1476, opened 2020-02-02
Add a type converter for ZonedDateTime

jskeet:


src/NodaTime.Test/Text/TypeConvertersTest.cs
```diff
@@ -116,18 +118,30 @@ public void HasConverter(Type type)
...
             AssertRoundtrip(text, new Period(years, months, weeks, days, hours, minutes, seconds, milliseconds, ticks, nanoseconds));
 
-        private static void AssertRoundtrip<T>(string input, T expected)
+        [Test]
+        [TestCase(2019, 1, 1, 4, 59, "Europe/Paris", "2019-01-01T04:59:00 Europe/Paris (+01)")]
+        [TestCase(2020, 2, 29, 23, 59, "America/Los_Angeles", "2020-02-29T23:59:00 America/Los_Angeles (-08)")]
+        [TestCase(2018, 12, 31, 13, 30, "Asia/Kathmandu", "2018-12-31T13:30:00 Asia/Kathmandu (+05:45)")]
+        public void ZonedDateTime_Roundtrip(int year, int month, int day, int hour, int minute, string zoneId, string text)
+        {
+            Assert.AreSame(DateTimeZoneProviders.Tzdb, DateTimeZoneProviders.ForTypeConverter,
```
malcolmr: Is this better than setting it explicitly in the test?

(Tests are run on a single thread, I assume?)

src/NodaTime.Test/Text/TypeConvertersTest.cs
```diff
@@ -116,18 +118,30 @@ public void HasConverter(Type type)
...
+
+        private static void AssertRoundtrip<T>(string textEquivalent, T nodaValue)
         {
             var converter = TypeDescriptor.GetConverter(typeof(T));
-            var actual = (T)converter.ConvertFrom(input);
+            var valueFromConverter = (T)converter.ConvertFrom(textEquivalent);
 
-            Assert.NotNull(actual);
-            Assert.AreEqual(expected, actual);
+            Assert.NotNull(valueFromConverter);
```
malcolmr: I know that this is pre-existing, but is it useful to do a non-null check immediately before an equality check? It seems like just calling `AreEqual()` would at least have the potential to give a more useful message in the event of a null?

Or, if you're concerned about `nodaValue` being null, seems like it would be reasonable to add that as a precondition at the start of the method?

(Likewise below.)

src/NodaTime/DateTimeZoneProviders.cs
```diff
@@ -81,5 +84,36 @@ public static IDateTimeZoneProvider Serialization
                 }
             }
         }
+
+        /// <summary>
+        /// Gets the <see cref="IDateTimeZoneProvider"/> to use to interpret a time zone ID read as part of
+        /// conversion using the default <see cref="System.ComponentModel.TypeConverter"/> for <see cref="ZonedDateTime"/>.
+        /// Note that if a value other than <see cref="DateTimeZoneProviders.Tzdb"/> is required, it should be set on
+        /// application startup, before any type converters are used. Type converters are cached internally by the framework,
+        /// so changes to this property after the first type converter for <see cref="ZonedDateTime"/> will not generally
```
malcolmr: Missing a word. Also I wonder if you'd be better off being less-specific here, and just saying "changes to this property after any type converters have been created will not generally take effect" or similar?

src/NodaTime.Test/Text/TypeConvertersTest.cs
jskeet: They're run on a single thread, but the concern is really that if we ever introduce another test that changes the `ForTypeConverter` property and fetches a type converter, it'll break this test without us understanding why. Just setting the property doesn't necessarily help, because of the caching.

src/NodaTime.Test/Text/TypeConvertersTest.cs
jskeet: Nope, I think it's fine to remove these - done.

src/NodaTime/DateTimeZoneProviders.cs
jskeet: While I could do that, I think it's more likely to be useful this way, because we know that type converters *are* currently cached (and that's unlikely to change, although it's possible). Type converters are often fetched implicitly, so it's not necessarily clear to the user when the type converter would be created - explicitly calling out the caching should clarify things.

PR #1487, opened 2020-03-18
Added a snippet for Max method in OffsetDemo

dekiertanna:
First pull request to make sure I follow the conventions.

### DIFF
--- a/src/NodaTime.Demo/OffsetDemo.cs
+++ b/src/NodaTime.Demo/OffsetDemo.cs
@@ -63,6 +63,19 @@ namespace NodaTime.Demo
         }
 
         [Test]
+        public void Max()
+        {
+            var offset = Offset.FromSeconds(10);
+            var offset2 = Offset.FromSeconds(20);
+            var expected = Offset.FromSeconds(20);
+
+            var actual = Snippet.For(Offset.Max(offset, offset2));
+
+            Assert.Less(offset, offset2);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
         public void Plus()
         {
             var offset = Offset.FromSeconds(100);
--- a/src/NodaTime.Demo/OffsetDemo.cs
+++ b/src/NodaTime.Demo/OffsetDemo.cs
@@ -65,14 +65,12 @@ namespace NodaTime.Demo
         [Test]
         public void Max()
         {
-            var offset = Offset.FromSeconds(10);
-            var offset2 = Offset.FromSeconds(20);
-            var expected = Offset.FromSeconds(20);
+            var smallerOffset = Offset.FromHours(3);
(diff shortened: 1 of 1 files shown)


jskeet:
Great, thanks - will look tomorrow (UK time)

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,19 @@ public void ConstructionFromTimeSpan()
             Assert.AreEqual(5400, offset.Seconds);
         }
 
+        [Test]
+        public void Max()
+        {
+            var offset = Offset.FromSeconds(10);
```
jskeet: I would call these two "smaller" and "larger", probably - and use more common offsets, e.g. one hour and three hours.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,19 @@ public void ConstructionFromTimeSpan()
...
+            var offset2 = Offset.FromSeconds(20);
+            var expected = Offset.FromSeconds(20);
```
jskeet: Rather than "expected" and "actual", I'd probably use "result" for the second variable, and not use the first at all - just inline that into the Assert.AreEqual part.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,19 @@ public void ConstructionFromTimeSpan()
...
+
+            var actual = Snippet.For(Offset.Max(offset, offset2));
+
+            Assert.Less(offset, offset2);
```
jskeet: Please remove this - we translate just a very few Assert methods into Console.WriteLine - the idea is that the snippets aren't shown as if they're tests at all. I'd expect to see something like:

**Snippet**

```csharp
var smaller = Offset.FromHours(1);
var larger = Offset.FromHours(3);

var max = Offset.Max(smaller, larger);
Console.WriteLine(max);
```

**Output**

```text
+03
```

dekiertanna:
Updated the method to follow your advice. Thank you.

jskeet:
Great - thanks! Assuming all my normal build infrastructure is still working (this bit of code is depressingly flaky around Roslyn versions) it should show up on the web site in about half an hour. I'll keep an eye and ping this PR when it's there (or if it fails).

dekiertanna:
Awesome! In the meantime, should I continue on adding demos or should I wait for the build to finish?

jskeet:
Sorry, I was on the phone - during which time the build happened. Observe the results of your handiwork: https://nodatime.org/unstable/api/NodaTime.Offset.html#NodaTime_Offset_Max_NodaTime_Offset_NodaTime_Offset_

:) So if you're as happy with that as I am, please go ahead. Thank you!

PR #1488, opened 2020-03-19
Added snippets for OffsetDemo in Demo

dekiertanna:
Added snippets for missing methods of Offset.

### DIFF
--- a/src/NodaTime.Demo/OffsetDemo.cs
+++ b/src/NodaTime.Demo/OffsetDemo.cs
@@ -63,6 +63,58 @@ namespace NodaTime.Demo
         }
 
         [Test]
+        public void Add()
+        {
+            var lefHandOffset = Offset.FromHours(5);
+            var rightHandOffset = Offset.FromHours(6);
+
+            var result = Snippet.For(Offset.Add(lefHandOffset, rightHandOffset));
+
+            Assert.AreEqual(result, lefHandOffset + rightHandOffset);
+        }
+
+        [Test]
+        public void Substract()
+        {
+            var lefHandOffset = Offset.FromHours(7);
+            var rightHandOffset = Offset.FromHours(5);
+
+            var result = Snippet.For(Offset.Subtract(lefHandOffset, rightHandOffset));
+
+            Assert.AreEqual(result, lefHandOffset - rightHandOffset);
+        }
+
+        [Test]
+        public void CompareTo()
+        {
+            var smallerOffset = Offset.FromHours(3);
+            var largerOffset = Offset.FromHours(5);
+            var equalOffset = Offset.FromHours(5);
+
+            var lessThan = smallerOffset.CompareTo(largerOffset);
+            var equalTo = largerOffset.CompareTo(equalOffset);
+            var largerThan = largerOffset.CompareTo(smallerOffset);
+
+            Assert.Less(lessThan, 0);
+            Assert.AreEqual(equalTo, 0);
+            Assert.Greater(largerThan, 0);
+        }
+
+        [Test]
+        public void Equals()
+        {
+            var equalOffset1 = Offset.FromHoursAndMinutes(1, 30);
+            var equalOffset2 = Offset.FromHoursAndMinutes(1, 30);
+            var inequalOffset = Offset.FromHours(2);
+
+            var isEqual = equalOffset1.Equals(equalOffset2);
+            var isInequal = equalOffset1.Equals(inequalOffset);
+
+            Assert.IsTrue(isEqual);
+            Assert.IsFalse(isInequal);
+        }
+
+        [Test]
         public void Max()
         {
             var smallerOffset = Offset.FromHours(3);
@@ -71,6 +123,27 @@ namespace NodaTime.Demo
             var result = Snippet.For(Offset.Max(smallerOffset, largerOffset));
 
             Assert.AreEqual(largerOffset, result);
+        }
+
+        [Test]
+        public void Min()
+        {
+            var smallerOffset = Offset.FromHours(3);
+            var largerOffset = Offset.FromHours(5);
+
+            var result = Snippet.For(Offset.Min(smallerOffset, largerOffset));
+
+            Assert.AreEqual(smallerOffset, result);
+        }
+
+        [Test]
+        public void Negate()
+        {
+            var offsetToNegate = Offset.FromHours(2);
+
+            var result = Snippet.For(Offset.Negate(offsetToNegate));
+
+            Assert.AreEqual(result, -offsetToNegate);
         }
 
         [Test]
--- a/src/NodaTime.Demo/OffsetDemo.cs
+++ b/src/NodaTime.Demo/OffsetDemo.cs
@@ -91,8 +91,12 @@ namespace NodaTime.Demo
             var largerOffset = Offset.FromHours(5);
 
             var lessThan = Snippet.For(smallerOffset.CompareTo(largerOffset));
+            var equal = smallerOffset.CompareTo(smallerOffset);
+            var greaterThan = largerOffset.CompareTo(smallerOffset);
 
             Assert.Less(lessThan, 0);
+            Assert.AreEqual(0, equal);
+            Assert.Greater(greaterThan, 0);
         }
 
         [Test]
@@ -102,8 +106,10 @@ namespace NodaTime.Demo
             var inequalOffset = Offset.FromHours(2);
 
             var isInequal = Snippet.For(offset1.Equals(inequalOffset));
+            var isEqual = offset1.Equals(offset1);
 
             Assert.False(isInequal);
+            Assert.True(isEqual);
         }
 
         [Test]
--- a/src/NodaTime.Demo/OffsetDemo.cs
+++ b/src/NodaTime.Demo/OffsetDemo.cs
@@ -61,27 +61,27 @@ namespace NodaTime.Demo
             Offset offset = Snippet.For(Offset.FromTimeSpan(timespan));
             Assert.AreEqual(5400, offset.Seconds);
         }
-
+        
         [Test]
         public void Add()
         {
-            var lefHandOffset = Offset.FromHours(5);
+            var leftHandOffset = Offset.FromHours(5);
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
...
 
+        [Test]
+        public void Add()
+        {
+            var lefHandOffset = Offset.FromHours(5);
+            var rightHandOffset = Offset.FromHours(6);
+
+            var result = Snippet.For(Offset.Add(lefHandOffset, rightHandOffset));
+
+            Assert.AreEqual(result, lefHandOffset + rightHandOffset);
```
jskeet: We need `result` to be on the RHS - basically this will currently be replaced with

```csharp
Console.WriteLine(leftHandOffset + rightHandOffset);
```

whereas we want to print `result`.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
             Assert.AreEqual(5400, offset.Seconds);
         }
 
+        [Test]
+        public void Add()
+        {
+            var lefHandOffset = Offset.FromHours(5);
```
jskeet: I suspect just calling these `offset1` and `offset2` is good enough here.
But if you want to keep the names, we need a "t" in left :)

src/NodaTime.Demo/OffsetDemo.cs
jskeet: In order to make the expected result more obvious, I'd probably use `Offset.FromHours(11)` too:

```csharp
Assert.AreEqual(Offset.FromHours(11), result);
```

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
...
+
+        [Test]
+        public void Substract()
+        {
+            var lefHandOffset = Offset.FromHours(7);
+            var rightHandOffset = Offset.FromHours(5);
+
+            var result = Snippet.For(Offset.Subtract(lefHandOffset, rightHandOffset));
+
+            Assert.AreEqual(result, lefHandOffset - rightHandOffset);
```
jskeet: Ditto comments from above.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
...
+        {
+            var smallerOffset = Offset.FromHours(3);
+            var largerOffset = Offset.FromHours(5);
+            var equalOffset = Offset.FromHours(5);
+
+            var lessThan = smallerOffset.CompareTo(largerOffset);
+            var equalTo = largerOffset.CompareTo(equalOffset);
+            var largerThan = largerOffset.CompareTo(smallerOffset);
+
+            Assert.Less(lessThan, 0);
```
jskeet: This won't currently work (IIRC) - although I could make it work.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
...
+            Assert.AreEqual(equalTo, 0);
```
jskeet: This should be `0, equalTo`.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
...
+        }
+
+        [Test]
+        public void CompareTo()
+        {
+            var smallerOffset = Offset.FromHours(3);
+            var largerOffset = Offset.FromHours(5);
+            var equalOffset = Offset.FromHours(5);
+
+            var lessThan = smallerOffset.CompareTo(largerOffset);
```
jskeet: We need Snippet.For on just one of these calls.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -62,6 +62,58 @@ public void ConstructionFromTimeSpan()
...
+        public void Equals()
+        {
+            var equalOffset1 = Offset.FromHoursAndMinutes(1, 30);
+            var equalOffset2 = Offset.FromHoursAndMinutes(1, 30);
+            var inequalOffset = Offset.FromHours(2);
+
+            var isEqual = equalOffset1.Equals(equalOffset2);
+            var isInequal = equalOffset1.Equals(inequalOffset);
+
+            Assert.IsTrue(isEqual);
```
jskeet: We don't support IsTrue and IsFalse, but we do support the True and False methods.
I can add support for those if you want, but it's probably simplest to just use Assert.True/False for now.

See https://github.com/nodatime/nodatime.org/blob/master/build/SnippetExtractor/SnippetRewriter.cs#L101 for the transformations we perform.

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -73,6 +125,27 @@ public void Max()
...
+        }
+
+        [Test]
+        public void Negate()
+        {
+            var offsetToNegate = Offset.FromHours(2);
+
+            var result = Snippet.For(Offset.Negate(offsetToNegate));
+
+            Assert.AreEqual(result, -offsetToNegate);
```
jskeet: Again, result needs to be second - and I'd use `Assert.AreEqual(Offset.FromHours(-2), result);`

src/NodaTime.Demo/OffsetDemo.cs
dekiertanna:
```
            var lessThan = smallerOffset.CompareTo(largerOffset);
            var equalTo = largerOffset.CompareTo(Snippet.For(Offset.FromHours(5)));
            var largerThan = largerOffset.CompareTo(smallerOffset);
```

Is this what is required? Just getting sure. Does the same follows for the method Equals, which currently looks like this:

```
 public void Equals()
        {
            var equalOffset1 = Offset.FromHoursAndMinutes(1, 30);
            var equalOffset2 = Offset.FromHoursAndMinutes(1, 30);
            var inequalOffset = Offset.FromHours(2);

            var isEqual = equalOffset1.Equals(equalOffset2);
            var isInequal = equalOffset1.Equals(inequalOffset);

            Assert.True(isEqual);
            Assert.False(isInequal);
        }
```

src/NodaTime.Demo/OffsetDemo.cs
dekiertanna: I think it would be beneficial to make these work. :)

dekiertanna:
Sorry for my late reply. I resolved the issues that has been pointed out, but I have a few questions I posted under highlighted snippets.

Hope my mistakes aren't too much of a nuisance!

jskeet:
Will have a look this evening. One thing you might want to do is clone the nodatime.org repo (https://github.com/nodatime/nodatime.org) and run `build/testsnippets.sh` (possibly tweaking it a bit if necessary) so you can see what the generated snippets look like. I'm very happy to hear ideas for how they can be improved :)

Thanks for all your efforts!

src/NodaTime.Demo/OffsetDemo.cs
jskeet: Yes - basically the purpose of `Snippet.For` is just to tell the snippet tooling "please include this snippet with the specified member" - so it's only needed once. But it *is* needed once - that Equals doesn't have any.

src/NodaTime.Demo/OffsetDemo.cs
jskeet: Okay, I'll try to do that tomorrow.

jskeet:
I don't think you've pushed the changes you've made yet, btw.

jskeet:
I've now merged the new features for snippets:

- Handle IsTrue and IsFalse like True and False
- Handle Less/Greater, so long as the second argument is 0

jskeet:
Don't worry about the AppVeyor issue - that's a dotCover problem I need to fix. I'll handle that separately.

dekiertanna:
Thanks! I compiled and am using SnippetExtractor now, so hopefully there will be less errors from now on. :)

jskeet:
Hooray. I don't know whether I'll get to all the coverage etc changes today - it depends on how tricky they are.

jskeet:
Still not entirely sure why the coverage was failing, but I've updated the projects to netcoreapp3.1 anyway. I've restarted the AppVeyor build - let's see what happens. Will merge anyway if I don't find anything else. Reviewing now.

src/NodaTime.Demo/OffsetDemo.cs
jskeet: I may have miscommunicated before - I think it's still worth doing greater than and equal to - you just don't need to include Snippet.For multiple times. So:

```csharp
 var smallerOffset = Offset.FromHours(3);
var largerOffset = Offset.FromHours(5);

var lessThan = Snippet.For(smallerOffset.CompareTo(largerOffset));
var equal = smallerOffset.CompareTo(smallerOffset);
var greaterThan = largerOffset.CompareTo(smallerOffset);

Assert.Less(lessThan, 0);
Assert.AreEqual(0, equal);
Assert.Less(greaterThan, 0);
```

src/NodaTime.Demo/OffsetDemo.cs
```diff
@@ -61,6 +61,50 @@ public void ConstructionFromTimeSpan()
...
+            Assert.Less(lessThan, 0);
+        }
+
+        [Test]
+        public void Equals()
+        {
+            var offset1 = Offset.FromHoursAndMinutes(1, 30);
+            var inequalOffset = Offset.FromHours(2);
+
+            var isInequal = Snippet.For(offset1.Equals(inequalOffset));
```
jskeet: Add an equal one as well? (Again, no need for Snippet.For.)

dekiertanna: ok! Working on it.

PR #1489, opened 2020-03-20
Delete global.json file

0xced:
Quoting Andrew Lock from https://andrewlock.net/exploring-the-new-rollforward-and-allowprerelease-settings-in-global-json/

> Having the additional flexibility to define ranges of SDK versions is definitely useful, but should be used sparingly where possible. It can be easy to add accidental onerous requirements on people trying to build your project. *Only add a global.json where it is necessary*, and try to use permissive roll-forward policies like latestMajor or latestMinor where possible.

Another benefit is that JetBrains Rider currently doesn't support global.json properly, see https://youtrack.jetbrains.com/issue/RIDER-38533


jskeet:
No, I'm definitely not going to do this. Feel free to delete it in your fork, but I find it very useful to clearly state the version of the SDK we expect. It doesn't affect end users at all, only other developers - and as the one who does the most development on the project, I'm willing to act somewhat dictatorially on this one.

PR #1490, opened 2020-03-20
Add rollForward=latestMinor in global.json

0xced:
So that people can work with NodaTIme with any .NET Core 3.x SDK, see https://andrewlock.net/exploring-the-new-rollforward-and-allowprerelease-settings-in-global-json/ 

Note that rollForward policies are not (yet) supported by Rider: https://youtrack.jetbrains.com/issue/RIDER-38533

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "3.0.100"
+    "version": "3.0.100",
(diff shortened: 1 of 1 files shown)


0xced:
Proposing a less drastic way to cope with global.json and multiple versions of the .NET Core SDK. 😉

PR #1495, opened 2020-03-22
Make LoadFileOrUrl truly async

0xced:
It's very easy to adapt command line tools since C# 7.1 and async Main support.

### DIFF
--- a/src/NodaTime.Tools.Common/FileUtility.cs
+++ b/src/NodaTime.Tools.Common/FileUtility.cs
@@ -10,25 +10,15 @@ namespace NodaTime.Tools.Common
 {
     public static class FileUtility
     {
-        public static byte[] LoadFileOrUrl(string source)
+        public static async Task<byte[]> LoadFileOrUrlAsync(string source)
         {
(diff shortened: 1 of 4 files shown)


src/NodaTime.Tools.Common/FileUtility.cs
```diff
@@ -10,24 +10,19 @@ namespace NodaTime.Tools.Common
...
             return File.ReadAllBytes(source);
 
             async Task<byte[]> FetchContentAsync(string url)
             {
-                using (var client = new HttpClient())
-                {
-                    var response = await client.GetAsync(source);
-                    return await response.EnsureSuccessStatusCode().Content.ReadAsByteArrayAsync();
-                }
+                using var client = new HttpClient();
```
jskeet: Possibly inline this code into the "if" now? When we had to call Task.Run it was helpful to have the local method, but I don't think it makes sense now.

0xced: Absolutely, fixed and force-pushed.

PR #1498, opened 2020-03-23
Improve nzd files validation

0xced:
* Validate nzd files with a progress bar
* Validate nzd files in parallel for a nice speed boost

See screen captures: 12+ seconds vs 2+ seconds! 🏎💨

ValidateHistoricalNzd-slow
ValidateHistoricalNzd-fast

### DIFF
--- a/src/NodaTime.Tools.ValidateHistoricalNzd/NodaTime.Tools.ValidateHistoricalNzd.csproj
+++ b/src/NodaTime.Tools.ValidateHistoricalNzd/NodaTime.Tools.ValidateHistoricalNzd.csproj
@@ -13,6 +13,7 @@
     <ProjectReference Include="..\NodaTime\NodaTime.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="AsyncEnumerator" Version="4.0.2" />
     <PackageReference Include="ShellProgressBar" Version="5.0.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


0xced:
> when everything's done in parallel it's going to be harder to diagnose when there _is_ a problem.

That should not be a problem, I'm aggregating all the exceptions and printing them (with the source URL) in case of failure.

> Do you have a particular reason for wanting to change this?

Just like the previous pull requests with proper async handling: set an example on good practices. In this case, parallelise what is parallel and display a clear progress, even in a CLI app.

Also, in a few months/years, when I won't remember the names of the libraries I used for async parallelism ([AsyncEnumerator](https://github.com/Dasync/AsyncEnumerable)) and CLI progress  ([ShellProgressBar](https://github.com/Mpdreamz/shellprogressbar)), I'll remember to look at NodaTime. 😉

jskeet:
See, I don't view "parallelize for the sake of it" as good practice. It adds complexity, which should always have a concrete value proposition attached. And the progress is already clear in terms of what it's doing - as well as being logfile friendly. (I haven't tried doing this as part of a Travis build or similar, but what would happen?)

0xced:
> See, I don't view "parallelize for the sake of it" as good practice. It adds complexity, which should always have a concrete value proposition attached.

I think 2 seconds vs 12 seconds is a pretty good value proposition. But you're absolutely right about the logfile friendliness that I had not considered and which is probably more important than speed in a continuous integration scenario.

I should probably write a blog post about it rather than this pull request. 😉

PR #1499, opened 2020-03-23
Added AnnualDate snippets in Demo

dekiertanna:
+AnnualDateDemo.cs in NodaTime.Demo

### DIFF
--- a/src/NodaTime.Demo/AnnualDateDemo.cs
+++ b/src/NodaTime.Demo/AnnualDateDemo.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2019 The Noda Time Authors. All rights reserved.
+﻿// Copyright 2020 The Noda Time Authors. All rights reserved.
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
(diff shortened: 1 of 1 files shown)


src/NodaTime.Demo/AnnualDateDemo.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright 2019 The Noda Time Authors. All rights reserved.
```
jskeet: Nit: please update to 2020 :)

src/NodaTime.Demo/AnnualDateDemo.cs
```diff
@@ -0,0 +1,60 @@
...
+using NUnit.Framework;
+using System;
+namespace NodaTime.Demo
+{
+    class AnnualDateDemo
+    {
+        [Test]
+        public void CompareTo()
+        {
+            var earlierDate = new AnnualDate(2, 23);
```
jskeet: I think it might be useful to name these `february23rd` and `august17th` for simplicity - ditto later.

That way, we don't end up with variables called `date` which don't really represent a full date, too.

src/NodaTime.Demo/AnnualDateDemo.cs
```diff
@@ -0,0 +1,60 @@
...
+            var equal = date1.Equals(date1);
+
+            Assert.False(unequal);
+            Assert.True(equal);
+        }
+
+        [Test]
+        public void InYear()
+        {
+            var annualDate = Snippet.For(new AnnualDate(3, 12));
```
jskeet: This Snippet.For should be round the InYear, not round the constructor call.

src/NodaTime.Demo/AnnualDateDemo.cs
```diff
@@ -0,0 +1,60 @@
...
+            var localDate = annualDate.InYear(2013);
+
+            Assert.AreEqual(new LocalDate(2013, 3, 12), localDate);
+        }
+
+        [Test]
+        public void IsValidYear()
+        {
+            var date = new AnnualDate(2, 29);
```
jskeet: I'd probably call this `leapDay`.

jskeet:
Thank you for all of these - a few nits, but nothing that should take long to address.

PR #1500, opened 2020-04-07
Add support for XML Schema Definition

0xced:
* This pull request introduces a new `XmlSchemaDefinition` public static class with a `NodaTimeXmlSchema` property describing the structure for all NodaTime types that implement the [IXmlSerializable](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.ixmlserializable) interface. This is typically used by the [XmlSchemaImporter](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.xmlschemaimporter) class.

* ~~The `AnnualDate` type now implements the `IXmlSerializable` interface.~~ Was done in #1502

* `NodaTime.Demo.XmlSchemaDemo` was also added to demonstrate how the NodaTime XML schema can be used to create C# classes from an [XmlSchemaSet](https://docs.microsoft.com/en-us/dotnet/api/system.xml.schema.xmlschemaset).


jskeet:
Just to set expectations, it may well be a few days before I can get to this. Maybe not, but we'll see.

0xced:
OK, I'll do this on another branch, currently on https://github.com/0xced/nodatime/tree/unit-tests-multi-frameworks which sits on top of https://github.com/0xced/nodatime/tree/XmlSchemaDefinition.

src/NodaTime.Test/XmlSchemaTest.XmlSchema.approved.xml
```diff
@@ -0,0 +1,709 @@
...
+      <xs:pattern value="-?[0-9]{4}-[0-9]{2}-[0-9]{2}" />
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="localDateTime">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="-?[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]{1,9})?" />
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="zone">
+    <xs:restriction base="xs:string">
```
jskeet, line 113: I haven't looked at the code yet, but this makes me nervous. The list of valid time zone IDs changes over time - I don't think it should be part of the schema unless it really has to be. (I wouldn't want an XML document to fail validation with a valid time zone ID just because the ID wasn't known when the schema was created.) Happy to be guided by you on this though.

src/NodaTime.Test/XmlSchemaTest.cs
```diff
@@ -0,0 +1,298 @@
...
+            namespaceManager.AddNamespace(NodaTimeXmlNamespace.Name, NodaTimeXmlNamespace.Namespace);
+            var xmlSchema = GetXmlSchema<NodaTimeObject>();
+            var xml = GetXmlSchemaString(xmlSchema, namespaceManager);
+            Approvals.VerifyWithExtension(xml, "xml");
+        }
+
+        [Test]
+        public void ValidXmlDocumentWithAllTypes()
+        {
+            var xml = "<?xml version=\"1.0\"?>" +
```
jskeet, line 39: I think a verbatim string literal would be clearer here.

src/NodaTime/AnnualDate.cs
```diff
@@ -28,7 +32,8 @@ namespace NodaTime
     /// </para>
     /// </remarks>
     [TypeConverter(typeof(AnnualDateTypeConverter))]
-    public readonly struct AnnualDate : IEquatable<AnnualDate>, IComparable<AnnualDate>, IComparable, IFormattable
+    [XmlSchemaProvider(nameof(GetSchema))]
+    public readonly struct AnnualDate : IEquatable<AnnualDate>, IComparable<AnnualDate>, IComparable, IFormattable, IXmlSerializable
```
jskeet, line 36: I would rather not add XML serialization for AnnualDate in the same PR as support for XML schemas. I don't mind which order we do them in, but I'd really like to do one at a time.

src/NodaTime/Duration.cs
```diff
@@ -1123,6 +1124,13 @@ public static Duration FromTimeSpan(TimeSpan timeSpan)
         public TimeSpan ToTimeSpan() => new TimeSpan(BclCompatibleTicks);
 
         #region XML serialization
+        /// <summary>
+        /// Adds the XML schema type describing the structure of the <see cref="Duration"/> XML serialization to the given <paramref name="xmlSchemaSet"/>.
+        /// </summary>
+        /// <param name="xmlSchemaSet">The XML schema set provided by <see cref="XmlSchemaExporter"/>.</param>
+        /// <returns>The qualified name of the schema type that was added to the <paramref name="xmlSchemaSet"/>.</returns>
+        public static XmlQualifiedName GetSchema(XmlSchemaSet xmlSchemaSet) => XmlSchemaDefinition.AddDurationSchemaType(xmlSchemaSet);
```
jskeet, line 1132: Is it still correct for IXmlSerializable.GetSchema to return null below?

jskeet, line 1132: Could the AddDurationSchemaType (etc) methods be extension methods, so we can just call `xmlSchemaSet.AddDurationSchemaType()`?

src/NodaTime/Duration.cs
```diff
@@ -61,6 +61,7 @@ namespace NodaTime
     /// </remarks>
     /// <threadsafety>This type is an immutable value type. See the thread safety section of the user guide for more information.</threadsafety>
     [TypeConverter(typeof(DurationTypeConverter))]
+    [XmlSchemaProvider(nameof(GetSchema))]
```
jskeet, line 64: Given that we get to pick the name here, could we use `AddSchema`? It feels like that describes what it's doing more than `GetSchema`.

src/NodaTime/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,241 @@
...
+    /// <summary>
+    /// Provides XML schema types.
+    /// </summary>
+    public static class XmlSchemaDefinition
+    {
+        /// <summary>
+        /// Gets the XML namespace for all NodaTime types.
+        /// </summary>
+        /// <remarks>See [Namespaces in XML 1.1 (Second Edition)](https://www.w3.org/TR/xml-names11/).</remarks>
+        public static readonly XmlQualifiedName NodaTimeXmlNamespace = new XmlQualifiedName("nodatime", "https://nodatime.org/api/");
```
jskeet, line 25: Please make this a property - at 2.0 I removed all public static fields, even readonly ones.

```suggestion
        public static XmlQualifiedName NodaTimeXmlNamespace { get; } = new XmlQualifiedName("nodatime", "https://nodatime.org/api/");
```

(Unless there's some particular reason it needs to be a field.)

src/NodaTime/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,241 @@
...
+using System.Xml.Schema;
+using System.Xml.Serialization;
+using NodaTime.Extensions;
+
+namespace NodaTime
+{
+    /// <summary>
+    /// Provides XML schema types.
+    /// </summary>
+    public static class XmlSchemaDefinition
```
jskeet, line 19: Does this actually need to be public? I haven't looked at the demo code yet - does it need to refer to this?

src/NodaTime/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,241 @@
...
+                    [LocalDateSchemaType.Value] = new[] { localDateRestriction.Value, calendarRestriction.Value },
+                    [LocalDateTimeSchemaType.Value] = new[] { localDateTimeRestriction.Value, calendarRestriction.Value },
+                    [ZonedDateTimeSchemaType.Value] = new[] { OffsetDateTimeSchemaType.Value, calendarRestriction.Value, zoneRestriction.Value },
+                }
+            );
+
+            XmlSchema = new Lazy<XmlSchema>(CreateNodaTimeXmlSchema);
+        }
+
+        private static readonly Lazy<XmlSchema> XmlSchema;
```
jskeet, line 106: Do these definitely need to be lazy? If we're only going to get here when XML schemas are being used (which I suspect is a small subset of code) is it worth doing?

src/NodaTime/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,241 @@
...
+        private const string DurationPattern = @"-?[0-9]{1,8}:[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]{1,9})?";
+
+        static XmlSchemaDefinition()
+        {
+            var xsStringType = XmlSchemaType.GetBuiltInSimpleType(XmlTypeCode.String);
+
+            var calendarRestriction = new Lazy<XmlSchemaType>(() => CreateEnumerationRestriction("calendar", xsStringType, CalendarSystem.Ids));
+            var localDateRestriction = new Lazy<XmlSchemaType>(() => CreatePatternRestriction("localDate", xsStringType, $"{YearPattern}-{MonthPattern}-{DayPattern}"));
+            var localDateTimeRestriction = new Lazy<XmlSchemaType>(() => CreatePatternRestriction("localDateTime", xsStringType,  $"{YearPattern}-{MonthPattern}-{DayPattern}T{TimePattern}"));
+            var zoneRestriction = new Lazy<XmlSchemaType>(() => CreateEnumerationRestriction("zone", xsStringType, DateTimeZoneProviders.Serialization.GetAllZones().Select(e => e.Id)));
```
jskeet, line 66: As before, this makes me really nervous.

src/NodaTime/Duration.cs
0xced: Yes, here's an excerpt from the remarks of the [IXmlSerializable.GetSchema](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.ixmlserializable.getschema) documentation (emphasis mine):
> When serializing or deserializing an object, the [XmlSerializer](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.xmlserializer) class does not perform XML validation. For this reason, it is often safe to omit schema information by providing a trivial implementation of this method, for example by returning `null` (`Nothing` in Visual Basic).
> 
> If providing schema information is necessary, you should use the [XmlSchemaProviderAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.xmlschemaproviderattribute) attribute. **The `GetSchema` method should still return `null` (or `Nothing`)**.

src/NodaTime/AnnualDate.cs
0xced: Done and you already merged #1502, thanks!

src/NodaTime.Test/XmlSchemaTest.cs
0xced: You're right, I will fix it and force push.

src/NodaTime.Test/XmlSchemaTest.XmlSchema.approved.xml
0xced: It doesn't *have to be*. You could define a schema with everything being `xs:string` without any restriction but that schema would not be very valuable. But you're absolutely right that rejecting a valid document would be bad. I'll see if it's possible to formally document the list of known valid time zone values while also validating any string, thus accommodating a potential new time zone.

src/NodaTime/Duration.cs
0xced: It could be. I ended up not doing it to avoid having to add `using NodaTime.XmlSchemaDefinition;` in the 14 files with types implementing the IXmlSerializable interface but if you prefer, I can turn them into extension methods.

src/NodaTime/Duration.cs
0xced: Actually, the method does both. It adds the schema to the provided `XmlSchemaSet` and returns its qualified name. The documentation about the [XmlSchemaProviderAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.xmlschemaproviderattribute) uses `MySchema` in its example which is even worse than `GetSchema` and `AddSchema`.
But given that all the internal methods are named `Add<NodaType>SchemaType` probably `AddSchema` makes more sense. I'll change the names to `AddSchema`.

src/NodaTime/XmlSchemaDefinition.cs
0xced: There was no particular reason, I changed it to a property.

src/NodaTime/XmlSchemaDefinition.cs
0xced: Yes, both `XmlSchemaDefinition.NodaTimeXmlNamespace` and `XmlSchemaDefinition.NodaTimeXmlSchema` need to be accessed in the demo code. These are the only two Xml schema related public properties. Everything else is internal or private.

src/NodaTime/XmlSchemaDefinition.cs
0xced: Not sure why I went crazy with all these lazy. I removed all the lazy types and created everything in the static constructor. The resulting code is much less cluttered!

src/NodaTime/XmlSchemaDefinition.cs
0xced: Do you mean that like time zone ids, the calendar ids may change over time?

jskeet: Nope, this was a reiteration of the zone ID concern.

src/NodaTime/Duration.cs
jskeet: Reasonable point about the using directive. I'm fine with keeping them as regular static methods - especially if it's all internal anyway.

src/NodaTime/XmlSchemaDefinition.cs
jskeet: (Please don't resolve a conversation until we've reached a consensus - it makes it hard to find things that still need discussion.)

Okay, if it needs to be public, I'd really like it to be out of the NodaTime namespace, given that it's not going to be significant for most users. I don't mind adding a NodaTime.Xml namespace for this - we could even potentially move the static NodaTime.DateTimeZoneProviders.Serialization property into a type within that namespace as well, to be clearer.

0xced: Sorry, I used _resolve conversation_ for me to keep track of what I replied to, I didn't realise that the conversation would be marked as resolved for everyone.

I moved these two properties in a new `NodaTime.Xml` namespace. I haven't moved the `NodaTime.DateTimeZoneProviders.Serialization` into that new namespace as part of this pull request though. We can probably do it separately. Also, this is a breaking change but maybe you don't mind since you are going to release a new major version?

0xced:
> I'd be interested to know whether this has to be done using these attributes, or whether it could be separated into a different project with a GetXmlSchema method or something like that.

Well, the [XmlSchemaProviderAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.xmlschemaproviderattribute) is really part of the [IXmlSerializable](https://docs.microsoft.com/en-us/dotnet/api/system.xml.serialization.ixmlserializable) interface (although in a bizarre way with an attribute and method signature convention). So I don't see how this could be separated into a different project.

I rebased on master and force pushed. All the concerns that you raised in the comments should have been addressed except for the *hardcoded* time zone ids where I still have to give consideration.

jskeet:
> Well, the XmlSchemaProviderAttribute is really part of the IXmlSerializable interface (although in a bizarre way with an attribute and method signature convention). So I don't see how this could be separated into a different project.

Understood. Happy to keep it in if we can't do anything else.

src/NodaTime/XmlSchemaDefinition.cs
jskeet: Definitely best to keep the "moving to a different namespace" for a separate PR - I'll take care of that after this one's in. And yes, it's a breaking change - I'll document it when I make the change.

src/NodaTime/Xml/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        private static XmlSchema CreateNodaTimeXmlSchema()
+        {
+            var schemaSetForCollecting = new XmlSchemaSet();
+            var xmlSerializableTypes = typeof(XmlSchemaDefinition).Assembly.GetExportedTypes().Where(t => t.GetInterfaces().Contains(typeof(IXmlSerializable)));
+            foreach (var xmlSerializableType in xmlSerializableTypes)
+            {
+                var schemaProvider = xmlSerializableType.GetCustomAttribute<XmlSchemaProviderAttribute>();
+                if (schemaProvider == null)
```
jskeet, line 145: Sorry for missing this before - please add braces around all single-line if statement bodies etc.

src/NodaTime/Xml/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,238 @@
...
+                }
+            }
+            schemaSet.Add(schema);
+            return schemaType.QualifiedName;
+        }
+
+        private static XmlSchema CreateNodaTimeXmlSchema()
+        {
+            var schemaSetForCollecting = new XmlSchemaSet();
+            var xmlSerializableTypes = typeof(XmlSchemaDefinition).Assembly.GetExportedTypes().Where(t => t.GetInterfaces().Contains(typeof(IXmlSerializable)));
```
jskeet, line 141: Rather than doing all of this via reflection, how about we have a hard-coded list of serializable types, and then a unit test that *does* use reflection to make sure it's accurate?

src/NodaTime/Xml/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,238 @@
...
+            foreach (var xmlSerializableType in xmlSerializableTypes)
+            {
+                var schemaProvider = xmlSerializableType.GetCustomAttribute<XmlSchemaProviderAttribute>();
+                if (schemaProvider == null)
+                    throw new InvalidOperationException($"The type '{xmlSerializableType}' is missing the '{nameof(XmlSchemaProviderAttribute)}' on its class definition.");
+                var schemaProviderMethod = xmlSerializableType.GetMethod(schemaProvider.MethodName, BindingFlags.Public | BindingFlags.Static);
```
jskeet, line 147: Again, do we actually need to use reflection here? It feels like we could manually call all the methods within this class. I understand we still need the attribute for XmlSchemaExporter, but I don't think we need to use it when we're doing the work ourselves. Unless I've missed something, which is entirely possible.

src/NodaTime/Xml/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,238 @@
...
+                content.Attributes.Add(attribute);
+            }
+            return new XmlSchemaComplexType
+            {
+                Name = typeof(T).Name,
+                ContentModel = new XmlSchemaSimpleContent { Content = content }
+            };
+        }
+
+        private static XmlSchemaSimpleType CreatePatternRestriction<T>(XmlSchemaType baseType, string pattern) where T : IXmlSerializable
```
jskeet, line 187: Use an expression-bodied member for this?

src/NodaTime/Xml/XmlSchemaDefinition.cs
```diff
@@ -0,0 +1,238 @@
...
+        {
+            return CreatePatternRestriction(typeof(T).Name, baseType, pattern);
+        }
+
+        private static XmlSchemaSimpleType CreatePatternRestriction(string name, XmlSchemaType baseType, string pattern)
```
jskeet, line 192: And again, an expression-bodied member?

src/NodaTime.Demo/NodaTime.Demo.csproj
```diff
@@ -17,4 +17,26 @@
     <PackageReference Include="NUnit3TestAdapter" Version="3.16.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
   </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFrameworkIdentifier) == '.NETFramework'">
```
jskeet, line 21: Now that we're not changing the targets, can we remove all the conditionality here, and just have a single demo source file?

src/NodaTime.Demo/XmlSchemaDemo.NetCore.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Xml;
+using System.Xml.Schema;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using NodaTime.Xml;
+using XmlSchemaClassGenerator;
+
+namespace NodaTime.Demo
```
jskeet, line 14: If all of this is really for XmlSchemaDemo, could we make all the supporting types nested types within that? You could potentially use partial classes and keep two files (but have one "nested" in the other effectively in Visual Studio). I just want to avoid anyone getting confused at all if they're not looking for this feature.

src/NodaTime/Xml/XmlSchemaDefinition.cs
0xced: Absolutely, done in 6c06ed466ce2770bc44e2b7d4e0199b807fd2c7d.

src/NodaTime/Xml/XmlSchemaDefinition.cs
0xced: You're absolutely right, we don't need reflection at all. I have removed reflection in 6c06ed466ce2770bc44e2b7d4e0199b807fd2c7d.

src/NodaTime/Xml/XmlSchemaDefinition.cs
0xced: Done in 4b690b0312668f78e728ff0dc712caea1588f16a.

src/NodaTime/Xml/XmlSchemaDefinition.cs
0xced: Done in 4b690b0312668f78e728ff0dc712caea1588f16a.

src/NodaTime/Xml/XmlSchemaDefinition.cs
0xced: This is not relevant anymore after 6c06ed466ce2770bc44e2b7d4e0199b807fd2c7d but I'll remember to always use braces, even for single-line if statements.

jskeet:
Thanks - I'll try to have another look either later today or early tomorrow.

src/NodaTime.Demo/XmlSchemaDemo.NetCore.cs
0xced: I have rewritten this with a partial XmlSchemaDemo class + private nested support classes to make it clear that those classes are only for the XML Schema demo.

src/NodaTime.Demo/NodaTime.Demo.csproj
0xced: I have removed the conditionality in the csproj but the new code (bd89466952f2a308830842bc91b721bbb0ee56c0) is still capable of handling both .NET Framework and .NET Core. Unlike `NodaTime.Test`, the `NodaTime.Demo` project doesn't exercises corner cases of NodaTime and it's not a problem to multi targe .NET Framework and .NET Core. I have tested: running NodaTime.Demo on both .NET Framework and .NET Core (with `<TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>`) works fine. 

In this particular test (`XmlSchemaDemo`) I think it's interesting to keep the `XmlCodeExporterAssemblyCreator` class and its related `NodaTimeSchemaImporterExtension` that demonstrates how to intercept the NodaTime XML types and convert them to their equivalent C# types.

The interesting parts of this demo are actually the `NodaTimeSchemaImporterExtension` and `NodaTimeNamingProvider` classes.

src/NodaTime.Demo/XmlSchemaDemo.XmlCodeExporterAssemblyCreator.cs
```diff
@@ -0,0 +1,111 @@
+using System;
```
jskeet, line 1: Could you add copyright statements to all files please? I can do it afterwards if not.

0xced: Done in 2080028524c3383d9c6cbce852c32bd440ceed22.

src/NodaTime.Test/XmlSchemaTest.XmlSchema.approved.xml
0xced: I implemented what I think is an elegant solution in d334f28d0a36794a14afa3752754bbeb0d75c39b, using an XML schema union between the enumeration of known zone ids (for documentation purpose) and a string (for validation of future, not yet known time zone ids).

0xced:
> Other than the one nit, I think we're just down to the matter of time zone IDs - any further thoughts on that?

About the time zone IDs, I found a solution and implemented it in d334f28d0a36794a14afa3752754bbeb0d75c39b, see https://github.com/nodatime/nodatime/pull/1500#discussion_r407729967

And just to make sure, what exact *nit* are you talking about?

jskeet:
The nit was the copyright, which you've done. So I'm hoping when I look tomorrow I can merge this.

src/NodaTime.Demo/NodaTime.Demo.csproj
jskeet: I'll need to check whether the changes to the demo project file break snippet generation (which has previously been quite brittle with respect to getting Roslyn to load it properly). But I can do that after merging.

src/NodaTime.Test/Xml/XmlSchemaTest.XmlSchema.approved.xml
```diff
@@ -110,6 +110,9 @@
     </xs:restriction>
   </xs:simpleType>
   <xs:simpleType name="zone">
+    <xs:union memberTypes="nodatime:zoneIds xs:string" />
+  </xs:simpleType>
+  <xs:simpleType name="zoneIds">
```
jskeet, line 115: It's just occurred to me that changes in the set of zone IDs is presumably going to mean this file needs to be rewritten. I'd want to automate that. Please could you provide instructions for how you generated this file, so I can add that to the time zone update script?

0xced: Here's how to generate this file:
```csharp
var settings = new XmlWriterSettings { Indent = true, NewLineChars = "\n", Encoding = new UTF8Encoding() };
using var xmlWriter = XmlWriter.Create(@"src\NodaTime.Test\Xml\XmlSchemaTest.XmlSchema.approved.xml", settings);
NodaTimeXmlSchema.Write(xmlWriter);
```
(Adapted from [NodaTime.Test.Xml.XmlSchemaTest.GetXmlSchemaString](https://github.com/nodatime/nodatime/blob/92e81ea52bfdb8088524515a3a0f37e7c274209c/src/NodaTime.Test/Xml/XmlSchemaTest.cs#L301-L309))

Also, it might be nice to host it on https://nodatime.org/api/NodaTime.xsd for example.

jskeet: Thanks. I've created a new issue to track the need for that.

jskeet: (The regeneration, at least. I'm not entirely convinced about the hosting yet, especially given that it depends on the time zone data in use.)

PR #1501, opened 2020-04-07
Enable .NET Framework 4.8 on all unit test projects

0xced:
Probably not yet ready to be merged, let's see what happens on AppVeyor and Travis…

### DIFF
--- a/src/NodaTime.TzdbCompiler/NodaTime.TzdbCompiler.csproj
+++ b/src/NodaTime.TzdbCompiler/NodaTime.TzdbCompiler.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
     <OutputType>Exe</OutputType>
     <AssemblyOriginatorKeyFile>../../NodaTime Release.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,16 @@
 # See http://docs.travis-ci.com/user/languages/csharp/ for details
 
-# TODO: Test against Mono as well, using matrix.include
 language: csharp
-mono: none
 dotnet: 3.1
-dist: xenial
+
+matrix:
+  include:
+    - name: ".NET Core"
+      mono: none
+      env: TARGET_FRAMEWORK=netcoreapp3.1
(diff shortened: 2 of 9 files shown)


jskeet:
Could you add to the PR what benefits you believe this will bring, bearing in mind the additional complexity? As I said before, I'd really rather avoid this if possible. It's lovely being down to a single target framework almost everywhere - this feels like a step backwards in terms of maintenance.  If there's a compelling reason, then sure - but it's got to be pretty convincing.

src/NodaTime.TzdbCompiler/NodaTime.TzdbCompiler.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
jskeet, line 4: TzdbCompiler is an application, not a library. This change at least should be reverted (or at least changed).

0xced:
The point of this pull request is to enable .NET Framework 4.8 **on unit test projects**. So that NodaTime *tests* are run on both .NET Core and .NET Framework (Windows on AppVeyor + Mono on Travis).

If you think testing on .NET Framework is not desired I'll close this pull request right away. Configuring `.travis.yml` is no fun, I already tried a dozen combination then force pushed and the build job isn't even created on Travis. 😠

src/NodaTime.TzdbCompiler/NodaTime.TzdbCompiler.csproj
0xced: Absolutely. I thought you'd get compilation error when trying to compile an exe targeting `netstandard2.0`, but apparently you don't!

0xced:
Also, there was this TODO in the `.travis.yml` file so I thought testing on Mono was something that you may want implemented.
```yml
# TODO: Test against Mono as well, using matrix.include
```

jskeet:
Given the way things are, I think it would be best not to build a rod for our own back like this. As you say, getting Travis to work like this can be fiddly - and I'm sure we'll then run into Mono bugs (already documented) that we'd need to skip tests for, etc.

If you want to create a new PR to remove the TODO (which I suspect is very old indeed) I'll happily merge that :)

jskeet:
(If I actually thought this would find bugs, I'd be more inclined to take the hit - but my experience is that the NodaTime code base just doesn't run into that sort of bug.)

0xced:
> we'll then run into Mono bugs (already documented)

And that's exactly what happened. 😒 Closing this pull request.

PR #1509, opened 2020-04-19
Extract conversion settings from DateTimeZoneProviders

### LINKED ISSUE #1505: Move DateTimeZoneProviders.Serialization
*Possibly* leave the old one in place in deprecated form.

jskeet:
Fixes #1505.

I've left a deprecated property for DateTimeZoneProviders.Serialization.
While moving to 3.0 means we *can* break anything, we actually
restricted the breaking changes to "removing binary serialization"
(and removing incorrectly named method that's been deprecated since
2.1 was released in 2017).

The ForTypeConverter property is new in 3.0, so doesn't need
deprecation. I'll add it to release notes for any existing beta
users, but it should be okay to move. (I'll release a new beta
shortly after this goes in.)

cc @0xced

### DIFF
--- a/src/NodaTime/DateTimeZoneProviders.cs
+++ b/src/NodaTime/DateTimeZoneProviders.cs
@@ -4,6 +4,7 @@
 
 using NodaTime.TimeZones;
 using NodaTime.Utility;
+using System;
 
 namespace NodaTime
 {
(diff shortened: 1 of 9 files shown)


src/NodaTime/DateTimeZoneProviders.cs
```diff
@@ -49,71 +50,17 @@ private static class BclHolder
...
-                {
-                    serializationProvider = Preconditions.CheckNotNull(value, nameof(value));
-                }
-            }
-        }
-
-        /// <summary>
-        /// Gets the <see cref="IDateTimeZoneProvider"/> to use to interpret a time zone ID read as part of
-        /// conversion using the default <see cref="System.ComponentModel.TypeConverter"/> for <see cref="ZonedDateTime"/>.
-        /// Note that if a value other than <see cref="DateTimeZoneProviders.Tzdb"/> is required, it should be set on
```
malcolmr, line 91: Any reason to drop some of this text from the new location?

src/NodaTime/Text/TypeConverterSettings.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// own type converters to apply if they wish to.
+    /// </summary>
+    public static class TypeConverterSettings
+    {
+        private static readonly object stateLock = new object();
+
+        private static IDateTimeZoneProvider dateTimeZoneProvider = DateTimeZoneProviders.Tzdb;
+
+        /// <summary>
+        /// Gets the <see cref="IDateTimeZoneProvider"/> to use to interpret a time zone ID read as part of
```
malcolmr, line 21: It's also used for serialisation, not just deserialisation. Not sure if we should be clearer about that here?

(Same goes for `XmlSerializationSettings`.)

src/NodaTime/DateTimeZoneProviders.cs
jskeet: Nope - restored now.

src/NodaTime/Text/TypeConverterSettings.cs
jskeet: I don't think it really is used for serialization - there's no need to look anything up. We just serialize the ID of the zone, regardless of which provider it came from.

jskeet:
(Will merge when CI is green. We can always tweak the docs later.)

src/NodaTime/Text/TypeConverterSettings.cs
malcolmr: Ah, that's a good point. Thanks!

PR #1526, opened 2020-05-02
Prep for handling issue 1524

jskeet:


src/NodaTime.Test/Ignore.cs
```diff
@@ -20,6 +18,16 @@ public static void When(bool condition, string message)
...
+        }
+
+        /// <summary>
+        /// Calls Assert.Ignore, which always throws an exception - but with a return type
+        /// of T, which makes it easier to call in expression-bodied members.
+        /// </summary>
+        public static T Throw<T>(string message)
+        {
+            Assert.Ignore(message);
+            throw new InvalidOperationException($"Expected Assert.Ignore to throw");
```
malcolmr, line 30: Any reason to use an interpolated string here?

jskeet: Nope. Will fix it up when I look at the FIXME I introduced.

PR #1527, opened 2020-05-02
Make BclDateTimeZone behave more reasonably for midnight transitions

### LINKED ISSUE #1524: TimeZone.AtStartofDate for day with switch to DLS in Brazil
In my code I have the following method:

```csharp
public DateTime GetUtcStartOfDay(DateTime day) {
 return TimeZone.AtStartOfDay(new NodaTime.LocalDate(day.Year, day.Month, day.Day)).ToDateTimeUtc();
}
```

for 2012/10/21 I would assume that it returns UTC 2012-10-21-03:00 as the time is advanced in Brazil 1 hour but nothing happened yet with respect to UTC.<br/>
Then for the following day 2012/10/22 I would assume to get UTC 2012-10-22-02:00 as the missing hour leads to the shorter day 21.<br/>
However for 2012/10/21 I get instead: 2012-10-21-02:59:59.999,  which is 1 ms less then expected.
This millisecond is already missing iin the AtStartOfDay result, so I assume something is wrong here?
For the following day 22 I get the full hour as expected.<br/>
Same applies for all other years for the day where the 1 hour is advanced.<br/>
BTW I use BclDateTimeZone.

jskeet:
Transitions that would be represented as 24:00 in the IANA time zone database are represented as 23:59:59.999 in the Windows database. It makes sense to correct this as we have the facility to do so; the intent of the data is clearly "one millisecond later" even though that's not what the Windows rule actually says. (I don't think anyone would argue that the transition actually happens when the Windows rule says it does.)

Fixes #1524.

### DIFF
--- a/src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
+++ b/src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
@@ -2,11 +2,12 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+using NodaTime.TimeZones;
+using NUnit.Framework;
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
-using NUnit.Framework;
-using NodaTime.TimeZones;
 
 namespace NodaTime.Test.TimeZones
 {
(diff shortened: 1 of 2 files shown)


src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
```diff
@@ -267,6 +268,50 @@ public void FakeDaylightSavingTime()
...
+        public void TransitionAtMidnight()
+        {
+            var bclZone = GetBclZoneOrIgnore("E. South America Standard Time");
+            var nodaTzdbZone = DateTimeZoneProviders.Tzdb["America/Sao_Paulo"];
+            var nodaBclZone = BclDateTimeZone.FromTimeZoneInfo(bclZone);
+
+            var expectedFallBack = Instant.FromUtc(2012, 2, 26, 2, 0, 0);
+            var expectedSpringForward = Instant.FromUtc(2012, 10, 21, 3, 0, 0);
+
+            // We assume that the fall-back happens before the spring-forward in the dates
```
malcolmr, line 285: Unfinished comment? (Also, do we have to assume anything? This test is for a concrete situation.)

src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
```diff
@@ -267,6 +268,50 @@ public void FakeDaylightSavingTime()
...
+            var betweenTransitions = expectedFallBack + Duration.FromDays(20);
+
+            // Check that these transitions are as expected when we use TZDB.
+            var nodaTzdbInterval = nodaTzdbZone.GetZoneInterval(betweenTransitions);
+            Assert.AreEqual(expectedFallBack, nodaTzdbInterval.Start);
+            Assert.AreEqual(expectedSpringForward, nodaTzdbInterval.End);
+
+            // Check that the real BCL time zone behaves as reported in the issue: the transitions occur one millisecond early
+            var rules = bclZone!.GetAdjustmentRules();
```
malcolmr, line 294: This isn't really related to this change, but I'm trying to understand why the null-forgiving operator is needed? From what I can see, the only useful use of this would be if we have an explicitly-nullable reference type (`TimeZoneInfo?`), but we don't?

(I'm looking at  https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-annotation-context for reference.)

(Likewise a little bit below.)

src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
jskeet: Thanks, will address.

src/NodaTime.Test/TimeZones/BclDateTimeZoneTest.cs
jskeet: Are - it's not now. It was in an earlier version. The local variable isn't needed either! (Was just useful when debugging)

jskeet:
Fixed the little problems; will merge on green.

PR #1530, opened 2020-05-08
Add equality operators for Period and ZoneInterval

### LINKED ISSUE #1529: Period does not implement operator ==
I happened to create two zero-length periods two different ways to test for equality like so:
`var period1 = Period.Zero;`
`var period2 = PeriodPattern.Roundtrip.Parse("P0D").Value;`

Doing `period1 == period2` compiles just fine. However, the result was not what I expected. This returns `false`, not `true`. On the other hand, doing `period1.Equals(period2)` does return the expected result of `true`.

jskeet:
Additionally, add a check to ensure we don't miss this again, at
least for anything implementing `IEquatable<T>`

Fixes #1529.

### DIFF
--- a/src/NodaTime.Test/ConsistencyTest.cs
+++ b/src/NodaTime.Test/ConsistencyTest.cs
@@ -1,0 +1,37 @@
+﻿// Copyright 2020 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using NodaTime.TimeZones.Cldr;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NodaTime.Test
+{
+    /// <summary>
+    /// Tests using reflection to check that types are consistent.
+    /// </summary>
+    public class ConsistencyTest
+    {
+        private static IEnumerable<Type> AllPublicTypes = typeof(LocalDate).Assembly
+            .GetTypes()
+            .Where(t => t.IsPublic)
+            .ToList();
+
+        [Test]
+        public void EqualityOperatorForEquatable()
+        {
+            var expectedExceptions = new[] { typeof(MapZone) };
+
+            var failures = AllPublicTypes
+                .Where(t => typeof(IEquatable<>).MakeGenericType(t).IsAssignableFrom(t))
+                .Where(t => t.GetMethod("op_Equality") is null)
+                .Except(expectedExceptions)
+                .ToList();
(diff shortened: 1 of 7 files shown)


src/NodaTime.Test/ConsistencyTest.cs
```diff
@@ -0,0 +1,37 @@
...
+            .ToList();
+
+        [Test]
+        public void EqualityOperatorForEquatable()
+        {
+            var expectedExceptions = new[] { typeof(MapZone) };
+
+            var failures = AllPublicTypes
+                .Where(t => typeof(IEquatable<>).MakeGenericType(t).IsAssignableFrom(t))
+                .Where(t => t.GetMethod("op_Equality") is null)
```
malcolmr, line 31: Presumably reasonable to add a similar check here for `op_Inequality` as well?

src/NodaTime.Test/ConsistencyTest.cs
```diff
@@ -0,0 +1,37 @@
...
+                .Except(expectedExceptions)
+                .ToList();
+            Assert.IsEmpty(failures);
+        }
+    }
```
malcolmr, line 36: At the risk of being somewhat pedantic, how about an additional test for `IComparable` types having all of {`<`, `<=`, `>=`, `>`} too?

src/NodaTime.Test/ConsistencyTest.cs
jskeet: No need - the C# language won't let you overload `==` without overloading `!=` too.

src/NodaTime.Test/ConsistencyTest.cs
jskeet: Hmm. I think we're likely to need a lot more exceptions for this, as it's quite reasonable to make things comparable without overloading those operators (e.g. string) but I'll try adding the test anyway - if we only need a few exceptions, it wouldn't hurt to have a reminder.

jskeet:
It turns out this *was* worth doing - I've added operators for YearMonth. PTAL.
(I've also noticed a possible performance issue in comparisons for both YearMonth and LocalDate - will benchmark that now.)

jskeet:
Just realized I need to add tests for the operators in YearMonth - that should be a matter of changing which helper method we call though.

src/NodaTime.Test/ConsistencyTest.cs
malcolmr: Hurrah for common sense! (Another option would have been to generate synthetic operators given `==` and `<`, but this is good too.)

PR #1532, opened 2020-05-09
Improve performance of LocalDate and YearMonth comparisons

jskeet:
Only the last commit is relevant here - the first two are from #1530.

### DIFF
--- a/src/NodaTime.Benchmarks/NodaTimeTests/LocalDateBenchmarks.cs
+++ b/src/NodaTime.Benchmarks/NodaTimeTests/LocalDateBenchmarks.cs
@@ -129,5 +129,11 @@ namespace NodaTime.Benchmarks.NodaTimeTests
 
         [Benchmark]
         public LocalDate FromDateTime_WithCalendar() => LocalDate.FromDateTime(SampleDateTime, CalendarSystem.Julian);
+        
+        [Benchmark]
+        public int CompareTo() => Sample.CompareTo(SampleBeforeEpoch);
+
+        [Benchmark]
+        public bool LessThanOperator() => Sample < SampleBeforeEpoch;
     }
 }
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -140,7 +140,9 @@ namespace NodaTime
 
         /// <summary>Gets the calendar system associated with this local date.</summary>
         /// <value>The calendar system associated with this local date.</value>
-        public CalendarSystem Calendar => CalendarSystem.ForOrdinal(yearMonthDayCalendar.CalendarOrdinal);
+        public CalendarSystem Calendar => CalendarSystem.ForOrdinal(CalendarOrdinal);
+
+        private CalendarOrdinal CalendarOrdinal => yearMonthDayCalendar.CalendarOrdinal;
 
         /// <summary>Gets the year of this local date.</summary>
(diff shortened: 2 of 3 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -140,7 +140,9 @@ public LocalDate(Era era, int yearOfEra, int month, int day, CalendarSystem cale
 
         /// <summary>Gets the calendar system associated with this local date.</summary>
         /// <value>The calendar system associated with this local date.</value>
-        public CalendarSystem Calendar => CalendarSystem.ForOrdinal(yearMonthDayCalendar.CalendarOrdinal);
+        public CalendarSystem Calendar => CalendarSystem.ForOrdinal(CalendarOrdinal);
+
+        private CalendarOrdinal CalendarOrdinal => yearMonthDayCalendar.CalendarOrdinal;
```
malcolmr, line 145: (I wondered if this maybe could do with a comment, though initially I misread this as a field rather than a property.)

src/NodaTime/LocalDate.cs
```diff
@@ -506,10 +508,16 @@ public static LocalDate FromYearMonthWeekAndDay(int year, int month, int occurre
...
-            return Calendar.Compare(YearMonthDay, other.YearMonthDay);
+            Preconditions.CheckArgument(CalendarOrdinal == other.CalendarOrdinal, nameof(other), "Only values with the same calendar system can be compared");
+            return TrustedCompareTo(other);
         }
 
+        /// <summary>
+        /// Performs a comparison with another date, trusting that the calendar of the other date is already correct.
+        /// This avoids duplicate calendar checks.
+        /// </summary>
+        private int TrustedCompareTo([Trusted] LocalDate other) => Calendar.Compare(YearMonthDay, other.YearMonthDay);
```
malcolmr, line 519: (Why is `Compare()` an instance method on `CalendarSystem`? Shouldn't it be static?)

src/NodaTime.Benchmarks/NodaTimeTests/LocalDateBenchmarks.cs
```diff
@@ -18,7 +18,7 @@ public class LocalDateBenchmarks
         private static readonly LocalDatePattern Pattern = LocalDatePattern.CreateWithInvariantCulture("dd/MM/uuuu");
 
         private static readonly Period SamplePeriod = new PeriodBuilder { Years = 1, Months = 2, Weeks = 3, Days = 4 }.Build();
-
+
```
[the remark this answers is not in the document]
jskeet: Oh yes. Not sure why. Will fix.

src/NodaTime/LocalDate.cs
jskeet: I think it's probably okay as it is. The comment would probably be a pretty no-op one. (And it's private, so it's not like we need to worry about code in other types getting confused.)

src/NodaTime/LocalDate.cs
jskeet: Nope, it depends on the calendar system. The Hebrew scriptural calendar system prevents the straightforward comparison you might expect.

malcolmr: Ah, sorry, too much Java lately: I misread `return YearMonthDayCalculator.Compare()` in the implementation as a static method call on `YearMonthDayCalculator`, not a call through the public member property (that's also called `YearMonthDayCalculator`).

PR #1549, opened 2020-05-22
Fix a typo in Interval documentation

malcolmr:
1fb657c fixed up the documentation for various types' equality/ordering, but accidentally uses "date interval" for both `DateInterval` and `Interval`.


malcolmr:
(Possibly worth backporting this to 3.0.x, since it's so trivial.)

Issue #1536, opened 2020-05-19
Convenience method YearMonth.PlusMonths(x)?


ArneMancofi:
In version 3, if I want to find the `YearMonth` in two months' time, I know I can do
```csharp
new YearMonth(2020, 05).OnDayOfMonth(1).PlusMonths(2).ToYearMonth()
```
but that seems rather long-winded. Does a convenience method like
```csharp
new YearMonth(2020, 05).PlusMonths(2)
```
exist anywhere in Noda Time?

jskeet:
No, this doesn't exist yet, but sounds reasonable. I don't expect to get to it before 3.0 though - you can always write your own extension method for now.

carlosschults:
Can I give this a try, @jskeet ?

jskeet:
Sure, that's fine :)

carlosschults:
@jskeet OK, I need a little bit of help here. I'm just trying to run the tests—with `dotnet test`—before doing any development, and I'm getting some failed tests.

The first failing test is "AllDetectedNamesAreMaped":

`    System.TypeInitializationException : The type initializer for 'NodaTime.TzdbCompiler.Test.Tzdb.NameIdMappingSupportTest' threw an exception.
      ----> System.ArgumentException : An item with the same key has already been added. Key: Hora Padrão Austrália Ocidental
  Rastreamento de Pilha: 
    NameIdMappingSupportTest.AllDetectedNamesAreMapped() linha 34
    --ArgumentException
    Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
    Dictionary`2.Add(TKey key, TValue value)
    Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
    Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
    NameIdMappingSupportTest.cctor() linha 16`

The second one is "AllDetectedNamesAreMappedCorrectly":

`System.TypeInitializationException : The type initializer for 'NodaTime.TzdbCompiler.Test.Tzdb.NameIdMappingSupportTest' threw an exception.
      ----> System.ArgumentException : An item with the same key has already been added. Key: Hora Padrão Austrália Ocidental
  Rastreamento de Pilha: 
[...]

PR #1550, opened 2020-05-25
Create method 'PlusMonths()'

carlosschults:
I've added a new method to the YearMonth structure called 'PlusMonths', which is a shortcut to obtaining a new YearMonth object representing the current one plus a number of months.

Addresses #1536


jskeet:
I'll look at this when I can, but it may be next weekend - this week is very busy for me.

jskeet:
(Sorry for not getting to this earlier.)

Issue #1450, opened 2019-12-16
Expose efficient single-unit difference methods


jskeet:
See https://stackoverflow.com/questions/58288463 for details.

Something like `Period.DaysBetween(LocalDate, LocalDate)` etc.

jskeet:
Fixed by #1557, at least for days. That's probably the most common use case (I needed it myself recently...) so I think we can close this issue for now, and follow a similar pattern if there are requests for other units.

PR #1557, opened 2020-06-17
Add 'DaysBetween' method to Period

carlosschults:
An attempt at addressing #1450.


jskeet:
Thanks for this - I may well not find time to look at it before the weekend, but thanks in advance :)

PR #1577, opened 2020-10-10
Added more LocalDate demo snippets

toadfish2:
I am happy to do more snippets, but I just wanted to ensure I was heading in the correct direction.

### DIFF
--- a/src/NodaTime.Demo/LocalDateDemo.cs
+++ b/src/NodaTime.Demo/LocalDateDemo.cs
@@ -121,7 +121,7 @@ namespace NodaTime.Demo
             DateTime earlyJune = new DateTime(2010, 6, 5);
             CalendarSystem calendar = CalendarSystem.ForId("Julian");
             LocalDate date = Snippet.For(LocalDate.FromDateTime(earlyJune, calendar));
-            //Until 2099, Julian calendar is 13 days behind Gregorian
+            // Between the years 2000 and 2099, the Julian calendar is 13 days behind the Gregorian calendar.
             Assert.AreEqual(2010, date.Year);
             Assert.AreEqual(5, date.Month);
             Assert.AreEqual(23, date.Day);
--- a/src/NodaTime.Demo/LocalDateDemo.cs
+++ b/src/NodaTime.Demo/LocalDateDemo.cs
@@ -128,13 +128,6 @@ namespace NodaTime.Demo
         }
 
         [Test]
-        public void CommonEra()
-        {
-            LocalDate date = Snippet.For(new LocalDate(Era.Common, 2010, 6, 16));
-            Assert.AreEqual(new LocalDate(2010, 6, 16), date);
-        }
-
-        [Test]
         public void BeforeCommonEra()
         {
             LocalDate date = Snippet.For(new LocalDate(Era.BeforeCommon, 2010, 6, 16));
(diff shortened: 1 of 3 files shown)


src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -104,5 +106,82 @@ public void Min()
...
+            Assert.AreEqual(new LocalDate(2010, 6, 5), date);
+        }
+
+        [Test]
+        public void FromDateTimeWithCalendarSystem()
+        {
+            DateTime earlyJune = new DateTime(2010, 6, 5);
+            CalendarSystem calendar = CalendarSystem.ForId("Julian");
+            LocalDate date = Snippet.For(LocalDate.FromDateTime(earlyJune, calendar));
+            //Until 2099, Julian calendar is 13 days behind Gregorian
```
jskeet, line 124:
```suggestion
            // Between the years 2000 and 2099, the Julian calendar is 13 days behind the Gregorian calendar.
```

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -104,5 +106,82 @@ public void Min()
...
+            Assert.AreEqual(2010, date.Year);
+            Assert.AreEqual(5, date.Month);
+            Assert.AreEqual(23, date.Day);
+        }
+
+        [Test]
+        public void CommonEra()
+        {
+            LocalDate date = Snippet.For(new LocalDate(Era.Common, 2010, 6, 16));
```
jskeet, line 133: Currently we only support one snippet per member (the constructor accepting an Era in this case).
I suspect the BeforeCommonEra would be a better example in this case, so it's worth removing this one.

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -104,5 +106,82 @@ public void Min()
...
+        {
+            LocalDate date = Snippet.For(new LocalDate(Era.BeforeCommon, 2010, 6, 16));
+            Assert.AreEqual(new LocalDate(Era.BeforeCommon, 2010, 6, 16), date);
+        }
+
+        [Test]
+        public void FromWeekYearWeekAndDay()
+        {
+            LocalDate date = Snippet.For(LocalDate.FromWeekYearWeekAndDay(2010,24,IsoDayOfWeek.Wednesday));
+            Assert.AreEqual(new LocalDate(2010, 6,16), date);
```
jskeet, line 148:
```suggestion
            LocalDate date = Snippet.For(LocalDate.FromWeekYearWeekAndDay(2010, 24, IsoDayOfWeek.Wednesday));
            Assert.AreEqual(new LocalDate(2010, 6, 16), date);
```

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -104,5 +106,82 @@ public void Min()
...
+        {
+            LocalDate date = Snippet.For(LocalDate.FromYearMonthWeekAndDay(2010, 6, 3, IsoDayOfWeek.Wednesday));
+            Assert.AreEqual(new LocalDate(2010, 6, 16), date);
+        }
+
+        [Test]
+        public void ToYearMonth()
+        {
+            YearMonth yearMonth = Snippet.For(new LocalDate(2010, 6, 16).ToYearMonth());
+            Assert.AreEqual(new YearMonth(2010,6), yearMonth);
```
jskeet, line 162:
```suggestion
            Assert.AreEqual(new YearMonth(2010, 6), yearMonth);
```

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -104,5 +106,82 @@ public void Min()
...
+        }
+
+        [Test]
+        public void Plus()
+        {
+            LocalDate date = Snippet.For(new LocalDate(2010, 6, 16).Plus(Period.FromDays(1)));
```
jskeet, line 168: This is okay as-is, but it *might* make sense to give an example where it's not as obvious, e.g. adding a month to January 30th. I don't much mind though.

src/NodaTime.Demo/LocalDateDemo.cs
```diff
@@ -104,5 +106,82 @@ public void Min()
...
+            Assert.AreEqual(new LocalDate(2010, 6, 17), date);
+        }
+
+        [Test]
+        public void Subtract()
+        {
+            LocalDate date = Snippet.For(LocalDate.Subtract(new LocalDate(2010, 6, 16), Period.FromDays(1)));
```
jskeet, line 175: I would suggest subtracting from "the result of the Plus example" - in particular, if you add a month to January 30th 2010, you get February 28th 2010... and subtracting a month from that *doesn't* get you back to January 30th...

jskeet (commented):
A few suggestions, but I think we'll be able to get this merged pretty soon - thanks so much!

src/NodaTime.Demo/LocalDateDemo.cs
toadfish2: I thought that might be the case but you did state you were looking at some enhancements for the snippet handling!

jskeet (approved):
Thanks - will merge now, squashing it all down into one commit.

PR #1584, opened 2020-10-31
Attempt to workaround analyzer detection of CA2119.

jskeet:
I don't believe this is actually a vulnerability, but it can be
worked around just by separating the internal interface method on
IZoneIntervalMap from the abstract method in DateTimeZone.


jskeet:
Closing as we really don't want to do this.

PR #1594, opened 2020-11-26
Add FxCop and fix the errors it produces

jskeet:
This PR looks huge, but each commit is relatively small. Definitely review one commit at a time!

I've deliberately *not* suppressed anything at a project level. Each suppression should be very explicit, so we don't get into bad habits. (There are some things that we might have done differently if we'd seen them from the start, but would now break things.)

This PR should be almost entirely invisible to users, with two exceptions:

- There are two static methods in LocalDate and LocalDateTime that will now (correctly) throw ArgumentNullException instead of NullReferenceException when null is passed to them
- There are two additional static methods in Period to match existing operators

### DIFF
--- a/src/NodaTime.Test/TimeZones/IO/IoStream.cs
+++ b/src/NodaTime.Test/TimeZones/IO/IoStream.cs
@@ -161,9 +161,9 @@ namespace NodaTime.Test.TimeZones.IO
             /// <returns>
             ///   A long value representing the length of the stream in bytes.
             /// </returns>
-            /// <exception cref="T:System.NotSupportedException">A class derived from Stream does not support seeking. 
+            /// <exception cref="System.NotSupportedException">A class derived from Stream does not support seeking. 
             /// </exception>
-            /// <exception cref="T:System.ObjectDisposedException">Methods were called after the stream was closed. 
+            /// <exception cref="System.ObjectDisposedException">Methods were called after the stream was closed. 
             /// </exception>
             /// <filterpriority>1</filterpriority>
             public override long Length
@@ -177,11 +177,11 @@ namespace NodaTime.Test.TimeZones.IO
             /// <returns>
             ///   The current position within the stream.
             /// </returns>
-            /// <exception cref="T:System.IO.IOException">An I/O error occurs. 
+            /// <exception cref="System.IO.IOException">An I/O error occurs. 
             /// </exception>
-            /// <exception cref="T:System.NotSupportedException">The stream does not support seeking. 
+            /// <exception cref="System.NotSupportedException">The stream does not support seeking. 
             /// </exception>
-            /// <exception cref="T:System.ObjectDisposedException">Methods were called after the stream was closed. 
+            /// <exception cref="System.ObjectDisposedException">Methods were called after the stream was closed. 
             /// </exception>
             /// <filterpriority>1</filterpriority>
             public override long Position
@@ -193,7 +193,7 @@ namespace NodaTime.Test.TimeZones.IO
             /// <summary>
             ///   When overridden in a derived class, clears all buffers for this stream and causes any buffered data to be written to the underlying device.
             /// </summary>
-            /// <exception cref="T:System.IO.IOException">An I/O error occurs. 
+            /// <exception cref="System.IO.IOException">An I/O error occurs. 
             /// </exception>
             /// <filterpriority>2</filterpriority>
             public override void Flush()
@@ -212,17 +212,17 @@ namespace NodaTime.Test.TimeZones.IO
             /// </param>
             /// <param name="count">The maximum number of bytes to be read from the current stream. 
             /// </param>
-            /// <exception cref="T:System.ArgumentException">The sum of <paramref name = "offset" /> and <paramref name = "count" /> is larger than the buffer length. 
+            /// <exception cref="System.ArgumentException">The sum of <paramref name = "offset" /> and <paramref name = "count" /> is larger than the buffer length. 
             /// </exception>
-            /// <exception cref="T:System.ArgumentNullException"><paramref name = "buffer" /> is null. 
+            /// <exception cref="System.ArgumentNullException"><paramref name = "buffer" /> is null. 
             /// </exception>
-            /// <exception cref="T:System.ArgumentOutOfRangeException"><paramref name = "offset" /> or <paramref name = "count" /> is negative. 
+            /// <exception cref="System.ArgumentOutOfRangeException"><paramref name = "offset" /> or <paramref name = "count" /> is negative. 
             /// </exception>
-            /// <exception cref="T:System.IO.IOException">An I/O error occurs. 
+            /// <exception cref="System.IO.IOException">An I/O error occurs. 
             /// </exception>
-            /// <exception cref="T:System.NotSupportedException">The stream does not support reading. 
+            /// <exception cref="System.NotSupportedException">The stream does not support reading. 
             /// </exception>
-            /// <exception cref="T:System.ObjectDisposedException">Methods were called after the stream was closed. 
+            /// <exception cref="System.ObjectDisposedException">Methods were called after the stream was closed. 
             /// </exception>
             /// <filterpriority>1</filterpriority>
             public override int Read(byte[] buffer, int offset, int count)
@@ -243,13 +243,13 @@ namespace NodaTime.Test.TimeZones.IO
             /// </returns>
             /// <param name="offset">A byte offset relative to the <paramref name = "origin" /> parameter. 
             /// </param>
-            /// <param name="origin">A value of type <see cref="T:System.IO.SeekOrigin" /> indicating the reference point used to obtain the new position. 
+            /// <param name="origin">A value of type <see cref="System.IO.SeekOrigin" /> indicating the reference point used to obtain the new position. 
             /// </param>
-            /// <exception cref="T:System.IO.IOException">An I/O error occurs. 
+            /// <exception cref="System.IO.IOException">An I/O error occurs. 
             /// </exception>
-            /// <exception cref="T:System.NotSupportedException">The stream does not support seeking, such as if the stream is constructed from a pipe or console output. 
+            /// <exception cref="System.NotSupportedException">The stream does not support seeking, such as if the stream is constructed from a pipe or console output. 
             /// </exception>
-            /// <exception cref="T:System.ObjectDisposedException">Methods were called after the stream was closed. 
+            /// <exception cref="System.ObjectDisposedException">Methods were called after the stream was closed. 
             /// </exception>
             /// <filterpriority>1</filterpriority>
             public override long Seek(long offset, SeekOrigin origin)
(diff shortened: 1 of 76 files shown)


src/NodaTime.Test/TimeZones/IO/IoStream.cs
```diff
@@ -161,9 +161,9 @@ public ReadStreamImpl(IoStream stream)
             /// <returns>
             ///   A long value representing the length of the stream in bytes.
             /// </returns>
-            /// <exception cref="T:System.NotSupportedException">A class derived from Stream does not support seeking.
```
malcolmr, line 164: I think these may have been needed at some point by Sandcastle (SHFB?), but if the output documentation still works, then lgtm.

src/NodaTime/Text/UnparsableValueException.cs
```diff
@@ -31,5 +31,14 @@ public UnparsableValueException(string message)
             : base(message)
         {
         }
+
+        /// <summary>
+        /// Creates a new UnparsableValueException with the given message and inner exception
+        /// </summary>
+        /// <param name="message">The failure message</param>
+        /// <param name="innerException">The inner exception</param>
+        public UnparsableValueException(string message, Exception innerException) : base(message, innerException)
```
malcolmr, line 40: okay but why? In case someone else wants to use it in a different context, externally?

src/NodaTime/CalendarSystem.cs
```diff
@@ -11,6 +11,11 @@
 using System.Diagnostics.CodeAnalysis;
 using static System.FormattableString;
 
+// Remove static constructors.
+// The calendar system initialization here can be slow or increase memory usage,
+// so the static constructors are very deliberately present to make them truly lazy.
```
malcolmr, line 16: Didn't we have an initialisation-order issue around this area that required them to be lazy? (or was that something else?)

src/NodaTime/Calendars/GJYearMonthDayCalculator.cs
```diff
@@ -15,22 +15,20 @@ internal abstract class GJYearMonthDayCalculator : RegularYearMonthDayCalculator
...
-        private static readonly int[] MinTotalDaysByMonth;
-        private static readonly int[] MaxTotalDaysByMonth;
+        // Note: these fields must be declared after MinDaysPerMonth and MaxDaysPerMonth so that the initialization
+        // is correct. This behavior (textual order for initialization) is guaranteed by the spec. We'd normally
+        // try to avoid relying on it, but that's quite hard here.
+        private static readonly int[] MinTotalDaysByMonth = GenerateTotalDaysByMonth(MinDaysPerMonth);
+        private static readonly int[] MaxTotalDaysByMonth = GenerateTotalDaysByMonth(MaxDaysPerMonth);
 
-        static GJYearMonthDayCalculator()
+        private static int[] GenerateTotalDaysByMonth(int[] monthLengths)
```
malcolmr, line 24: Add trivial documentation? If I read correctly, this is just Haskell's `scanl1 (+)` with a leading zero, but either a comment here or (maybe better) above the *TotalDaysByMonth arrays would be nice.

src/NodaTime/Calendars/IslamicYearMonthDayCalculator.cs
```diff
@@ -41,22 +41,23 @@ internal sealed class IslamicYearMonthDayCalculator : RegularYearMonthDayCalcula
         /// <summary>The pattern of leap years within a cycle, one bit per year, for this calendar.</summary>
         private readonly int leapYearPatternBits;
 
-        private static readonly int[] TotalDaysByMonth;
+        private static readonly int[] TotalDaysByMonth = GenerateTotalDaysByMonth();
```
malcolmr, line 44: again, a comment as to what this contains would be nice, since it's not immediately obvious whether the result is `[0, 29, ...]` or `[29, 59, ...]`.

src/NodaTime/Calendars/PersianYearMonthDayCalculator.cs
```diff
@@ -24,23 +24,24 @@ internal abstract class PersianYearMonthDayCalculator : RegularYearMonthDayCalcu
         private const int AverageDaysPer10Years = (DaysPerNonLeapYear * 25 + DaysPerLeapYear * 8) * 10 / 33;
         internal const int MaxPersianYear = 9377;
 
-        private static readonly int[] TotalDaysByMonth;
+        private static readonly int[] TotalDaysByMonth = GenerateTotalDaysByMonth();
```
malcolmr, line 27: (same)

src/NodaTime/Text/Patterns/SteppedPatternBuilder.cs
```diff
@@ -247,13 +247,15 @@ internal static void HandleBackslash(PatternCursor pattern, SteppedPatternBuilde
         }
 
 #pragma warning disable IDE0060
```
malcolmr, line 249: This appears to be the same as CA1801. Do we need both? (Do we need this elsewhere?)

src/NodaTime/DateInterval.cs
```diff
@@ -204,7 +204,7 @@ public void Deconstruct(out LocalDate start, out LocalDate end)
         /// </returns>
         /// <exception cref="ArgumentException"><paramref name="interval" /> uses a different
         /// calendar to this date interval.</exception>
-        public DateInterval? Intersection(DateInterval interval) =>
+        public DateInterval? Intersection([ValidatedNotNull] DateInterval interval) =>
```
malcolmr, line 207: Odd, but I assume that this one _is_ still needed even though the first thing that happens is to pass `interval` to a methods with `[ValidatedNotNull]`?

(That is, we can't just stick this on the lower-level methods like `ValidateInterval()` and be done with it.)

src/NodaTime/LocalDate.cs
```diff
@@ -235,6 +235,7 @@ public static LocalDate FromDateTime(DateTime dateTime)
         /// <returns>A new <see cref="LocalDate"/> with the same values as the specified <c>DateTime</c>.</returns>
         public static LocalDate FromDateTime(DateTime dateTime, CalendarSystem calendar)
         {
+            Preconditions.CheckNotNull(calendar, nameof(calendar));
```
malcolmr, line 238: (I'm surprised we didn't already have tests that do something like Guava's `NullPointerTester` to exercise all the public static/instance methods.)

src/NodaTime.Testing/TimeZones/MultiTransitionDateTimeZone.cs
```diff
@@ -70,11 +70,16 @@ public override ZoneInterval GetZoneInterval(Instant instant)
             throw new InvalidOperationException($"Instant {instant} did not exist in time zone {Id}");
         }
 
+// This isn't really a collection type; it only implements IEnumerable to enable collection initializers.
+#pragma warning disable CA1010 // Implement IEnumerable<T>
+#pragma warning disable CA1710 // Rename to class end in "Collection"
```
malcolmr, line 75: "Rename class to end"?

src/NodaTime/TimeZones/BclDateTimeZoneSource.cs
```diff
@@ -97,6 +97,8 @@ private static IEnumerable<string> GetTimeZoneInfoLocalIdOrEmpty()
         /// </remarks>        
         DateTimeZone IDateTimeZoneSource.ForId(string id) => ForId(id);
 
+// Even though this member could be static, it would be inconsistent with the interface member.
```
malcolmr, line 100: (also a breaking change, as you note in the commit message.)

src/NodaTime/TimeZones/IO/TzdbStreamData.cs
```diff
@@ -118,10 +119,16 @@ public DateTimeZone CreateZone(string id, string canonicalId)
         internal static TzdbStreamData FromStream(Stream stream)
         {
             Preconditions.CheckNotNull(stream, nameof(stream));
-            int version = new BinaryReader(stream).ReadInt32();
-            if (version != AcceptedVersion)
+
+            // Using statement to satisfy FxCop, but disposable won't do anything anyway, because
+            // we deliberately leave the stream open.
+            using (var reader = new BinaryReader(stream, Encoding.UTF8, leaveOpen: true))
```
malcolmr, line 125: ... but isn't this actually fixing a bug? Disposal can happen at any time(?), and if `BinaryReader.Dispose()` closes the stream by default (as it seems like it does), then this could have failed future reads, couldn't it?

src/NodaTime/TimeZones/IO/TzdbStreamData.cs
```diff
@@ -118,10 +119,16 @@ public DateTimeZone CreateZone(string id, string canonicalId)
         internal static TzdbStreamData FromStream(Stream stream)
         {
             Preconditions.CheckNotNull(stream, nameof(stream));
-            int version = new BinaryReader(stream).ReadInt32();
-            if (version != AcceptedVersion)
+
+            // Using statement to satisfy FxCop, but disposable won't do anything anyway, because
```
malcolmr, line 123: s/disposable/dispose/?

src/NodaTime/Period.cs
```diff
@@ -354,6 +363,16 @@ private Period(long hours, long minutes, long seconds, long milliseconds, long t
                 minuend.Nanoseconds - subtrahend.Nanoseconds);
         }
 
+        /// <summary>
+        /// Subtracts one period from another, by simply subtracting each property value.
+        /// </summary>
+        /// <param name="minuend">The period to subtract the second operand from</param>
+        /// <param name="subtrahend">The period to subtract the first operand from</param>
+        /// <returns>The result of subtracting all the values in the second operand from the values in the first. The
+        /// units of the result will be the union of both periods, even if the subtraction caused some properties to
```
malcolmr, line 372: On reading this, I wondered if we should say (here and above) that it's valid to e.g. do 2w - 2w1d = -1d (where we treat the first argument as implicitly 2w0d)... but I couldn't find a good way to do so, and maybe it's obvious?

malcolmr (approved):
LGTM. Only blocking question should be about whether the switch to culture-insensitive string comparisons could possibly break anyone in practice; everything else is optional.

jskeet (commented):
Thanks for all the comments. I've added another commit to the end to address them accordingly; will merge on green.

src/NodaTime/TimeZones/IO/TzdbStreamData.cs
malcolmr: Oh, right, implementing `IDisposable` doesn't imply a finalizer, good point.

src/NodaTime/YearMonthDay.cs
```diff
@@ -56,7 +56,7 @@ internal YearMonthDay(int year, int month, int day)
         internal static YearMonthDay Parse(string text)
         {
             // Handle a leading - to negate the year
-            if (text.StartsWith("-"))
+            if (text.StartsWith("-", StringComparison.Ordinal))
```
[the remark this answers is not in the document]
malcolmr: (I hadn't actually noticed these were testing methods.)

PR #1598, opened 2020-12-02
Define all XML schema types as complex types with simple content

### LINKED ISSUE #1597: OffsetDateTime gets generated as String by svcutil
When generating data contracts with `svcutil` any property of type `OffsetDateTime` will get generated as a `string` instead of `OffsetDateTime`. I attached a simple project that reproduces the problem:
1. Install Nuget packages and build the project
2. Run `Run.bat` (it expects that .NET Framework 4.7.2 SDK tools are installed).
3. The outputted `Client.cs` file has string as the type of property `Test1` of class `Test` instead of type `NodaTime.OffsetDateTime`.

I think this happens because the Xml schema of `OffsetDatetime` [XmlSchemaDefinition.cs#L100](https://github.com/nodatime/nodatime/blob/master/src/NodaTime/Xml/XmlSchemaDefinition.cs#L100) is defined as a pattern and does not have a `calendarRestriction` like for example [XmlSchemaDefinition.cs#L111](https://github.com/nodatime/nodatime/blob/master/src/NodaTime/Xml/XmlSchemaDefinition.cs#L111). 

At least if I manually change the `nodatime.org.api.xsd` file that `svcutil` generates to include the `calendarRestriction` then it works as expected.

[NodaTime3SvcUtilOffsetDateTime.zip](https://github.com/nodatime/nodatime/files/5617338/NodaTime3SvcUtilOffsetDateTime.zip)

0xced:
From the validation and documentation perspective, having a simple type or a complex type is equivalent.

The benefit of having all NodaTime CLR types described as complex XML schema types is for external tools such as svcutil which behave better with complex types.

Fixes #1597

### DIFF
--- a/src/NodaTime.Test/Xml/XmlSchemaTest.XmlSchema.approved.xml
+++ b/src/NodaTime.Test/Xml/XmlSchemaTest.XmlSchema.approved.xml
@@ -1,23 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xs:schema xmlns:nodatime="https://nodatime.org/api/" targetNamespace="https://nodatime.org/api/" xmlns:xs="http://www.w3.org/2001/XMLSchema">
-  <xs:simpleType name="AnnualDate">
-    <xs:restriction base="xs:string">
-      <xs:pattern value="[0-9]{2}-[0-9]{2}" />
-    </xs:restriction>
-  </xs:simpleType>
-  <xs:simpleType name="Duration">
-    <xs:restriction base="xs:string">
-      <xs:pattern value="-?[0-9]{1,8}:[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]{1,9})?" />
-    </xs:restriction>
-  </xs:simpleType>
-  <xs:simpleType name="Instant">
-    <xs:restriction base="xs:string">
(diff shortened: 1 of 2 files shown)


jskeet:
Will have a look at this when I get the change - it may not be for a little while.

jskeet (commented):
Okay, I've now had a look... and I really don't understand it, I'm afraid! I can somewhat see the point, but I don't know how to review it for correctness. Is there anything I could usefully review carefully, or are you sufficiently confident that I should just merge it?

0xced:
All the tests that validate the XML documents still pass which is the main point of the XML schema. When I first wrote the XML schema, I did not give much thought about using simple types vs complex types, except for `LocalDate`, `LocalDateTime` and `ZonedDateTime` where the complex type was [technically required](https://stackoverflow.com/questions/626319/add-attributes-to-a-simpletype-or-restriction-to-a-complextype-in-xml-schema/626385#626385) because of the XML attributes.

I think it's a good idea to have all types described homogeneously as complex types. As @panthus found out and confirmed (see #1597), external tooling (such as `svcutil`) gives better results with complex types. This is a compelling reason to switch all types as complex types in my opinion.

In short, I'm pretty confident you can merge this pull request.

jskeet:
Great, thanks very much :)

PR #1620, opened 2021-03-13
First pass at GitHub actions

jskeet:
Even though "Build push" and "Build PR" are the same now, they're
separate so that they can diverge easily in the future, and to avoid
confusing the badges.


jskeet:
(It will be easiest to merge this immediately, then take the fall-out... the first pass at GitHub actions tends to be a bit frustrating.)

PR #1624, opened 2021-03-23
Mark all build scripts as executable

eatdrinksleepcode:
Executable permissions are necessary to run scripts on non-Windows OSes.

`chmod u+x **/*.sh`


jskeet (approved):
Thanks :)

PR #1630, opened 2021-04-30
Microsecond constants

lilinus:
Felt like these time adjusters and nanoseconds per microsecond constant was missing.

Useful for example in postgresql where times are stored with microsecond precision.

### DIFF
--- a/src/NodaTime/NodaConstants.cs
+++ b/src/NodaTime/NodaConstants.cs
@@ -84,7 +84,7 @@ namespace NodaTime
         /// A constant for the number of microseconds per second.
         /// The value of this constant is 1,000,000.
         /// </summary>
-        public const long MicrosecondsPerMillisecond = 1000000;
+        public const long MicrosecondsPerSecond = 1000000;
 
         /// <summary>
         /// A constant for the number of milliseconds per second.
--- a/src/NodaTime/NodaConstants.cs
+++ b/src/NodaTime/NodaConstants.cs
@@ -82,6 +82,12 @@ namespace NodaTime
 
         /// <summary>
         /// A constant for the number of milliseconds per second.
+        /// The value of this constant is 1,000,000.
+        /// </summary>
+        public const long MicrosecondsPerMillisecond = 1000000;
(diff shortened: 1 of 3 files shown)


jskeet:
I'm happy with adding constants for microseconds - I'd suggest both NanosecondsPerMicrosecond and MicrosecondsPerSecond, but probably only those.
I'm also happy with a TruncateToMillisecond time adjuster (not TruncateToMilliseconds - please be consistent with the other adjusters) - milliseconds are widely used in the NodaTime API.

However, I'd really prefer not to add a time adjuster for TruncateToMicrosecond, as that would be the *only* place in the NodaTime API that referred to microseconds (other than the new constants). It feels to me like the new constants are a good way of those who want microseconds writing code that uses them without hard-coding any magic numbers... but I don't want to introduce them either inconsistently or throughout the API.

It's easy enough for you to add the time adjuster within your own codebase, and I think that's better than introducing it in NodaTime.

src/NodaTime/NodaConstants.cs
```diff
@@ -75,6 +80,12 @@ public static class NodaConstants
         /// </summary>
         public const long TicksPerWeek = TicksPerDay * DaysPerWeek;
 
+        /// <summary>
+        /// A constant for the number of milliseconds per second.
```
jskeet, line 84:
```suggestion
        /// A constant for the number of microseconds per second.
```

lilinus:
Thanks for the feedback. Have changed the branch so only the two constants mentioned are added.

src/NodaTime/NodaConstants.cs
```diff
@@ -75,6 +80,12 @@ public static class NodaConstants
...
+        /// The value of this constant is 1,000,000.
+        /// </summary>
+        public const long MicrosecondsPerMillisecond = 1000000;
```
jskeet, line 87:
```suggestion
        public const long MicrosecondsPerSecond = 1000000;
```

jskeet (approved):

PR #1631, opened 2021-05-01
Format Period.Zero as "P0D" in the roundtrip pattern

### LINKED ISSUE #1629: ToString on Empty Period does not follow ISO8601 returns 'P'
When doing ToString on a zero period (happen on both lines in the unittest below), I get `'P'` as the result. This is not valid ISO8601, as this requires a zero period to have at least a single field given, such as `'P0S'`, of `'P0D'`. This also seems to contradict the behavior described in ticket #46.

```
        [Test]
        public void DurationToString()
        {
            Assert.AreEqual("P0S", Period.FromSeconds(0).ToString(), "Period 0 seconds"); // Actually is 'P'
            Assert.AreEqual("P0S", Period.Zero.ToString(), "Period without fields"); // Actually is 'P'
        }
```

Note that we run into empty Periods when storing a 0 period in our postgresql database, and retrieve it using the NodaTime type mapping in entity framework these periods then get serialized to JSON using  `NodaTime.Serialization.SystemTextJson`. 

Tested using nodatime 3.0.5

jskeet:
Fixes #1629

Note that we still allow "P" on its own to be parsed, for backward compatibility. (This came for free, as it happens.)

### DIFF
--- a/src/NodaTime/Text/PeriodPattern.cs
+++ b/src/NodaTime/Text/PeriodPattern.cs
@@ -184,6 +184,15 @@ namespace NodaTime.Text
             {
                 Preconditions.CheckNotNull(value, nameof(value));
                 Preconditions.CheckNotNull(builder, nameof(builder));
+                // Always ensure we've got *some* unit to ensure the result is valid in ISO-8601; arbitrarily pick days.
+                // Note: "P0S" might be nicer here, but NormalizingIsoPatternImpl picked "P0D" a
+                // long time ago and we want to be consistent between the two.
+                // We might want to make both pattern implementations configurable at some point.
+                if (value.Equals(Period.Zero))
+                {
+                    builder.Append("P0D");
+                    return builder;
+                }
                 builder.Append('P');
                 AppendValue(builder, value.Years, 'Y');
                 AppendValue(builder, value.Months, 'M');
--- a/src/NodaTime.Test/PeriodTest.cs
+++ b/src/NodaTime.Test/PeriodTest.cs
@@ -617,7 +617,7 @@ namespace NodaTime.Test
         [Test]
         public void ToString_Zero()
         {
-            Assert.AreEqual("P", Period.Zero.ToString());
+            Assert.AreEqual("P0D", Period.Zero.ToString());
         }
 
         [Test]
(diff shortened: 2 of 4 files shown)


jskeet:
cc @mennolodder
I'll cherry-pick this into the 2.4.x and 3.0.x branches after merging, and it'll be picked up that way next time there's a TZDB release. (That's the simplest way of getting it released.)

src/NodaTime/Text/PeriodPattern.cs
```diff
@@ -184,6 +184,14 @@ public StringBuilder AppendFormat(Period value, StringBuilder builder)
             {
                 Preconditions.CheckNotNull(value, nameof(value));
                 Preconditions.CheckNotNull(builder, nameof(builder));
+                // Always ensure we've got *some* unit; arbitrarily pick days.
```
malcolmr, line 187: "... to ensure that the result is valid per ISO 8601" or similar?

src/NodaTime/Text/PeriodPattern.cs
```diff
@@ -184,6 +184,14 @@ public StringBuilder AppendFormat(Period value, StringBuilder builder)
...
+                // Note: "P0S" might be nicer here, but the normalizing version picked "P0D"
```
malcolmr, line 188: I was originally confused as to what "the normalizing version" and "both" (below) referred to; perhaps expand a bit?

malcolmr (approved):
LGTM

src/NodaTime/Text/PeriodPattern.cs
jskeet: Done.

src/NodaTime/Text/PeriodPattern.cs
jskeet: Done - PTAL. (Will rebase back to a single comment once you're happy with it.)

src/NodaTime/Text/PeriodPattern.cs
```diff
@@ -184,6 +184,14 @@ public StringBuilder AppendFormat(Period value, StringBuilder builder)
             {
                 Preconditions.CheckNotNull(value, nameof(value));
                 Preconditions.CheckNotNull(builder, nameof(builder));
+                // Always ensure we've got *some* unit to ensure the result is valid in ISO-8601; arbitrarily pick days.
+                // Note: "P0S" might be nicer here, but NormalizingIsoPatternImpl picked "P0D" a long time ago.
```
malcolmr, line 188: "... and we want to be consistent"? (or is there some other reason they need to be the same?)

malcolmr (approved):
nitpicking below, but still LGTM :)

src/NodaTime/Text/PeriodPattern.cs
jskeet: Nope, that makes sense. Will add that, rebase, force push, and merge.

PR #1638, opened 2021-06-17
Fix global.json

0xced:
The SDK version must have major.minor.patch else it doesn't work as expected (i.e. always use the latest available SDK).

- With with following .NET SDKs installed:
  2.1.816 [/usr/local/share/dotnet/sdk]
  3.1.410 [/usr/local/share/dotnet/sdk]
  5.0.301 [/usr/local/share/dotnet/sdk]
  6.0.100-preview.4.21255.9 [/usr/local/share/dotnet/sdk]
  - With "rollForward": "latestMinor"
    - With "version": "3.1" in global.json
      ❌ `dotnet --version` returns 6.0.100-preview.4.21255.9
    - With "version": "3.1.410" in global.json
      ✅ `dotnet --version` returns 3.1.410

  - With "rollForward": "latestMajor" and "allowPrerelease": false,
    - With "version": "3.1" in global.json
      ❌ `dotnet --version` returns 6.0.100-preview.4.21255.9 (works by chance)
    - With "version": "3.1.410" in global.json
      ✅ `dotnet --version` returns 5.0.301

That's a lot of words to say that you can't use `3.1` but must use `3.1.410`.

Also, explicitly set `allowPrerelease` to `false`.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,7 @@
 {
   "sdk": {
-    "version": "3.1",
+    "version": "3.1.410",
+    "allowPrerelease": false,
     "rollForward": "latestMinor"
   }
 }
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json.schemastore.org/global",
   "sdk": {
-    "version": "3.1.410",
+    "version": "5.0.301",
     "allowPrerelease": false,
     "rollForward": "latestMinor"
   }
--- a/global.json
+++ b/global.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "http://json.schemastore.org/global",
   "sdk": {
     "version": "3.1.410",
     "allowPrerelease": false,
--- a/src/NodaTime.sln
+++ b/src/NodaTime.sln
@@ -32,6 +32,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "MSBuild", "MSBuild", "{33DA
 	ProjectSection(SolutionItems) = preProject
 		..\Directory.Build.props = ..\Directory.Build.props
 		..\Directory.Build.targets = ..\Directory.Build.targets
+		..\global.json = ..\global.json
 	EndProjectSection
 EndProject
 Global


global.json
```diff
@@ -1,6 +1,8 @@
 {
+  "$schema": "http://json.schemastore.org/global",
   "sdk": {
-    "version": "3.1",
+    "version": "5.0.301",
```
jskeet, line 4: Any reason for going to 5.0.301 rather than a specific 3.1 SDK? I'd really prefer to stick to LTS SDK releases if at all possible.

0xced: That's because the `NodaTime.TzdbCompiler` project multi-targets `netcoreapp3.1` + `net5.0` (introduced in fb880b06a2c41def4f1f9a218703d302e3523492). The `net5.0` target framework requires the .NET 5 SDK and since the `global.json` was misconfigured, the SDK 5 was actually always used.

I have removed the `net5.0` target in order to stay on the .NET Core 3.1 LTS SDK in 54f3bf9e8db5c16a0b3d801de5ece0e1d83e1e2e.

jskeet: Aha... okay. I'll need to have a think about this. I can't remember offhand why I wanted to be able to run the TzdbCompiler under .NET 5 explicitly... maybe we'll want to use 5.0 after all. I'm busy for the rest of today, but I'll hopefully get back to you about this tomorrow.

0xced: You wrote this in commit fb880b06a2c41def4f1f9a218703d302e3523492:
> .NET 5.0 has more instrumentation for null references in LINQ to XML

I'll let you think about it and I'll adapt the version in `global.json` and the target frameworks to whatever you think is the best for Noda Time.

jskeet: Right, I've had a think - I suspect that what you had before, with global.json using the 5.0 SDK, is probably the best course of action for the moment. I'll update it to 6 when that comes out. Sorry about the inconvenience here!

jskeet (approved):

jskeet:
Will merge on green - thanks!

global.json
0xced: I think that's a good call. 👍

PR #1640, opened 2021-07-02
Correct documentation on constructor

PaulEthertonDP:
The summary of one of the YearMonth constructors contained an erroneous reference to "Day", presumably because the documentation was copied from [here](https://github.com/nodatime/nodatime/blob/aa49ace02d8a46668033e75e2fec0fbbf935e068/src/NodaTime/LocalDate.cs#L96) and edited when first created.

The change is simply to correct this.

### DIFF
--- a/src/NodaTime/YearMonth.cs
+++ b/src/NodaTime/YearMonth.cs
@@ -97,7 +97,7 @@ namespace NodaTime
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


jskeet (approved):

jskeet:
Thanks! I'll fix it up in the 3.0.x branch too.

PR #1642, opened 2021-07-03
Make slow tests opt-in instead of opt-out 

### LINKED ISSUE #1625: Make "slow" tests opt-in instead of opt-out
(Copied from [#1182](https://github.com/nodatime/nodatime/issues/1182#issuecomment-751635345))

CI skips "slow" tests (the ones which verify an entire calendar), which are apparently only run prior to release. However, the instructions for running the tests don't mention the "NoSlowTests.runsettings" file, which on my machine drops the time for "dotnet test NodaTime.Test" from about 2.5 minutes to 40 seconds. It's also not necessarily easy to skip the slow tests when running through an IDE's test runner, and doing so would be IDE-specific.

It would be great if running these slow tests were opt-in instead of opt-out, which would make skipping them the default for developers. This could be achieved using the Explicit attribute instead of Category, or possibly by using Assert.Ignore (based on an environment variable?).

eatdrinksleepcode:
Closes #1625

The NUnit test adapter update is required for the `--filter` switch to work as desired for this use case in `dotnet test`. See https://github.com/nunit/nunit3-vs-adapter/issues/473.

### DIFF
--- a/build/appveyor.sh
+++ b/build/appveyor.sh
@@ -14,7 +14,7 @@ dotnet --info
 
 dotnet build -c Release src/NodaTime.sln
 
-dotnet test -c Release src/NodaTime.Test -s src/NodaTime.Test/NoSlowTests.runsettings
+dotnet test -c Release src/NodaTime.Test
 dotnet test -c Release src/NodaTime.Demo
 
 dotnet build -c Release src/NodaTime.TzdbCompiler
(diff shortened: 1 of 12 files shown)


jskeet (approved):

jskeet:
That took less time to review than expected - thanks so much for doing this!

eatdrinksleepcode:
My pleasure! Sorry it took so long...ya know, life.

PR #1643, opened 2021-07-03
Run test coverage on .NET 5 and on Mac

eatdrinksleepcode:
While working on #1642 I needed to change the [coverageparams.xml](/nodatime/nodatime/blob/master/src/NodaTime.Test/coverageparams.xml) file, and I wanted to verify the change. However, I work mostly on a Mac, and the `coverage.sh` script is dependent on Windows: both because it runs against .NET Framework, and because it runs exe files.

This PR updates the versions of dotCover and ReportGenerator, changes them to run against .NET 5 instead of .NET Framework, removes the dependency on `nuget.exe` in favor of `dotnet restore`, and creates a `coverage-mac.sh` script which will run coverage on a Mac.

I did verify that the `coverage.xml` output file is the same after the change to running against .NET 5 (there are some changes compared to master, but those are caused by the version change), and that the report can be generated and _appears_ correct on both Mac and Windows.

### DIFF
--- a/build/coverage.sh
+++ b/build/coverage.sh
@@ -11,16 +11,17 @@ declare -r REPORTGENERATOR_VERSION=4.8.11
 
 export REPORTGENERATOR_VERSION
 export DOTCOVER_VERSION
+export DOTCOVER_PACKAGE_SUFFIX
 dotnet restore $ROOT/build/Coverage.proj --packages $ROOT/packages
 
-declare -r DOTCOVER_DIR=$ROOT/packages/jetbrains.dotcover.commandlinetools/$DOTCOVER_VERSION/tools
+declare -r DOTCOVER_DIR=$ROOT/packages/jetbrains.dotcover.commandlinetools${DOTCOVER_PACKAGE_SUFFIX}/$DOTCOVER_VERSION/tools
 
 rm -rf $ROOT/coverage
 mkdir $ROOT/coverage
 
 # Run the tests under dotCover
 (cd $DOTCOVER_DIR; 
- ./dotCover.exe dotnet $TEST/coverageparams.xml --Output=$ROOT/coverage/coverage.xml --ReportType=DetailedXML --ReturnTargetExitCode -- test $TEST)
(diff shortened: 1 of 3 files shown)


build/coverage.sh
```diff
@@ -6,28 +6,28 @@ set -e
...
-declare -r DOTCOVER_VERSION=2019.3.4
-declare -r REPORTGENERATOR_VERSION=4.0.12
+declare -r DOTCOVER_VERSION=2021.1.3
+declare -r REPORTGENERATOR_VERSION=4.8.11
 
-nuget.exe install -Verbosity quiet -OutputDirectory $ROOT/packages -Version $DOTCOVER_VERSION JetBrains.dotCover.CommandLineTools
+export REPORTGENERATOR_VERSION
+export DOTCOVER_VERSION
+export DOTCOVER_PACKAGE_SUFFIX
+dotnet restore $ROOT/build/Coverage.proj --packages $ROOT/packages
```
jskeet, line 15: Gosh, I haven't seen this before. I may very well use this elsewhere :)

build/coverage.sh
```diff
@@ -6,28 +6,28 @@ set -e
...
 
 rm -rf $ROOT/coverage
 mkdir $ROOT/coverage
 
 # Run the tests under dotCover
-(cd $TEST; 
- $DOTCOVER cover coverageparams.xml -ReturnTargetExitCode)
-
-$DOTCOVER report --Source=$ROOT/coverage/NodaTime.dvcr --Output=$ROOT/coverage/coverage.xml --ReportType=DetailedXML ""
+(cd $DOTCOVER_DIR;
```
jskeet, line 23: We've lost the -ReturnTargetExitCode parameter here - was that deliberate because the behavior has changed, or accidental?

jskeet (commented):
Mostly fine - just one question

build/coverage.sh
eatdrinksleepcode: I thought at one point that I was getting a "not used parameter" warning from dotCover for ReturnTargetExitCode, but I just checked and that is not the case. I added it back.

jskeet (approved):

PR #1653, opened 2021-09-25
Add rfc3339 pattern for ZonedDateTime

sam-mfb:
Add a standard pattern for the zoned date time format proposed as an [extension to RFC 3339](https://ryzokuken.dev/draft-ryzokuken-datetime-extended/documents/rfc-3339.html),

This format appears likely to be approved next year (see [this issue](https://github.com/tc39/proposal-temporal/issues/1450)).  It is the format that the proposed [Temporal](https://tc39.es/proposal-temporal/docs/iso-string-ext.html) addition to ECMA wants to use, and that is currently used by [java.time](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#toString--).

The PR is being submitted in draft form because, even if otherwise acceptable, it should await final approval of the RFC 3339 proposal.  There will be an [accompanying PR for the serialization library](https://github.com/nodatime/nodatime.serialization/pull/77) as well, which I'm happy to keep working on when I have a little more time.

#jskeet a few thoughts/questions in putting this together:
- I went with "e" and "E" as the pattern characters; I didn't see any collisions with other patterns and they suggest "extension"
- It's not clear to me that technically RFC3339 will support time zone providers other than IANA, but I didn't see any reason to so restrict the pattern being exposed; this way it's consistent with G and F

### DIFF
--- a/src/NodaTime.Test/Text/ZonedDateTimePatternTest.cs
+++ b/src/NodaTime.Test/Text/ZonedDateTimePatternTest.cs
@@ -145,13 +145,13 @@ namespace NodaTime.Test.Text
             // Standard patterns without a DateTimeZoneProvider
             new Data(MsdnStandardExampleNoMillis) { StandardPattern = ZonedDateTimePattern.GeneralFormatOnlyIso, Pattern = "G", Text = "2009-06-15T13:45:30 UTC (+00)", Culture = Cultures.FrFr, ZoneProvider = null},
             new Data(MsdnStandardExample) { StandardPattern = ZonedDateTimePattern.ExtendedFormatOnlyIso, Pattern = "F", Text = "2009-06-15T13:45:30.09 UTC (+00)", Culture = Cultures.FrFr, ZoneProvider = null },
-            new Data(MsdnStandardExampleNoMillis) { StandardPattern = ZonedDateTimePattern.GeneralFormatRfc3339SuffixOnly, Pattern = "e", Text = "2009-06-15T13:45:30+00[UTC]", Culture = Cultures.FrFr, ZoneProvider = null},
-            new Data(MsdnStandardExample) { StandardPattern = ZonedDateTimePattern.ExtendedFormatRfc3339SuffixOnly, Pattern = "E", Text = "2009-06-15T13:45:30.09+00[UTC]", Culture = Cultures.FrFr, ZoneProvider = null },
+            new Data(MsdnStandardExampleNoMillis) { StandardPattern = ZonedDateTimePattern.GeneralRfc3339TimeZoneSuffix, Pattern = "e", Text = "2009-06-15T13:45:30+00[UTC]", Culture = Cultures.FrFr, ZoneProvider = null},
+            new Data(MsdnStandardExample) { StandardPattern = ZonedDateTimePattern.ExtendedRfc3339TimeZoneSuffix, Pattern = "E", Text = "2009-06-15T13:45:30.09+00[UTC]", Culture = Cultures.FrFr, ZoneProvider = null },
             // Standard patterns without a resolver
             new Data(MsdnStandardExampleNoMillis) { StandardPattern = ZonedDateTimePattern.GeneralFormatOnlyIso, Pattern = "G", Text = "2009-06-15T13:45:30 UTC (+00)", Culture = Cultures.FrFr, Resolver = null},
             new Data(MsdnStandardExample) { StandardPattern = ZonedDateTimePattern.ExtendedFormatOnlyIso, Pattern = "F", Text = "2009-06-15T13:45:30.09 UTC (+00)", Culture = Cultures.FrFr, Resolver = null },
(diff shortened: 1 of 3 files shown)


jskeet:
Thanks, I'll try to have a look at this hopefully late tomorrow afternoon...

jskeet:
Actual implementation is fine (and pleasantly brief - it's nice when abstractions just work :)

Given that we always have an IANA zone provider available, and that the RFC effectively expects that, I think it would be reasonable to make that the default (and remove the FormatOnly part from the internal fields). 

Not sure about the names "GeneralFormatRfc3339SuffixOnly" and "ExtendedFormatRfc3339SuffixOnly". Presumably the "suffix-only" part is intended to express the lack of support for parameters?

sam-mfb:
>Actual implementation is fine (and pleasantly brief - it's nice when abstractions just work :)

This codebase is really really nice.

>Given that we always have an IANA zone provider available, and that the RFC effectively expects that, I think it would be reasonable to make that the default (and remove the FormatOnly part from the internal fields).

I'll re-work that.

>Not sure about the names "GeneralFormatRfc3339SuffixOnly" and "ExtendedFormatRfc3339SuffixOnly". Presumably the "suffix-only" part is intended to express the lack of support for parameters?

Yes, but that doesn't quite scan, now that I read it.  What about (in connection with the change you suggested to allow it to Parse using IANA by default): `GeneralRfc3339TimeZoneOnly` and `ExtendedRfc3339TimeZoneOnly`?

jskeet:
> Yes, but that doesn't quite scan, now that I read it. What about (in connection with the change you suggested to allow it to Parse using IANA by default): GeneralRfc3339TimeZoneOnly and ExtendedRfc3339TimeZoneOnly?

Hmm... that sounds like it's only got the time zone rather than anything else. Let's both continue mulling - there's relatively little urgency, given the state of the RFC. But thanks very much for doing this!

sam-mfb:
I added tzdb as the default provider and tweaked the names accordingly (better?).  The concept of a "default" provider for a pattern is new (I think), so I had to (slightly) adjust the pattern parser and test framework to accommodate that.  Let me know if you see a more elegant way to do that.

src/NodaTime.Test/Text/ZonedDateTimePatternTest.cs
```diff
@@ -208,6 +212,12 @@ public class ZonedDateTimePatternTest : PatternTestBase<ZonedDateTime>
...
             new Data(2013, 01, 13, 15, 44, 30, 90, TestZone1) { StandardPattern = ZonedDateTimePattern.ExtendedFormatOnlyIso.WithZoneProvider(TestProvider), Pattern = "F", Text = "2013-01-13T15:44:30.09 ab (+02)", Culture = Cultures.FrFr },
+            new Data(2013, 01, 13, 15, 44, 30, 0, TestZone1) { StandardPattern = ZonedDateTimePattern.GeneralRfc3339TimeZoneSuffix.WithZoneProvider(TestProvider) , Pattern = "e", Text = "2013-01-13T15:44:30+02[ab]", Culture = Cultures.FrFr },
+            new Data(2013, 01, 13, 15, 44, 30, 90, TestZone1) { StandardPattern = ZonedDateTimePattern.ExtendedRfc3339TimeZoneSuffix.WithZoneProvider(TestProvider), Pattern = "E", Text = "2013-01-13T15:44:30.09+02[ab]", Culture = Cultures.FrFr },
+
+            // RFC 3339 standard patterns with default time zone provider
+            new Data(2013, 09, 13, 15, 44, 30, 0, Athens, true) { StandardPattern = ZonedDateTimePattern.GeneralRfc3339TimeZoneSuffix , Pattern = "e", Text = "2013-09-13T15:44:30+03[Europe/Athens]", Culture = Cultures.EnUs },
```
jskeet, line 219: I think it's worth using named arguments here to explain the `true`

src/NodaTime/Text/ZonedDateTimePatternParser.cs
```diff
@@ -82,6 +82,18 @@ public IPattern<ZonedDateTime> ParsePattern(string patternText, NodaFormatInfo f
                     'F' => ZonedDateTimePattern.Patterns.ExtendedFormatOnlyPatternImpl
                             .WithZoneProvider(zoneProvider)
                             .WithResolver(resolver),
+                    'e' => (zoneProvider == null) ?
```
jskeet, line 85: Hmm. I need to think about this a bit more. I think it's probably most consistent to be able to pass in null as the zone provider and get a format-only pattern... but still keeping the public properties defaulting to TZDB.

I suspect we may end up needing more internal properties for this, I'm afraid. I'll try to experiment with it later in the week.

sam-mfb: absolutely.  no urgency at all.  if i get some time, i'll work on the serialization side of things as we discussed in the other PR.  no urgency there either.  i just figured as I have time--and before thinking about timezones drops out of my cache :)-- I'll try to push the ball along so that hopefully things are a little easier for you when you get the chance to focus on it.  but yeah, on the current rfc timeline, it looks like this is a 2022 merge regardless.  good luck with the avalanche!

Issue #1635, opened 2021-05-28
Support conversion to/from DateOnly and TimeOnly


mattjohnsonpint:
.NET 6 Preview 4 added `DateOnly` and `TimeOnly` structs.  I would like to propose that Noda Time should support two-way conversion between `DateOnly` and `LocalDate`, and between `TimeOnly` and `LocalTime`.

```csharp
public struct LocalDate
{
    public static LocalDate FromDateOnly(DateOnly date);
    public static LocalDate FromDateOnly(DateOnly date, CalendarSystem calendar);
    public DateOnly ToDateOnly();
}

public struct LocalTime
{
    public static LocalTime FromTimeOnly(TimeOnly time);
    public TimeOnly ToTimeOnly();
}
```

jskeet:
Hmm. I'm definitely very happy with this in principle... I do feel we took a positive step in dropping all our conditional logic back in 3.0 to target just netstandard2.0. Presumably this would add a target of net60.

One *option* would be to have a separate package just for this. Not sure that helps much either though. I can probably live with the multi-targeting if we have to... it does feel like a useful thing to have.

malcolmr:
Seems reasonable to me in general, but would there be a way of doing this without making the NodaTime library .NET 6-only?

- The conversion from `LocalTime` to `TimeOnly` would be slightly lossy (`LocalTime` has nanosecond resolution; AIUI `TimeOnly` has 100ns resolution, presumably we'd just round down).

- `DateOnly` is a bit more complex. To start, I can't tell whether it can represent years prior to 1 CE, but if not, then there are some `LocalDate` in the ISO/Gregorian calendar that wouldn't convert (we support back to 9999 BCE, IIRC).

- It's also not obvious to me whether `DateOnly` assumes any particular calendar. I'm _guessing_ it just wants numbers that fit into a perpetual Gregorian calendar, which makes me wonder whether the conversion should just attempt to make a `DateOnly` with matching day/month/year (even though that probably wouldn't make too much sense for non-Gregorian calendars, and might even fail for e.g. some dates in the Hebrew calendar), or whether it should have an implicit `.WithCalendar(CalendarSystem.Iso)` first (which could also fail in some cases, of course).

[...]

PR #1665, opened 2021-11-19
Document conversions to BCL types for non-Gregorian calendars

jskeet:
This doesn't change the behavior at all; it just documents it.

This is the first step towards implementing #1635.

### DIFF
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -198,10 +198,18 @@ namespace NodaTime
         /// by this value.
         /// </summary>
         /// <remarks>
+        /// <para>
         /// <see cref="DateTimeKind.Unspecified"/> is slightly odd - it can be treated as UTC if you use <see cref="DateTime.ToLocalTime"/>
         /// or as system local time if you use <see cref="DateTime.ToUniversalTime"/>, but it's the only kind which allows
         /// you to construct a <see cref="DateTimeOffset"/> with an arbitrary offset, which makes it as close to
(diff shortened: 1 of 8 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -198,10 +198,18 @@ public LocalDate(Era era, int yearOfEra, int month, int day, CalendarSystem cale
...
         /// </summary>
         /// <remarks>
+        /// <para>
         /// <see cref="DateTimeKind.Unspecified"/> is slightly odd - it can be treated as UTC if you use <see cref="DateTime.ToLocalTime"/>
         /// or as system local time if you use <see cref="DateTime.ToUniversalTime"/>, but it's the only kind which allows
         /// you to construct a <see cref="DateTimeOffset"/> with an arbitrary offset, which makes it as close to
         /// the Noda Time non-system-specific "local" concept as exists in .NET.
+        /// </para>
+        /// <para>
+        /// <see cref="DateTime"/> does not contain any state for the calendar system of the value.
```
malcolmr, line 208: I wonder if another way to put this would be that `DateTime` uses the Gregorian calendar by definition, and so the value will be effectively converted to the Gregorian calendar first, before being represented as a `DateTime`?

malcolmr (approved):
LGTM

src/NodaTime/LocalDate.cs
jskeet: Yup. I'll make a first stab at the change and merge on green; we can always iterate on the precise wording.

PR #1666, opened 2021-11-20
Upgrade targets to include .NET 6

jskeet:
This is along the way to #1635. It's a bit of a shame that we have to go back to multiple targets, but there may well be other things we want to be able to do with .NET 6 as well, over time.

### DIFF
--- a/src/NodaTime.Test/NodaTime.Test.csproj
+++ b/src/NodaTime.Test/NodaTime.Test.csproj
@@ -11,7 +11,7 @@
     <ProjectReference Include="..\NodaTime\NodaTime.csproj" />
     <ProjectReference Include="..\NodaTime.Testing\NodaTime.Testing.csproj" />
     <PackageReference Include="NUnit" Version="3.13.2" />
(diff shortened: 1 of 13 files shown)


jskeet:
@malcolmr: Removed you as a reviewer for now, as I'll need to work out what's failing on Linux.
(A lot has changed in .NET 6 in terms of handling of time zone IDs, so that's probably what it's about.)

silkfire:
Could we remove the dependency on `System.Runtime.CompilerServices.Unsafe` for .NET 6 as it's included in the framework since .NET Core 3.1?

jskeet:
@silkfire: Yes, quite possibly. (It may automatically be removed - I'm not sure.)

The bigger problem with .NET 6 is around time zones - I need to find a solid few hours to work on that, and I'm not sure that'll be soon :(

silkfire:
Take your time, literally :)

jskeet:
This PR is now basically about fixing the time zone issues for .NET 6 - #1675 already upgraded to use the .NET 6.0 SDK.

PR #1671, opened 2022-01-23
Implement ToString() for YearMonth

### LINKED ISSUE #1670: Add default ToString() or Debugger Display for YearMonth
I've just started using NodaTime (thanks for the great library!).

I had a set of parsed YearMonth values that I was debugging, but I could easily tell what any of them were because there isn't a default `ToString` or a Debugger Display value set.

jskeet:
Fixes #1670

Unfortunately it's too late to make `yearMonth.ToString(null, null)` yield the same results as `yearMonth.ToString()`, due to compatibility reasons. But the result isn't *too* bad, really.
This will need a documentation update to document the new "G" format once it's merged.

### DIFF
--- a/src/NodaTime.Test/YearMonthTest.Misc.cs
+++ b/src/NodaTime.Test/YearMonthTest.Misc.cs
@@ -2,8 +2,12 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+using NodaTime.Test.Calendars;
+using NodaTime.Test.Text;
 using NUnit.Framework;
 using System;
+using System.Globalization;
+using System.Linq;
 
 namespace NodaTime.Test
 {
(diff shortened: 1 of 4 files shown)


src/NodaTime.Test/YearMonthTest.Misc.cs
```diff
@@ -52,5 +56,53 @@ public void PlusMonths(int year, int month, int monthsToAdd, int expectedYear, i
...
+
+        [Test]
+        [TestCaseSource(typeof(Cultures), nameof(Cultures.AllCultures))]
+        public void ToStringTest_Iso(CultureInfo culture)
+        {
+            var yearMonth = new YearMonth(2022, 1);
+            using (CultureSaver.SetCultures(culture))
+            {
+                Assert.AreEqual(yearMonth.ToString(culture.DateTimeFormat.YearMonthPattern, culture), yearMonth.ToString());
+                //Assert.AreEqual("xyz", yearMonth.ToString());
```
malcolmr, line 68: remove?

src/NodaTime.Test/YearMonthTest.Misc.cs
```diff
@@ -52,5 +56,53 @@ public void PlusMonths(int year, int month, int monthsToAdd, int expectedYear, i
...
+        [TestCaseSource(typeof(Cultures), nameof(Cultures.AllCultures))]
+        public void ToStringBclEquality(CultureInfo culture)
+        {
+            // The BCL *sometimes* use the genitive month names for year/month, even though I don't
+            // think it should. There may be more complexity here, but for the moment, we'll just skip the cultures we
+            // expect to fail. (We would definitely want to know if new cultures started failing.)
+            string[] expectedFailures = { "ast-ES", "ca-AD", "ca-ES", "ca-ES-valencia", "ca-FR", "ca-IT", "es-PE", "es-UY", "gl-ES", "oc-FR" };
+            if (expectedFailures.Contains(culture.Name))
+            {
+                return;
```
malcolmr, line 82: (Guessing there's no way to flag a skipped test to the test runner, as you can in JUnit?)

src/NodaTime/YearMonth.cs
```diff
@@ -319,11 +320,16 @@ int IComparable.CompareTo(object obj)
...
+        /// <remarks>
+        /// Unlike most <see cref="IFormattable"/> implementations, a <paramref name="patternText"/> of null with
+        /// the current thread's culture does not yield the same result as the parameterless <see cref="ToString()"/>
+        /// overload, for backward-compatibility reasons. (It uses the ISO format, which is culture-insensitive.)
+        /// </remarks>
         /// <returns>
         /// A <see cref="System.String" /> containing the value of the current instance in the specified format.
         /// </returns>
         /// <param name="patternText">The <see cref="System.String" /> specifying the pattern to use,
-        /// or null to use the default format pattern ("D").
```
malcolmr, line 326: Just so I understand: I'm assuming this was a typo before; the default was actually `g` anyway?

malcolmr (approved):
LGTM

src/NodaTime.Test/YearMonthTest.Misc.cs
jskeet: Whoops, yes.

src/NodaTime.Test/YearMonthTest.Misc.cs
jskeet: I don't *think* so (from within the test). It's feasible in xUnit, but I don't know of a way of doing so in NUnit. (We wouldn't want to mark this as skipped in the same way as some others anyway - it's more "infeasible" than "feasible but skipped".)

src/NodaTime/YearMonth.cs
jskeet: Correct. I think it was a copy/paste error.

jskeet (commented):
Thanks - will remove the commented-out line, then push and merge.

PR #1672, opened 2022-02-22
Add safeguard to prevent non webassembly friendly code

buvinghausen:
Jon,

This addition to the CSPROJ is just to prevent anyone from accidentally adding any code that might make NodaTime incompatible with Blazor Web Assembly.

Feel free to disregard it if you'd like but I'm trying to get the libraries I intend to use in Blazor at least marked with the equivalent safeguard so I don't get stung by an unintended minor change.

Buvy

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json.schemastore.org/global",
   "sdk": {
-    "version": "5.0.301",
+    "version": "6.0.200",
     "allowPrerelease": false,
     "rollForward": "latestMinor"
   }
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json.schemastore.org/global",
   "sdk": {
-    "version": "6.0.200",
+    "version": "5.0.301",
     "allowPrerelease": false,
     "rollForward": "latestMinor"
   }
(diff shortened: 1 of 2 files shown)


jskeet:
Could you point me to some documentation that would help me work out what constraints that might put on us in the future? I'm fine with it in theory (and I've used Noda Time in Blazor before) but it would be nice to have more info about what we're letting ourselves in for :)

buvinghausen:
Jon,

Sure thing so take for example the MD5 hashing algorithm.  Because it was deemed unsafe it was never ported into the WASM runtime for .NET and thus cannot be used in Blazor. 

They have decorated the [class](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.md5) with this [attribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.versioning.unsupportedosplatformattribute)

```csharp
[System.Runtime.Versioning.UnsupportedOSPlatform("browser")]
```

By putting this element into the core nodatime.csproj you just ensure you will get a compiler error if anyone were to use a class decorated as such in one of their edits.

If you create a blazorwasm hosted library you can see that the Microsoft template puts the element in the csproj on the shared library.

```
dotnet new blazorwasm -ho -o BlazorDemo
```

Regards,
Buvy

jskeet (commented):
Righto, that seems entirely reasonable. If it starts causing problems, we'll come back to you on it :)

jskeet (approved):

buvinghausen:
@jskeet by all means please do! :)

PR #1696, opened 2022-07-28
Fix rounding errors when creating Durations from doubles

### LINKED ISSUE #1695: Rounding issues when creating a Duration from a double
# Issue
Calling `Duration.FromSeconds(4.1)` creates a `Duration` with a time of `0:00:00:04.099999999` (`4099999999` nanoseconds).

https://dotnetfiddle.net/2tpYee

# Expected Behavior
A `Duration` with a time of `0:00:00:04.1` should be created.

# Information
Using NodaTime 3.1.0

Sibusten:
Duration methods that take doubles (such as `FromSeconds`) can produce values with rounding errors due to precision issues with the underlying `double` type. For example, `Duration.FromSeconds(4.1)` produces a duration with a value of `0:00:00:04.099999999`. This is confusing from a user perspective.

This proposed solution uses rounding after converting to nanoseconds, which preserves the original user intent. `FromNanoseconds` is untouched, since its comment states "Any fractional parts of the value are truncated towards zero."

Fixes #1695

### DIFF
--- a/src/NodaTime.Test/DurationTest.cs
+++ b/src/NodaTime.Test/DurationTest.cs
@@ -371,5 +371,16 @@ namespace NodaTime.Test
             Assert.AreEqual(x, Duration.Min(Duration.MaxValue, x));
             Assert.AreEqual(x, Duration.Min(x, Duration.MaxValue));
         }
+
+        [Test]
+        public void DoubleConversionHandlesRoundingErrors()
+        {
+            // This value produces rounding errors when converting to nanoseconds
+            const double expectedSeconds = 4.1;
+
+            Duration duration = Duration.FromSeconds(expectedSeconds);
+
+            Assert.AreEqual(4100000000d, duration.TotalNanoseconds);
+        }
     }
 }
--- a/src/NodaTime.Test/DurationTest.cs
+++ b/src/NodaTime.Test/DurationTest.cs
@@ -373,14 +373,40 @@ namespace NodaTime.Test
         }
 
         [Test]
-        public void DoubleConversionHandlesRoundingErrors()
+        public void FromDaysHandlesRoundingErrors()
         {
-            // This value produces rounding errors when converting to nanoseconds
-            const double expectedSeconds = 4.1;
+            Duration duration = Duration.FromDays(2.53);
+            Assert.AreEqual(218592000000000d, duration.TotalNanoseconds);
+        }
 
-            Duration duration = Duration.FromSeconds(expectedSeconds);
+        [Test]
+        public void FromHoursHandlesRoundingErrors() {
+            Duration duration = Duration.FromHours(4.1);
+            Assert.AreEqual(14760000000000d, duration.TotalNanoseconds);
+        }
 
+        [Test]
+        public void FromMinutesHandlesRoundingErrors() {
+            Duration duration = Duration.FromMinutes(4.1);
+            Assert.AreEqual(246000000000d, duration.TotalNanoseconds);
+        }
+
+        [Test]
+        public void FromSecondsHandlesRoundingErrors() {
+            Duration duration = Duration.FromSeconds(4.1);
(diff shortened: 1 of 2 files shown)


jskeet:
I'm in two minds about this. I think I would rather add more documentation to make the current behaviour crystal clear. Will think about it for a bit.

Sibusten:
If the current behavior is kept, then the `double` methods seem very unappealing. For a more significant example, `Duration.FromTicks(4.1)` produces `409` nanoseconds instead of `410`, and it's only multiplying by 100.

I could use internal logic to convert the `double` to a `long` myself, but it seems excessive when there are methods that take `double`.

Another option would be adding overloads for `decimal`, but that may make the interface more confusing.

jskeet:
It's unappealing to use for literals, yes - but that feels an odd scenario to me. As I say, I'll think about it. But probably not for a while, as I'm on holiday at the moment.

Sibusten:
No worries, enjoy your holiday!

jskeet:
Okay, I've been thinking about this more, and I'm still on the fence. Here's the analysis as I've thought about it:

**Performance**

I've just tested with our benchmarking suite, and the change appears to be lost in the noise. (I initially thought it was changing a 10ns test to 12ns, but it looks like that was just noise.)

**Compatibility**

This is my main concern. We've had the current behavior for over 5 years, so there's a real risk that any change could break existing users. I think it's a *reasonably* small risk, but it's definitely there.

**Consistency**

This is an interesting one:

- `TimeSpan.FromSeconds` does appear to round, so this would make it more consistent with that
- We don't use `Math.Round` *anywhere* else in Noda Time - we *always* truncate (e.g. `Offset.FromMilliseconds`, `Instant.ToUnixTimeMilliseconds`, `Instant.ToDateTimeUtc` etc). So this change would be inconsistent with that. 

Will haul in another couple of reviewers and see what they think.

jskeet:
@malcolmr and @mattjohnsonpint: please see my comment above in terms of the pros and cons as I see them - what have I missed? If we were to design this from scratch without any compatibility concerns, what would you expect?

Sibusten:
Adding to the note about `TimeSpan`, here's the relevant source:
https://github.com/microsoft/referencesource/blob/master/mscorlib/system/timespan.cs#L228

They actually don't use `Math.Round`, but they do custom rounding to fix the scaling problem:

```cs
double tmp = value * scale;
double millis = tmp + (value >= 0? 0.5: -0.5);
```

mattjohnsonpint:
Related (sorta), you might want to take a look at https://github.com/dotnet/runtime/pull/73198 and in particular, Tanner's comments about some edge cases.

jskeet:
@malcolmr: Ping for thoughts on this?

malcolmr:
I think it would be reasonable to treat this as a fix to a long-running bug, but I don't feel I'm that familiar with floating-point usage in practice. Certainly, if we're mostly talking about 'small' values, then it would be nice if (e.g.) `Duration.FromSeconds(4.1)` returned a duration with exactly that input value — and the `FromTicks(4.1)` example is compelling, if perhaps not that likely to occur in practice.

I do worry that if we're fixing some values, we might potentially be making this worse for some other values, but I'm not sure how to do that analysis. (I'm also aware that "being uncertain" isn't necessarily a great point to be arguing from.)

jskeet:
I keep putting off making a decision on this :) Looking at the comments on https://github.com/dotnet/runtime/pull/73198 (particularly around large numbers) does make me think a larger overhaul may be more useful, particularly for large values (where the multiplication goes above 2^53 quickly - and as we're talking nanoseconds, 2^53 nanoseconds is only 104 days. So `Duration.FromDays(200)` (for example) is almost certainly pretty bad at the moment.

I'm going to try to put together a bunch of test cases over this (long) weekend.

Sibusten:
Closing in favor of #1711

PR #1708, opened 2022-09-06
Cleaned up switch using C# 9.0 or pattern

londospark:
I was browsing the source of NodaTime as part of mentoring devs at work and saw this. While I agree that `(2, 4, 6, 10, 13) => 29` would be cleaner than what we have and would be preferred to the current solution I thought it might be nice to at least compress the five cases into a single one.


jskeet (approved):
Yup, this is exactly what I was planning to do at some point. (This syntax wasn't supported by switch expressions initially.)

I think there's another case somewhere else, but I can't remember where offhand.

PR #1711, opened 2022-09-19
Change how Duration.FromXyz(double) methods are handled

### LINKED ISSUE #1695: Rounding issues when creating a Duration from a double

jskeet:
The previous implementation had limits that were too strict: we should be able to use Duration.FromDays(MaxDays + 0.5) for example. Additionally, the new implementation rounds values that aren't precise number of nanoseconds. This fixes #1695.

The result is almost certainly slower, but more predictable. Users who are negatively impacted by this for production code can almost certainly use a more efficient approach.

This is an alternative to #1696.


jskeet:
cc @Sibusten.

src/NodaTime.Test/DurationTest.Construction.cs
```diff
@@ -262,7 +330,16 @@ public void FactoryMethods_OutOfRange()
...
             AssertOutOfRange(Duration.FromNanoseconds, bigBadDoubles);
-            
+
+            AssertOutOfRange(Duration.FromDays, new double[] { Duration.MinDays - 0.01, Duration.MaxDays + 1 });
+            AssertOutOfRange(Duration.FromHours, new double[] { Duration.MinDays * HoursPerDay - 0.01, (Duration.MaxDays + 1) * HoursPerDay });
+            AssertOutOfRange(Duration.FromMinutes, new double[] { (double) Duration.MinDays * MinutesPerDay - 0.01, (double) (Duration.MaxDays + 1) * MinutesPerDay });
+            AssertOutOfRange(Duration.FromSeconds, new double[] { (double) Duration.MinDays * SecondsPerDay - 0.01, (double) (Duration.MaxDays + 1) * SecondsPerDay });
+            // We can't actually get to 100th of a millisecond precision with a 64-bit floating point number at this scale.
+            AssertOutOfRange(Duration.FromMilliseconds, new double[] { (double) Duration.MinDays * MillisecondsPerDay - 1, (double) (Duration.MaxDays + 1) * MillisecondsPerDay });
+            // We subtract a million ticks rather than just 1 to make sure it's actually significant.
```
malcolmr, line 340: Likewise here?

src/NodaTime.Test/DurationTest.Construction.cs
```diff
@@ -165,16 +202,47 @@ public void FromSeconds_Double(double seconds, int expectedDays, long expectedNa
...
+        [TestCase(0.1, 0, 10)]
+        [TestCase(4.01, 0, 401)]
+        [TestCase(4.1, 0, 410)]
+        [TestCase(4 * TicksPerDay + 0.1, 4, NanosecondsPerTick / 10)]
+        [TestCase(4 * TicksPerDay + 0.01, 4, NanosecondsPerTick / 100)]
+        // Rounding down
+        [TestCase(86_399_999_999_9.994d, 0, NanosecondsPerDay - 1)]
+        // Rounding up across a day boundary
+        [TestCase(86_399_999_999_9.995d, 1, 0)]
+        // We need to add a lot of ticks to the start of "max day" just to get within significant digits.
```
malcolmr, line 236: This seems fluffy. If you want the next-highest double from a given value, then write that? (Probably something like `BitConverter.UInt64BitsToDouble(BitConverter.DoubleToUInt64Bits(value) + 1)`?)

src/NodaTime/Duration.cs
```diff
@@ -172,6 +172,41 @@ private Duration(long units, string paramName, long minValue, long maxValue, lon
             }
         }
 
+        // Note: unitsPerDay is always actually an integer, but we always want to perform floating point arithmetic
+        // with it.
+        private Duration(double units, string paramName, double unitsPerDay, long nanosPerUnit)
+        {
+            // Note: we don't check the original double values other than for NaN and infinity.
+            // Instead, we do the relevant conversion and then check that we're in bounds.
+            // That way we won't get invalid values due to weird rounding issues.
```
malcolmr, line 181: I think I get what you're saying here (easier and more reliable to post-verify than attempt to work out exactly when the value will exceed some computed limit), but is there some way to write this comment that doesn't make it sound like we don't know how rounding works?

src/NodaTime/Duration.cs
```diff
@@ -172,6 +172,41 @@ private Duration(long units, string paramName, long minValue, long maxValue, lon
...
+                days = (int)(units / unitsPerDay);
+                double unitOfDay = units - (unitsPerDay * days);
+                if (unitOfDay < 0)
+                {
+                    days--;
+                    unitOfDay += unitsPerDay;
+                }
+                nanoOfDay = (long) (unitOfDay * nanosPerUnit + 0.5);
+
+                // Just in case the rounding up tips us into a new day...
```
malcolmr, line 197: (Likewise, I'm assuming this is covered by some of the test cases rather than being just a theoretical possibility.)

src/NodaTime/Duration.cs
```diff
@@ -172,6 +172,41 @@ private Duration(long units, string paramName, long minValue, long maxValue, lon
...
+            // That way we won't get invalid values due to weird rounding issues.
+            if (double.IsNaN(units) || double.IsInfinity(units))
+            {
+                throw new ArgumentOutOfRangeException(paramName, "Value must be a finite, valid number.");
+            }
+            unchecked
+            {
+                days = (int)(units / unitsPerDay);
+                double unitOfDay = units - (unitsPerDay * days);
+                if (unitOfDay < 0)
```
malcolmr, line 190: Just checking: is this reachable? (My guess is that it is, but only when the space between consecutive `double` values exceeds `unitsPerDay`, but I'm still struggling to see how we get a negative value here, even then.)

malcolmr (approved):
Some questions/suggestions below, but LGTM.

src/NodaTime/Duration.cs
jskeet: Will have a go.

src/NodaTime.Test/DurationTest.Construction.cs
jskeet: Except we can't write that in an attribute value (because it's not a constant expression) and it's entirely possible that by the time we've done arithmetic, that difference would become irrelevant. I can try moving it into a separate test and see whether it actually works though...

src/NodaTime/Duration.cs
jskeet: Hmm... will check when I get a chance. (Probably not for a while.) We've got the same thing in the earlier integer version too.

src/NodaTime/Duration.cs
```diff
@@ -890,17 +925,12 @@ int IComparable.CompareTo(object? obj)
         /// Returns a <see cref="Duration"/> that represents the given number of days, assuming a 'standard' 24-hour
         /// day.
         /// </summary>
+        /// <remarks>If the value does not represent a precise number of nanoseconds, it is rounded, with half values
```
Sibusten, line 928: The modified `From...` methods mention rounding in `<remarks>`, but `FromNanoseconds` mentions truncation in `<summary>`. Should this be made consistent?

src/NodaTime.Test/DurationTest.Construction.cs
```diff
@@ -120,10 +120,20 @@ public void FromTicks(long ticks)
...
         [Test]
+        [TestCase(Duration.MinDays, Duration.MinDays, 0)]
         [TestCase(1.5, 1, NanosecondsPerDay / 2)]
         [TestCase(-0.25, -1, 3 * NanosecondsPerDay / 4)]
         [TestCase(100000.5, 100000, NanosecondsPerDay / 2)]
         [TestCase(-5000, -5000, 0)]
+        [TestCase(-0.1, -1, 9 * NanosecondsPerDay / 10)]
+        [TestCase(0.1, 0, NanosecondsPerDay / 10)]
+        [TestCase(0.01, 0, NanosecondsPerDay / 100)]
+        [TestCase(4.1, 4, NanosecondsPerDay / 10)]
```
Sibusten, line 131: I found that `Duration.FromDays(4.1)` actually worked correctly with the old logic, unlike with the other methods. That's why I had tested with `2.53` on my PR instead, as it was a value I found that did cause rounding issues. Up to you if you want to change or add that as another case

src/NodaTime/Duration.cs
Sibusten, line 197: `// Just in case the rounding up tips us into a new day...`

Would it be possible to avoid this block by doing the conversion to `nanoOfDay` (with rounding up) first and then accounting for negative values?

Sibusten (commented):
Looks good, just a few small things I noticed

src/NodaTime/Duration.cs
jskeet: Yes, will definitely do that when I get back to this.

src/NodaTime.Test/DurationTest.Construction.cs
jskeet: Ah, great - yes, will add that as another test.

src/NodaTime/Duration.cs
jskeet: Hmm. Possibly - will need to try it.

KhaledMabroukTbs (approved):

src/NodaTime/Duration.cs
KhaledMabroukTbs: Help~~~~

jskeet:
(Reported @KhaledMabroukTbs for spam comments/approvals.)

jskeet:
(I still want to do this at some point, but it'll now be in 3.2.x rather than 3.1.x...)

PR #1715, opened 2022-10-09
Update CommandLine.cs

viveknuna:
The shared resource could increase the chance of deadlocks.


jskeet:
There's no need to do this. The command line parser is only used from a single thread anyway - the locking is pointless, but I believe it's in the original code this was taken from. I don't want to make unnecessary changes.

PR #1721, opened 2022-11-04
Allow two-digit-max-year customization in patterns

jskeet:


jskeet:
It's been a while, but I think this is ready now :) Lots of copy/paste, so if you find any typos in docs etc, I'll no doubt need to fix them everywhere...

src/NodaTime/Text/InstantPattern.cs
```diff
@@ -69,16 +69,22 @@ private static class Patterns
         /// <value>The value used as a template for parsing.</value>
         public Instant TemplateValue { get; }
 
+        /// <summary>
+        /// Maximum two-digit-year in the template to treat as the current century.
+        /// </summary>
+        public int TwoDigitYearMax { get; }
```
malcolmr, line 75: optional: worth adding a `<value>` per above?

src/NodaTime/Text/InstantPattern.cs
```diff
@@ -189,6 +196,14 @@ private static InstantPattern Create(string patternText, NodaFormatInfo formatIn
         /// <param name="newTemplateValue">The template value for the new pattern, used to fill in unspecified fields.</param>
         /// <returns>A new pattern with the given template value.</returns>
         public InstantPattern WithTemplateValue(Instant newTemplateValue) =>
-            Create(PatternText, FormatInfo, newTemplateValue);
+            Create(PatternText, FormatInfo, newTemplateValue, TwoDigitYearMax);
+
+        /// <summary>
+        /// Creates a pattern like this one, but with a different <see cref="TwoDigitYearMax"/> value.
+        /// </summary>
+        /// <param name="twoDigitYearMax">The value to use for <see cref="TwoDigitYearMax"/> in the new pattern.</param>
```
malcolmr, line 204: worth adding something like "The value (0-99) to use..." to make the range clear?

Separately, where do we document the default values for this type? Probably we should document the default value for this new property?

malcolmr (approved):
LGTM

src/NodaTime/Text/InstantPattern.cs
jskeet: Done.

src/NodaTime/Text/InstantPattern.cs
jskeet: Done. Will add the default into the description of the property.

jskeet:
Thanks for the review - will merge when I've finished tweaking the docs, and when it's gone greeen.

PR #1722, opened 2022-11-04
Fix warnings around code analysis

jskeet:


jskeet:
Merging as test failures are not relevant to this.

Issue #1725, opened 2022-11-10
Need ability to support ISO 8601 patterns in LocalTime TypeConverter


jsearles:
Right now, the `TypeConverter` associated with `LocalTime` requires the precision to at least include seconds. I'm looking to support text with a pattern of `HH':'mm` which is a valid ISO 8601 pattern.

Specifically, the problem I am facing is that since the `LocalTime` is decorated with the specific `TypeConverter` I am not finding a clean way to replace it. Is there a way to configure the patterns supported by a given `TypeConverter`? Or perhaps a way to replace the registered TypeConverters that I am missing?

I believe this is related to issue https://github.com/nodatime/nodatime/issues/607

jsearles:
Actually I have no issue with the formatting. I prefer that the seconds are included. The problem I have is we are using [JSJoda](https://github.com/js-joda/js-joda) on the client side and using their implementation of LocalTime excludes the seconds if they are zero.

```js
JSJoda.LocalTime.parse("03:00:00").toJSON();
> '03:00'
```

jsearles:
Also, I have no issue with using a custom type converter - I just don't see a good way to override the built-in converter since it is applied via an attribute.

jskeet:
> Also, I have no issue with using a custom type converter - I just don't see a good way to override the built-in converter since it is applied via an attribute.

Well you've still told us *nothing* about how your context. No, there's no way of changing the default converter. There may very well be a way of specifying a different one, but we won't be able to help you find that way without knowing what kind of app you're writing.

I don't expect to change the default converter, so you should really be focusing on working out how to specify a custom one in your context.

malcolmr:
Context would be useful, yes.

The `TypeConverter` API is horrible, so the only way I can see to introduce flexibility here would be to allow statically configuring our existing converters to replace the default pattern (either via `TypeConverterSettings` or `TypeConverterBase`).

Though — as well as making this mess even more complex — it's not obvious whether that would actually help in some cases, since I'm not sure you always have the ability to run code in the context of the process doing the (de)serialisation.

jsearles:
@malcolmr - yes, those approaches would help - I agree. This is the only way that I see to make it configurable. That or remove the attribute from the types and add a helper method to wire them up for the default case. Or make this the default behavior. Probably the first approach is the right mix of configuration without breaking changes.

[...]

PR #1727, opened 2022-11-19
Provide more LocalTime ISO formats

jskeet:
These patterns allow for ISO-valid values such as "12:35" or just "12".

The same approach can then be extended to LocalDateTimePattern, although we'll need to consult ISO-8601 carefully to see whether the precision for a "date and time" can legitimately be reduced to "just a date".

### DIFF
--- a/src/NodaTime.Test/Text/LocalTimePatternTest.cs
+++ b/src/NodaTime.Test/Text/LocalTimePatternTest.cs
@@ -408,7 +408,96 @@ namespace NodaTime.Test.Text
             var pattern = LocalTimePattern.CreateWithInvariantCulture("HH").WithTemplateValue(newValue);
             Assert.AreEqual(newValue, pattern.TemplateValue);
         }
-        
+
+        [Test]
+        [TestCase("00")]
+        [TestCase("23")]
+        [TestCase("05")]
+        public void HourIso_Roundtrip(string text)
+        {
+            var result = LocalTimePattern.HourIso.Parse(text);
+            Assert.True(result.Success);
+            var time = result.Value;
+            Assert.AreEqual(0, time.Minute);
+            Assert.AreEqual(0, time.Second);
+            Assert.AreEqual(0, time.NanosecondOfSecond);
+            var formatted = LocalTimePattern.HourIso.Format(time);
+            Assert.AreEqual(text, formatted);
+        }
+
+        [Test]
+        [TestCase("-05")]
+        [TestCase("05:00")]
+        [TestCase("5")]
+        [TestCase("24")]
+        [TestCase("99")]
+        public void HourIso_Invalid(string text)
+        {
+            var result = LocalTimePattern.HourIso.Parse(text);
+            Assert.False(result.Success);
+        }
+
+        [Test]
+        [TestCase("00:31")]
+        [TestCase("23:10")]
+        public void HourMinuteIso_Roundtrip(string text)
+        {
+            var result = LocalTimePattern.HourMinuteIso.Parse(text);
+            Assert.True(result.Success);
+            var time = result.Value;
+            Assert.AreEqual(0, time.Second);
+            Assert.AreEqual(0, time.NanosecondOfSecond);
+            var formatted = LocalTimePattern.HourMinuteIso.Format(time);
+            Assert.AreEqual(text, formatted);
+        }
+
+        [Test]
+        [TestCase("-05:00")]
+        [TestCase("5:00")]
+        [TestCase("24:00")]
+        [TestCase("99:00")]
+        [TestCase("10:60")]
+        [TestCase("10:70")]
+        public void HourMinuteIso_Invalid(string text)
+        {
+            var result = LocalTimePattern.HourMinuteIso.Parse(text);
+            Assert.False(result.Success);
+        }
+
+        [Test]
+        [TestCase("03", "03:00", "03:00:00")]
+        [TestCase("12", "12:00", "12:00:00", "12:00:00.000000", "12:00:00.000000000")]
+        [TestCase("12:01", "12:01:00", "12:01:00.000000")]
+        [TestCase("12:00:01", "12:00:01.000000")]
+        [TestCase("12:00:01.123", "12:00:01.123000", "12:00:01.123000000")]
+        [TestCase("12:00:01.123456789")]
+        public void VariablePrecision_Valid(string canonical, params string[] alternatives)
+        {
+            var pattern = LocalTimePattern.VariablePrecisionIso;
+            foreach (var text in new[] { canonical }.Concat(alternatives))
+            {
+                var result = pattern.Parse(text);
+                Assert.True(result.Success);
+                var time = result.Value;
+                var formatted = pattern.Format(time);
+                Assert.AreEqual(canonical, formatted);
+            }
+        }
+
+        [Test]
+        [TestCase("24:00")]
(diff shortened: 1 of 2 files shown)


jskeet:
(This PR would make the custom binder implementation in #1725 a bit simpler.)

src/NodaTime/Text/LocalTimePattern.cs
```diff
@@ -50,6 +50,32 @@ public sealed class LocalTimePattern : IPattern<LocalTime>
...
+
+        /// <summary>
+        /// Gets an invariant local time pattern which is ISO-8601 compatible, with a precision of just hours.
+        /// This corresponds to the text pattern "HH".
+        /// </summary>
+        /// <value>An invariant local time pattern which is ISO-8601 compatible, with no sub-hour precision.</value>
+        public static LocalTimePattern HourIso => Patterns.HourIsoPatternImpl;
+
+        /// <summary>
+        /// Gets an invariant local time pattern which can parse any ISO-8601 compatible value, (just hours;
```
malcolmr, line 68: Not quite sure about "... any ISO-8601 compatible value, (just ...)." Are there ISO-8601 time formats that this can't parse? (If so, perhaps "any ... value that ..."?) If not, then I'm not sure what the parenthetical is trying to say.

malcolmr, line 68: nit: patterns can all parse and format by definition, so it's a bit weird for this to say that it "can parse", when obviously it can format as well (but I agree that parsing is likely to be much more useful that formatting for this one).

src/NodaTime.Test/Text/LocalTimePatternTest.cs
```diff
@@ -408,7 +408,99 @@ public void WithTemplateValue_PropertyFetch()
...
+            }
+        }
+
+        [Test]
+        [TestCase("24:00")]
+        [TestCase("24")]
+        [TestCase("25")]
+        [TestCase("25:61")]
+        [TestCase("12:23:45.9999999999999999")]
+        [TestCase("12:23:45.9999999999999999")]
```
malcolmr, line 494: duplicate?

src/NodaTime.Test/Text/LocalTimePatternTest.cs
```diff
@@ -408,7 +408,99 @@ public void WithTemplateValue_PropertyFetch()
...
+                Assert.AreEqual(canonical, formatted);
+            }
+        }
+
+        [Test]
+        [TestCase("24:00")]
+        [TestCase("24")]
+        [TestCase("25")]
+        [TestCase("25:61")]
+        [TestCase("12:23:45.9999999999999999")]
```
malcolmr, line 493: I assume that this is invalid because it has too many fractional digits? If so, perhaps use something closer to a valid input used above, like `"12:00:00.0000000000"` (and consider also adding a `// one too many fractional digits` comment so that it's immediately obvious why this is invalid?)

src/NodaTime.Test/Text/LocalTimePatternTest.cs
```diff
@@ -408,7 +408,99 @@ public void WithTemplateValue_PropertyFetch()
...
+            Assert.AreEqual(0, time.Minute);
+            Assert.AreEqual(0, time.Second);
+            Assert.AreEqual(0, time.NanosecondOfSecond);
+            var formatted = LocalTimePattern.HourIso.Format(time);
+            Assert.AreEqual(text, formatted);
+        }
+
+        [Test]
+        [TestCase("-05:00")]
+        [TestCase("5:00")]
```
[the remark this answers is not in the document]
jskeet: No, I didn't. Editing now...

src/NodaTime.Test/Text/LocalTimePatternTest.cs
jskeet: Deleted.

src/NodaTime.Test/Text/LocalTimePatternTest.cs
jskeet: Done.

src/NodaTime/Text/LocalTimePattern.cs
```diff
@@ -50,6 +50,32 @@ public sealed class LocalTimePattern : IPattern<LocalTime>
         /// <value>An invariant local time pattern which is ISO-8601 compatible, with no sub-second precision.</value>
         public static LocalTimePattern GeneralIso => Patterns.GeneralIsoPatternImpl;
 
+        /// <summary>
+        /// Gets an invariant local time pattern which is ISO-8601 compatible, with precision of just minutes.
+        /// This corresponds to the text pattern "HH':'mm':'ss".
```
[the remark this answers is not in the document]
jskeet: Fixed :)

src/NodaTime/Text/LocalTimePattern.cs
jskeet: It's saying that it can parse *any* ISO-value, as opposed to the other patterns which only parse ones with a specific precision. The emphasis is intended to be on the latter part. Suggestions for anything that would make it clearer would be great :)

jskeet: The parenthetical part is meant to emphasize where this is different to the other patterns. Maybe they should be in a new sentence instead of in parentheses? How about:

> Gets an invariant local time pattern which can parse any ISO-8601 compatible value, regardless of precision. Valid values include "just hours", "hours and minutes", "hours, minutes and seconds", and extended values with fractions of seconds (as far as nanoseconds).

That said, possibly "any value" is overstating it, given that we *do* require colons. Perhaps "any value in extended format" (as opposed to basic format, which could be 05, 0510, 051020).

src/NodaTime.Test/Text/LocalTimePatternTest.cs
```diff
@@ -408,7 +408,97 @@ public void WithTemplateValue_PropertyFetch()
...
+        [TestCase("25")]
+        [TestCase("25:61")]
+        [TestCase("12:23:45.0000000000")] // Too many fractional digits
+        [TestCase("05:63")]
+        [TestCase("05:00:63")]
+        public void VariablePrecision_Invalid(string text)
+        {
+            var result = LocalTimePattern.VariablePrecisionIso.Parse(text);
+            Assert.False(result.Success);
+
```
malcolmr, line 499: nit: blank line

src/NodaTime/Text/LocalTimePattern.cs
malcolmr: Yes, two sentences is much clearer, thanks!

We don't really do much with the basic format (and most of our patterns don't mention basic or extented [e.g. `GeneralIso` above]). How about "can parse any ISO-8601 compatible value (in extended format, that is, with separators), regardless of..." ?

malcolmr: That's fine, I think as-is is okay on reflection (and I can't come up with anything significantly better myself). As noted above, I think it's fine to put the emphasis on parsing for this pattern, since formatting with this is much less likely to be useful.

malcolmr (approved):
LGTM

jskeet (commented):
Thanks. Comment adjusted; will merge on green... then start on the LocalDateTime version.
(I'm *not* planning on doing the same for OffsetDateTime, ZonedDateTime etc... we can always do that later if necessary.)

PR #1731, opened 2022-12-29
Fix badge

jskeet:


README.md
```diff
@@ -1,4 +1,4 @@
-[![GitHub Actions status](https://img.shields.io/github/workflow/status/nodatime/nodatime/Build%20push)](https://github.com/nodatime/nodatime/actions?query=workflow%3A%22Build+push%22)
+[![GitHub Actions status](https://img.shields.io/github/actions/workflow/status/nodatime/nodatime/push.yaml?branch=main)](https://github.com/nodatime/nodatime/actions?query=workflow%3A%22Build+push%22)
```
malcolmr, line 1: not that it matters too much, but do you want the link destination to be updated to https://github.com/nodatime/nodatime/actions/workflows/push.yaml as well?

jskeet: Nope, I think the results of the runs is better.

malcolmr: Both links (https://github.com/nodatime/nodatime/actions?query=workflow%3A%22Build+push%22 and https://github.com/nodatime/nodatime/actions/workflows/push.yaml) show a similar list of runs; the only difference is that the former shows all workflows named "Build push" while the latter shows all workflows instantiated from that specific action. Those groups only differ because some of the historical runs for the pull-request action were also called "Build push" too.

jskeet: Ah, right. Will sort that a bit later on today.

PR #1747, opened 2023-06-03
Dependency and platform updates

jskeet:
We now don't test on .NET Core 3.1 as it's unsupported - but I've reintroduced the .NET 4.7.1 tests to run locally, as that's pretty important. (We could use Windows-based GH-actions at some point too.)


jskeet:
(I'd forgotten we're still using AppVeyor, so we should get .NET Framework tests already...)

PR #1752, opened 2023-07-13
Fix documentation for Duration.TotalMilliseconds

wdoeland:
The remark for the property TotalMilliseconds of a Duration is incorrect. It uses the remark and example from TotalSeconds, but this does not apply to TotalMilliseconds. Rewrote it to reflect the TotalMilliseconds property.

The screenshots below show that the updated example contains the correct values:
image
and 
image

Thanks for looking at this,
Wouter


jskeet (approved):

jskeet:
Thanks. As an aside, the screenshots are rather harder to read than just quoted code as text - just for reference for any future issues or PRs.

wdoeland:
Oops you're right... Thanks!

PR #1753, opened 2023-08-07
Add missing double*Duration methods

rstarkov:
There are already methods for `Duration * long`, `long * Duration`, and `Duration * double` . The one variant missing is `double * Duration`. This PR adds those, along with unit tests.

### DIFF
--- a/src/NodaTime.Test/DurationTest.Operators.cs
+++ b/src/NodaTime.Test/DurationTest.Operators.cs
@@ -184,12 +184,15 @@ namespace NodaTime.Test
             var actual = start * rightOperand;
             var expected = new Duration(expectedDays, expectedNanos);
             Assert.AreEqual(expected, actual);
+            actual = rightOperand * start;
+            Assert.AreEqual(expected, actual);
         }
-        
+
         [Test]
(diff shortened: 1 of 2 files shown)


jskeet:
Hmmm... I'm surprised we've got the `Multiply(long, Duration)` and equivalent operator at all. I definitely don't think of multiplying a number by a duration - only a duration by a number.
Still, I think it makes sense to be consistent. Will see what @malcolmr thinks.

rstarkov:
I was going to create an issue to discuss the addition of these overloads but then I found that they already exist for `long`, so decided to go straight for a PR. FWIW for commutative operators to only support one ordering and not the other is also a little weird.

As for `Multiply` I don't mind, I only added it for consistency thinking it was an oversight.

jskeet (approved):

jskeet:
Have decided to go ahead with this, as the consistency aspect is important.

malcolmr:
Sorry for taking a while to reply here, but I agree: _given_ that we already support `long * Duration`, I can't see any reason not to also support `double * Duration`.

PR #1756, opened 2023-08-12
Remove locks when reading cached fields of NodaFormatInfo

### LINKED ISSUE #1757: Lock Contention when calling ToString from multiple threads
The default `ToString` method on a lot of the NodaTime classes utilize `PatternBclSupport` which in turn tries to read cached fields from `NodaFormatInfo`. However, since the reads share the same lock, this causes lock contention whenever multiple threads try to serialize dates.

alitas:
Fixes #1757.

Since these fields are initialized only in the `EnsureInitialized` methods, they can never be overwritten once a not-null value has been read. Therefore, locking shouldn't be necessary around the reads.

### DIFF
--- a/src/NodaTime/Globalization/NodaFormatInfo.cs
+++ b/src/NodaTime/Globalization/NodaFormatInfo.cs
@@ -115,6 +115,12 @@ namespace NodaTime.Globalization
 
             lock (fieldLock)
             {
+                // Check again in case it has been initialized in another thread
+                if (longMonthNames != null)
+                {
+                    return;
+                }
+
                 // Turn month names into 1-based read-only lists
                 longMonthNames = ConvertMonthArray(DateTimeFormat.MonthNames);
                 shortMonthNames = ConvertMonthArray(DateTimeFormat.AbbreviatedMonthNames);
@@ -142,6 +148,12 @@ namespace NodaTime.Globalization
 
             lock (fieldLock)
             {
+                // Check again in case it has been initialized in another thread
+                if (longDayNames != null)
+                {
+                    return;
+                }
+                
                 longDayNames = ConvertDayArray(DateTimeFormat.DayNames);
                 shortDayNames = ConvertDayArray(DateTimeFormat.AbbreviatedDayNames);
             }
@@ -234,7 +246,7 @@ namespace NodaTime.Globalization
                 {
                     field = localConstruction;
                 }
-                return field!;
+                return field;
             }
         }
 
--- a/src/NodaTime/Globalization/NodaFormatInfo.cs
+++ b/src/NodaTime/Globalization/NodaFormatInfo.cs
@@ -108,7 +108,7 @@ namespace NodaTime.Globalization
 
         private void EnsureMonthsInitialized()
         {
-            if (longMonthNames != null)
+            if (longMonthNames != null && shortMonthNames != null && longMonthGenitiveNames != null && shortMonthGenitiveNames != null)
             {
                 return;
             }
@@ -141,7 +141,7 @@ namespace NodaTime.Globalization
 
         private void EnsureDaysInitialized()
         {
-            if (longDayNames != null)
+            if (longDayNames != null && shortDayNames != null)
             {
                 return;
             }
--- a/src/NodaTime/Globalization/NodaFormatInfo.cs
+++ b/src/NodaTime/Globalization/NodaFormatInfo.cs
@@ -108,12 +108,13 @@ namespace NodaTime.Globalization
 
         private void EnsureMonthsInitialized()
         {
+            if (longMonthNames != null)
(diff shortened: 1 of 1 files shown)


jskeet:
I'll look at this on Monday, but at first glance on a mobile I suspect it's vulnerable to race conditions with highly opitimizing JITs. I'll look at other options, including `Lazy<T>`

alitas:
> I'll look at this on Monday, but at first glance on a mobile I suspect it's vulnerable to race conditions with highly opitimizing JITs which can reorder writes. I'll look at other options, including `Lazy<T>`

I'm not sure if the race conditions apply here since the fields are already locally constructed, and the only operation happening inside the locks is assignment. Since all of the assignments are references, I wouldn't worry about torn reads either.

While `Lazy<T>` would be simpler, it does create closures for each of the value factories because of the `this` reference, and that might cause a lot of allocations here.

jskeet:
I'm not worried about torn reads - I'm worried about the fields being visibly non-null before all the assignments within the construction of the objects they refer to being visible.

Can I ask about the context in which you've seen this being a problem? Do you have some example code which you've run before and after your change, demonstrating that this is a significant issue? My understanding is that uncontended locks are really cheap - and within the lock we're doing almost *no* work unless the field actually needs to be populated. I'd expect the bottleneck for the lock to be miniscule compared with the time taken doing the rest of the operation.

alitas:
We have a pipeline that reads from multiple files and processes them in parallel, and we've noticed 17% of our CPU time was being spent in EnsureFixedFormatInitialized method. 

The locks are cheap when not contended, but they seem to add up once you're calling ToString from multiple threads at once.

jskeet:
> The locks are cheap when not contended, but they seem to add up once you're calling ToString from multiple threads at once.

How many threads do you have doing formatting? Are you able to reproduce this in a minimal application that I can play with?

One option you might want to investigate is using specific formatters which would avoid this path entirely. While I'm happy to consider changes, I don't want to do anything that makes me less confident in the robustness of the code, for the sake of a performance issue that has never been reported before and which might be avoidable. I'm not saying no - I'm just explaining my caution.

alitas:
We used `dotnet trace` for profiling, and you're probably right on the tracing overemphasizing the impact of locks. 

We did move most of our formatting to specific formatters as a workaround, and are no longer affected by this for the most part, since we don't see the `Ensure` methods anywhere in the traces. However, there are still cases where we call `Object.ToString`, which calls the default `ToString` method. Adding type checks everywhere makes the code a little too verbose at times, but it's manageable.

I believe marking the fields as `volatile` would solve the issue with the race conditions, although we can't use the helper method anymore as `ref` wouldn't work. 

Here's the snippet I used for benchmarking. No matter what the parallelism is set at (up to # of cores), per row time is fixed. (~160 ns in my tests). Removing the locks or using a specific formatter, the time is reduced to `1/Parallelism`, basically showing that everything's running in parallel.   

```csharp
private static readonly LocalTime _localTime = SystemClock.Instance.GetCurrentInstant().InUtc().TimeOfDay;
private const int _NRows = 10_000;
private const int _NParallel = 8;

public async Task RunInParallel()
{
    await Task.WhenAll(Enumerable.Range(0, _NParallel).Select(_ => FormatTime()).ToArray()).ConfigureAwait(false);
}

private static async Task<string> FormatTime()
{
    await Task.Yield();
    var result = string.Empty;
    for (var i = 0; i < _NRows; i++)
    {
        result = _localTime.ToString();
    }

    return result;
}


```

jskeet:
Just as a progress report, I've got an idea using `Interlocked.CompareExchange` (then followed by a lock) which makes a really big difference, and would be safe.
But I think I want to take a slightly bigger step back and see whether we can usefully divide NodaFormatInfo into "lazily initialized" and "eagerly initialized" - where the latter would initialize all the sub-patterns immediately, and never need to lock. More work needed.

jskeet:
It looks like `CultureInfo.CurrentCulture` returns a read-only culture, which means it should be fine to initialize all the patterns at that point and remember that we've done so in a simple Boolean field.
I'll try to get to this some time in August - although even if I do, I don't know when I'll do a new minor release. (I could *potentially* backport it for a patch release of 3.1 - I'll have to see.)

alitas:
> It looks like `CultureInfo.CurrentCulture` returns a read-only culture, which means it should be fine to initialize all the patterns at that point and remember that we've done so in a simple Boolean field. I'll try to get to this some time in August - although even if I do, I don't know when I'll do a new minor release. (I could _potentially_ backport it for a patch release of 3.1 - I'll have to see.)

Makes perfect sense to me. Thanks a lot for the quick response. I'll be closing this PR for now, we can track the issue in https://github.com/nodatime/nodatime/issues/1757.

PR #1760, opened 2023-10-05
Add unparsed value field to UnparsableValueException

alitas:
Having access to the raw value makes it easier to both log it better, and to hide the value from the original message by replacing it with `*****` on critical systems.

### DIFF
--- a/src/NodaTime/Text/UnparsableValueException.cs
+++ b/src/NodaTime/Text/UnparsableValueException.cs
@@ -21,24 +21,34 @@ namespace NodaTime.Text
         /// </summary>
         public UnparsableValueException()
         {
+            Value = string.Empty;
         }
 
         /// <summary>
         /// Creates a new UnparsableValueException with the given message.
         /// </summary>
         /// <param name="message">The failure message</param>
-        public UnparsableValueException(string message)
+        /// <param name="value">The value which could not be parsed</param>
+        public UnparsableValueException(string message, string value)
             : base(message)
         {
+            Value = value;
         }
 
         /// <summary>
         /// Creates a new UnparsableValueException with the given message and inner exception
         /// </summary>
         /// <param name="message">The failure message</param>
+        /// <param name="value">The value which could not be parsed</param>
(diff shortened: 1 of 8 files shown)


jskeet:
Will have a look at this when I get a chance, but it may well not be for a while I'm afraid.

alitas:
> Will have a look at this when I get a chance, but it may well not be for a while I'm afraid.

No worries, it's just an enhancement that might be handy in the future.

jskeet:
Okay, I've managed to have a look now, and I'm broadly in favour.

Currently, the `Value` property isn't *just* the original value - it includes the caret indicating the position at which the parsing failed. Do you think that might be confusing? Perhaps use `cursor.Value` instead of `cursor.ToString()`?
If you'd prefer to have two properties, `Value` and `ValueWithPosition` (or something with a better name!) that would be fine by me too - but equally it *might* be overkill. Thoughts welcome.

jskeet:
Sounds good. I'll look when I can - time is tight at the moment.

src/NodaTime/Text/UnparsableValueException.cs
```diff
@@ -21,24 +21,43 @@ public sealed class UnparsableValueException : FormatException
...
+        /// <param name="index">The index within the value where parsing failed</param>
         /// <param name="innerException">The inner exception</param>
-        public UnparsableValueException(string message, Exception innerException) : base(message, innerException)
+        public UnparsableValueException(string message, string value, int index, Exception innerException) : base(message, innerException)
         {
+            Value = value;
+            Index = index;
         }
+        
+        /// <summary>
```
jskeet, line 53: Post-merge, I'll update these with a little more detail.

jskeet (approved):
Thanks very much - merging now. (I don't know when we'll end up doing an minor release - it may be a while. But this'll be in at that point...)

jskeet:
I'd failed to spot this is actually a breaking change - it removes existing constructor overloads. Will fix that in a new PR.

PR #1762, opened 2023-10-10
fix: Fix the arguments passed up to exception base classes

### LINKED ISSUE #1761: Does `SkippedTimeException` use the right base constructor?
`SkippedTimeException` inherits from `ArgumentOutOfRangeException`, and calls the base constructor with one string parameter. This parameter is called `paramName`, suggesting that it should be the name of the parameter that is invalid. However what's being passed in looks like an exception message. Should it be using a different base constructor?

`AmbiguousTimeException` seems to have the same problem too.

jskeet:
Fixes #1761

This could theoretically break someone relying on ParamName actually containing the message - but since that's *clearly* a bug, I'd have hoped folks would at least have reported it at the same time, in which case we'd have fixed it earlier.

### DIFF
--- a/src/NodaTime.Test/DateTimeZoneTest.LocalConversions.cs
+++ b/src/NodaTime.Test/DateTimeZoneTest.LocalConversions.cs
@@ -111,7 +111,9 @@ namespace NodaTime.Test
             var e = Assert.Throws<SkippedTimeException>(() => mapping.Single())!;
             Assert.AreEqual(localTime, e.LocalDateTime);
             Assert.AreEqual(zone, e.Zone);
-            
+            Assert.Null(e.ParamName);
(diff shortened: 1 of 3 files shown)


malcolmr (approved):
LGTM.

It is a bit surprising that these two extend `ArgumentOutOfRangeException` but don't have a way to specify the argument involved — but this is clearly better than what we had before, and backward-compatibility concerns seem very unlikely here.

jskeet:
We *could* thread the parameter name right through the calling code, but it would be a bit messy. These are only thrown in `ZoneLocalMapping.Single`, which doesn't even *have* any parameters - but that's usually called by other code which *does* have parameters, and would need to pass them through.
Arguably it would be better as InvalidOperationException, but I'm definitely not going to change the exception type now :)

Issue #1765, opened 2023-11-13
Custom Pattern Support for text serialization


apavelm:
Currently, NodaTime can deserialize NodaTime types from string only by using the default pattern. 
Discussion was started here: https://github.com/nodatime/nodatime.serialization/issues/110
Why it is needed also mentioned there.

I'm gonna prepare a PR with a small modification that Allows to add such a customization. Feel free to modify it if necessary.

jskeet:
> Currently, NodaTime can deserialize NodaTime types from string only by using the default pattern.

When using TypeConverter - that's a *really, really* important clarification.

jskeet:
One question that could guide both the design and whether it's really a good idea: how would you configure this without Noda Time, if you were using just DateTime and DateTimeOffset? If you wanted to deserialize using type converters with custom patterns there - would it be feasible? If not, I'm not sure that doing so in Noda Time is really sensible.

jskeet:
It looks like you'd want to use `[TypeConverterAttribute]` in the model to specify which one to use for any given property. Any reason that doesn't already work in Noda Time? (We could make `TypeConverterBase` public in a new release, if that makes things simpler.)

apavelm:
> One question that could guide both the design and whether it's really a good idea: how would you configure this without Noda Time, if you were using just DateTime and DateTimeOffset? If you wanted to deserialize using type converters with custom patterns there - would it be feasible? If not, I'm not sure that doing so in Noda Time is really sensible.

I hardly can imagine the case when I will use DateTime in data contract voluntarily. It is never a good idea, I cannot say anything about DateTimeOffset. But basing on my purposes, there is no equivalent in .NET data types for "HH:mm". 
For date+time it gonna be "string" in GeneralIso format for time - string. So no way to use DateTime. Basically, I doubt that this extension will often be used, but to have it will be useful.

> It looks like you'd want to use `[TypeConverterAttribute]` in the model to specify which one to use for any given property. Any reason that doesn't already work in Noda Time? (We could make `TypeConverterBase` public in a new release, if that makes things simpler.)

Sure you can (and it will work), I was thinking about it from the beginning. But I conclude that it won't be logically correct. I mean, it would be a right way if we need to convert NodaType to any other type complex type, but we still need to have string-to-noda conversion, that requires just an additional pattern. For serialization (noda-to-string), it still use the default one (not to break the anything)

jskeet:
[...]

PR #1766, opened 2023-11-13
Added support for custom pattern on a text serialization + tests

apavelm:
Modification that allows to define custom patterns using built-in type converters for string to noda deserialization

### DIFF
--- a/src/NodaTime.Test/Text/TypeConvertersTest.cs
+++ b/src/NodaTime.Test/Text/TypeConvertersTest.cs
@@ -6,6 +6,9 @@ using NodaTime.Text;
 using NUnit.Framework;
 using System;
 using System.ComponentModel;
+using System.Globalization;
+using NodaTime.Extensions;
+using System.Text.RegularExpressions;
 
 namespace NodaTime.Test.Text
 {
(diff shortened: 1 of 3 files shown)


apavelm:
@jskeet Hi there! Xmas is coming, what about the fix/extension? :-)

jskeet:
If you could gift me another couple of hours a day, that would be great. Otherwise, I'm afraid it'll have to wait a bit.

jskeet:
I *have* had enough time to look very briefly though - and I'm not keen. This introduces more global mutable state, which is something I'm always *very* nervous of. As I said on #1765, it seems to me like the expected way to specify custom type converters is to use the `[TypeConverter]` attribute on the properties.

If you can provide examples of a global TypeConverter re-configuration scheme like you're proposing here in other significant projects (ideally in .NET itself) then that would be a lot more compelling. But I really, really don't want to introduce global mutable state just so that you can avoid specifying TypeConverter attributes.

apavelm:
If I could extend a day for a couple of hours... Unfortunately, I can't

The proposed way looks to me the most optimal considering the needs, but personally I have no objections, if you could suggest another way how to reach the goal. 
Let me remind from what all the discussions began  [110](https://github.com/nodatime/nodatime.serialization/issues/110)
It started around nodatime.serialization for AspNet.Core projects. As was identified, deserialization using custom format pattern works fine in POST requests, but doesn't work in GET requests, because for get requests it uses "string -> nodetime" converter. Currently, there are no known options how to extend it and to make it work with custom pattern formats.
If you could suggest any workable alternative, it would be good.

The suggestion to use string as a request parameter type instead of NodaTime models in GET requests and deserialize them manually - is not serious, and it also will require introducing additional static variable (jsonSerializerSettings) that will be used only for these purposes. To me, it looks clear that such a thing should be hidden inside the serialization library as an "operator" of all converters.
And again, if you prefer to go another way (discussed also), by exposing some internals (making some fields public) it will also solve the problem.

jskeet:
As I've said before, I think the most appropriate option is for you to decorate your properties with `[TypeConverter]`. You've said you don't like that, and that's fine - but I've said *I* don't like adding more mutable global state and a non-idiomatic type converter configuration mechanism.

I'd be happy to make `TypeConverterBase<T>` public (possibly renaming it in the process) to make it *easier* to add the attributes, but that's very different from introducing global state. (And even that is unlikely to get into a new release very soon - I've still got some other features/fixes to work on before I do another release, and unexpected personal and work time pressures have prevented me from getting to those yet.)

PR #1775, opened 2024-02-02
Improve performance for concurrent formatting

jskeet:
There are a couple of other minor changes in here, in separate commits. I thought it easiest to get them reviewed together. It'll be easiest to review one commit at a time.

### DIFF
--- a/src/NodaTime.Benchmarks.Custom/ConcurrentNewString.cs
+++ b/src/NodaTime.Benchmarks.Custom/ConcurrentNewString.cs
@@ -1,0 +1,38 @@
+﻿// Copyright 2024 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+using System.Threading;
+
+namespace NodaTime.Benchmarks.Custom;
+
+internal class ConcurrentNewString : ConcurrencyCommandBase
+{
+    private ConcurrentNewString(int threadCount, Duration executionTime) : base(threadCount, executionTime)
+    {
+    }
+
+    protected override void DoWork()
+    {
+        while (Interlocked.Add(ref iterations, ChunkSize) > 0)
+        {
+            for (int i = 0; i < ChunkSize; i++)
+            {
+                _ = new string('x', 8);
+            }
+        }
+    }
+
+    internal static void Execute(string[] args)
+    {
+        if (args.Length != 2)
+        {
+            throw new ConfigurationException("Command arguments: <processing thread count> <execution time in seconds>");
+        }
+
+        int threadCount = int.Parse(args[0]);
+        int executionTimeSeconds = int.Parse(args[1]);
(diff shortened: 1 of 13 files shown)


src/NodaTime/Globalization/NodaFormatInfo.cs
```diff
@@ -398,9 +448,11 @@ internal static NodaFormatInfo GetFormatInfo(CultureInfo cultureInfo)
                 return InvariantInfo;
             }
             // Never cache (or consult the cache) for non-read-only cultures.
+            // We don't initialize eagerly in this case, to preserve previous behavior -
```
malcolmr, line 451: I think this is the right decision (to not pre-initialize), but I'm not completely sure about the rationale here: does having a non-read-only `CultureInfo` input make it more likely that the output `NodaFormatInfo` will only be used briefly?

src/NodaTime.Benchmarks.Custom/ConcurrentNewString.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+    }
+
+    protected override void DoWork()
+    {
+        while (Interlocked.Add(ref iterations, ChunkSize) > 0)
+        {
+            for (int i = 0; i < ChunkSize; i++)
+            {
+                _ = new string('x', 8);
```
malcolmr, line 21: Also, I'm somewhat surprised these benchmarks don't end up getting optimized out of existence (either at the IL level or when JITted).

malcolmr (approved):
LGTM. You didn't share the results of these benchmarks, but I assume the main commit ends up making them better :)

jskeet (commented):
I'm publishing a new version of this PR with comments addressed, and was going to supply benchmarks - but now I can't reproduce the behaviour I was seeing before. My memory says I was seeing results (for 10 threads) of about 5 million iterations per second before the change, and 30 million after. It's now 60 million both before and after, which leaves me very confused. I may wait until I'm actually home (currently in a hotel) before merging this. I wonder whether I've broken something...

(I've added a new benchmark to the first commit which *should* allow me to compare results without having to go back to an earlier state.)

jskeet:
I'm now doubly confused about the benchmarks, as I don't think formatting with this LocalTime pattern would do anything with the affected properties. Looking further...

jskeet:
Doh! Yes, concurrent-patternformat is fast; I'd previously been observing concurrent-tostring, which *does* exhibit the desired change in performance.

Before: 5-6 million iterations per second
After: 25-30 million iterations per second

The reason `ToString()` is slower is that it goes through `NodaFormatInfo.LocalTimePatternParser` on each call, and *that* needed the lock.

I'm now *not* going to try the "do all the NodaFormatInfo initialization inside a redundant lock" testing, as none of our benchmarks (or expected behaviour for eager initialization) create lots and lots of NodaFormatInfo instances - so I'm not too worried about shaving off nanoseconds there.

jskeet:
(And I've removed the *attempt* at "show performance with an uncached instance" as in order to go through that path, we have to remove another level of caching anyway.)

PR #1780, opened 2024-02-05
Add Duration.ToInt128Nanoseconds and Duration.FromNanoseconds(Int128)

### LINKED ISSUE #1777: Use Int128 in .NET 8 target
(e.g. for Duration.ToInt128Nanoseconds etc)

We could *experiment* with a conditional implementation, using Int128 in Duration when it's available - but I suspect it'll be too fiddly.

jskeet:
These are only available for the .NET 8 build (although in source code they're .NET 7+ as that's when Int128 was introduced; this is just in case we introduce a .NET 7 build).

Fixes #1777.

Benchmarks show that the Int128 operations are about the same as BigInteger in the worst cases, and about twice as fast in better cases.

### DIFF
--- a/src/NodaTime/Duration.cs
+++ b/src/NodaTime/Duration.cs
@@ -71,6 +71,10 @@ namespace NodaTime
         internal const int MinDays = ~MaxDays;
         internal static readonly BigInteger MinNanoseconds = (BigInteger) MinDays * NanosecondsPerDay;
         internal static readonly BigInteger MaxNanoseconds = (MaxDays + BigInteger.One) * NanosecondsPerDay - BigInteger.One;
+#if NET7_0_OR_GREATER
+        private static readonly Int128 MinInt128Nanoseconds = (Int128) MinNanoseconds;
+        private static readonly Int128 MaxInt128Nanoseconds = (Int128) MaxNanoseconds;
+#endif
         internal static readonly decimal MinDecimalNanoseconds = (decimal) MinNanoseconds;
         internal static readonly decimal MaxDecimalNanoseconds = (decimal) MaxNanoseconds;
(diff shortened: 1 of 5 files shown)


jskeet:
(This will probably mark the end of my small spate of Noda Time PRs, btw.)

src/NodaTime/Duration.cs
```diff
@@ -1103,6 +1108,28 @@ public static Duration FromNanoseconds(BigInteger nanoseconds)
...
+        /// <summary>
+        /// Converts a number of nanoseconds expressed as an <see cref="Int128"/> into a duration.
+        /// </summary>
+        /// <param name="nanoseconds">The number of nanoseconds to represent.</param>
+        /// <returns>A duration with the given number of nanoseconds.</returns>
+        public static Duration FromNanoseconds(Int128 nanoseconds)
+        {
+            if (nanoseconds < MinInt128Nanoseconds || nanoseconds > MaxInt128Nanoseconds)
+            {
+                throw new ArgumentOutOfRangeException(nameof(nanoseconds), $"Value should be in range [{MinNanoseconds}-{MaxNanoseconds}]");
```
malcolmr, line 1121: We probably should have a trivial test that values outside the range throw the expected exception/message, perhaps once for all of the `FromNanoseconds()` overloads?

malcolmr (approved):
LGTM

src/NodaTime/Duration.cs
jskeet: Yup, sounds good to me.

jskeet: It turns out we have FactoryMethods_OutOfRange already - I just need to add to that.

PR #1800, opened 2024-05-20
Tidy up unreferenced test case source collections in `LocalTimePatternTest.cs`

chrisimcevoy:
Hello, my name is Chris. Towards the end of last year, I contacted Jon to more or less ask for his blessing to work on a [Python port of this library](https://github.com/chrisimcevoy/pyoda-time).

In that context, I noticed while reading `LocalTimePatternTest.cs` that there are a couple of historic `TestCaseSource` collections which have fallen into disuse, i.e. there are no remaining references to them.

- [`DefaultPatternData`](https://github.com/nodatime/nodatime/blob/8c41fe377b8ff8121b2942e749c2a3b81e797e20/src/NodaTime.Test/Text/LocalTimePatternTest.cs#L153-L164)
- [`TemplateValueData`](https://github.com/nodatime/nodatime/blob/8c41fe377b8ff8121b2942e749c2a3b81e797e20/src/NodaTime.Test/Text/LocalTimePatternTest.cs#L166-L183)

The last reference to `DefaultPatternData` was [removed in 2012](https://github.com/nodatime/nodatime/commit/74bf0896f65428351a46cd1bb0b9c7bcace72004#diff-28c4459b9497dc18d0893c4b399d0430bbd0eb97bb9b7e242e80a21038e81ee0L338-L351). I am not sure how beneficial it is to retain these test cases, but I have moved them into the `FormatAndParseData` collection for now.

The last reference to `TemplateValueData` was [removed in 2011](https://github.com/nodatime/nodatime/commit/d0013f3d67924fedcffc90bb701b9a7bc3808bc9#diff-1709b0cc502a466a88d57bdd64bd64fe5a2508871801be052c3116fd83ae4de7L39-L49), although the individual test cases live on, mostly unchanged, [here](https://github.com/nodatime/nodatime/blob/e913a53d2d20636bdc56dd2a8786976731792b0c/src/NodaTime.Test/Text/LocalTimePatternTest.cs#L269-L297). So it seems clear that the unreferenced collection can just be removed...?

### DIFF
--- a/src/NodaTime.Test/Text/LocalTimePatternTest.cs
+++ b/src/NodaTime.Test/Text/LocalTimePatternTest.cs
@@ -150,19 +150,6 @@ namespace NodaTime.Test.Text
             new Data(14, 15, 16) { Culture = Cultures.Invariant, Text = "14:15", Pattern = "t" },
         };
 
-        internal static Data[] DefaultPatternData = {
-            // Invariant culture uses HH:mm:ss for the "long" pattern
-            new Data(5, 0, 0, 0) { Text = "05:00:00" },
-            new Data(5, 12, 0, 0) { Text = "05:12:00" },
-            new Data(5, 12, 34, 0) { Text = "05:12:34" },
-
-            // US uses hh:mm:ss tt for the "long" pattern
-            new Data(17, 0, 0, 0) { Culture = Cultures.EnUs, Text = "5:00:00 PM" },
-            new Data(5, 0, 0, 0) { Culture = Cultures.EnUs, Text = "5:00:00 AM" },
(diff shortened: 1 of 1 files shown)


jskeet:
Thanks, I'll have a look when I get a chance. I seem to remember that there's some oddity here, where something may not *appear* to be used, but actually is - I'll see what we get in terms of test results.

jskeet:
Right, it genuinely looks like the tests aren't used... I'll take a closer look at your PR to check that we're still actually covering the same ground we expected to.

jskeet (approved):
Yes, after looking for a bit, I think this is fine.

PR #1806, opened 2024-06-17
Remove nonsensical aspect of capture for invariant methods.

### LINKED ISSUE #1804: Potential mistake in the docs for `LocalDatePattern.CreateWithInvariantCulture`
In https://nodatime.org/3.1.x/api/NodaTime.Text.LocalDatePattern.html#NodaTime_Text_LocalDatePattern_CreateWithInvariantCulture_System_String_ it says:

> Note that the current culture is captured at the time this method is called - it is not captured at the point of parsing or formatting values.

I think this is a copy-paste mistake, because this method shouldn't capture the current culture - only the invariant culture. In fact I don't know whether the invariant culture even needs capturing, as the whole point is that it's invariant?!

I suspect this issue might apply to other pattern types, but I haven't checked!

jskeet:
Fixes #1804.

Equivalent to #1805, but on the 3.1.x branch.

### DIFF
--- a/src/NodaTime/Text/AnnualDatePattern.cs
+++ b/src/NodaTime/Text/AnnualDatePattern.cs
@@ -173,9 +173,7 @@ namespace NodaTime.Text
         /// Creates a pattern for the given pattern text in the invariant culture.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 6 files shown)


jskeet:
Merging despite the test failure

PR #1808, opened 2024-06-21
feat: add Period.MinValue and Period.MaxValue

### LINKED ISSUE #1807: Consider adding Period.{Max,Min}Value
Similar to other NodaTime types.

The specific trigger here is PostgreSQL introducing `-infinity` and `infinity` to its `interval` type, which corresponds to Period (see https://github.com/npgsql/npgsql/issues/5696#issuecomment-2179805408).

azan-n:
closes #1807

### DIFF
--- a/src/NodaTime.Test/PeriodTest.cs
+++ b/src/NodaTime.Test/PeriodTest.cs
@@ -629,7 +629,7 @@ namespace NodaTime.Test
         [Test]
         public void ToString_NegativeInifinity() 
         {
-            Assert.AreEqual("-infinity", Period.PositiveInfinity.ToString());
+            Assert.AreEqual("-infinity", Period.NegativeInfinity.ToString());
         }
 
         [Test]
--- a/src/NodaTime.Test/PeriodTest.cs
+++ b/src/NodaTime.Test/PeriodTest.cs
@@ -940,6 +940,39 @@ namespace NodaTime.Test
             Assert.AreEqual(0, februaryComparer.Compare(month, month));
         }
 
+        public static TestCaseData[] PeriodMaxAndMinValues
+        {
+            get
+            {
+                return [
+                    new TestCaseData(Period.MaxValue, int.MaxValue, long.MaxValue).SetName("All members should be MaxValue"), 
+                    new TestCaseData(Period.MinValue, int.MinValue, long.MinValue).SetName("All members should be MinValue")
+                ];
+            }
+        }
+
+        /// <summary>
+        /// Ensure that Period.MaxValue and Period.MinValue contain the max/min value assignable to each property/member.
+        /// </summary>
+        [Test]
+        [TestCaseSource(nameof(PeriodMaxAndMinValues))]
+        public void Period_MaxAndMinValues_AllMembers(Period period, int expectedIntValue, long expectedLongValue)
+        {
+            // Integer values
+            Assert.AreEqual(expectedIntValue, period.Years);
+            Assert.AreEqual(expectedIntValue, period.Months);
+            Assert.AreEqual(expectedIntValue, period.Weeks);
+            Assert.AreEqual(expectedIntValue, period.Days);
+
+            // Long values
+            Assert.AreEqual(expectedLongValue, period.Hours);
+            Assert.AreEqual(expectedLongValue, period.Minutes);
+            Assert.AreEqual(expectedLongValue, period.Seconds);
+            Assert.AreEqual(expectedLongValue, period.Milliseconds);
+            Assert.AreEqual(expectedLongValue, period.Ticks);
+            Assert.AreEqual(expectedLongValue, period.Nanoseconds);
+        }
+
         [Test]
         [TestCaseSource(nameof(AllPeriodUnits))]
         public void Between_ExtremeValues(PeriodUnits units)
--- a/src/NodaTime.Test/PeriodTest.cs
+++ b/src/NodaTime.Test/PeriodTest.cs
@@ -388,6 +388,15 @@ namespace NodaTime.Test
         }
 
         [Test]
+        public void Addition_MaxAndMinValue()
+        {
+            Period p1 = Period.MaxValue;
+            Period p2 = Period.MinValue;
+            Period sum = p1 + p2;
+            Assert.AreEqual(new Period(-1, -1, -1, -1, -1, -1, -1, -1, -1, -1), sum);
+        }
+
+        [Test]
         public void Addition_With_IdenticalPeriodTypes()
         {
             Period p1 = Period.FromHours(3);
--- a/src/NodaTime.Test/PeriodTest.cs
+++ b/src/NodaTime.Test/PeriodTest.cs
@@ -3,6 +3,7 @@
 // as found in the LICENSE.txt file.
 
 using System;
+using System.Reflection;
 using NodaTime.Calendars;
 using NodaTime.Text;
 using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


src/NodaTime/Period.cs
```diff
@@ -831,7 +844,7 @@ public Duration ToDuration()
         /// changes made to the builder are not reflected in this period.
         /// </summary>
         /// <returns>A builder with the same values and units as this period.</returns>
-        [Pure] [TestExemption(TestExemptionCategory.ConversionName)] public PeriodBuilder ToBuilder() => new PeriodBuilder(this);
+        [Pure][TestExemption(TestExemptionCategory.ConversionName)] public PeriodBuilder ToBuilder() => new PeriodBuilder(this);
```
azan-n, line 847: @jskeet apologies for the formatting change. I might not have set up my editor correctly. Is there an .editorconfig or similar that I should work with?

src/NodaTime/Text/PeriodPattern.cs
```diff
@@ -193,6 +193,19 @@ public StringBuilder AppendFormat(Period value, StringBuilder builder)
                     builder.Append("P0D");
                     return builder;
                 }
+
+                if (value.Equals(Period.PositiveInfinity))
+                {
+                    builder.Append("infinity");
```
azan-n, line 199: @roji @jskeet 

Both Parse and ToString() will need to adapt to this change.  

It seems like there might be information in the [latest standard]((https://www.iso.org/obp/ui/en/#iso:std:iso:8601:-1:ed-1:v1:en) about how to do this but I can only access parts of it and will have to pay to access the whole document, so much for having a standard that you can only implement if you spend enough. I digress.


Do you have suggestions on how to represent (and expect) the string format?

roji: Not sure the infinity values should have any specific ToString/Parse behavior; these are not special values (in the PostgreSQL sense) - they're exactly the same as a Period manually constructed with MaxValue components...

azan-n: I see what you mean. It makes more sense and does not pollute NodaTime with Npgsql domain specifics.

src/NodaTime/Period.cs
jskeet: No, I don't have an editorconfig for this repo at the moment I'm afraid. Can you just revert this change and the one below manually?
(And you can revert the change to PeriodPattern.cs completely.)

src/NodaTime/Period.cs
```diff
@@ -69,6 +69,19 @@ public sealed class Period : IEquatable<Period?>
         /// <value>A period containing only zero-valued properties.</value>
         public static Period Zero { get; } = new Period(0, 0, 0, 0);
 
+        /// <summary>
+        /// A period containing the maximum value for all properties.
+        /// </summary>
+        /// <value>A period containing the maximum value for all properties.</value>
+        public static Period MaxValue { get; } = new Period(int.MaxValue, int.MaxValue, int.MaxValue, int.MaxValue, long.MaxValue, long.MaxValue, long.MaxValue, long.MaxValue, long.MaxValue, long.MaxValue);
+
```
jskeet, line 77: Nit: remove one of these blank lines.

jskeet (commented):
I think this is going to be much simpler than trying to work out infinity semantics.

We *could* add a test that checks that the value of each property is the min/max value of that type - that would mean we don't need to worry about having accidentally supplied `int.MinValue` for a `long` property, for example. I could write that in a separate PR though.

src/NodaTime/Period.cs
azan-n: Done. I'm generally careful about these things before I push a commit but I guess it just slipped my mind.

azan-n:
> We _could_ add a test that checks that the value of each property is the min/max value of that type - that would mean we don't need to worry about having accidentally supplied `int.MinValue` for a `long` property, for example. I could write that in a separate PR though.

On it.

src/NodaTime/Period.cs
azan-n: Done.

src/NodaTime.Test/PeriodTest.cs
```diff
@@ -931,6 +940,39 @@ public void Comparer_NonDurationablePeriods()
...
+                ];
+            }
+        }
+
+        /// <summary>
+        /// Ensure that Period.MaxValue and Period.MinValue contain the max/min value assignable to each property/member.
+        /// </summary>
+        [Test]
+        [TestCaseSource(nameof(PeriodMaxAndMinValues))]
+        public void Period_MaxAndMinValues_AllMembers(Period period, int expectedIntValue, long expectedLongValue)
```
azan-n, line 959: I was thinking I'll work with reflection here but it just feels wrong to me.
This is the first time I'm working with NUnit so if there's changes needed here, please let me know.

src/NodaTime.Test/PeriodTest.cs
```diff
@@ -931,6 +940,39 @@ public void Comparer_NonDurationablePeriods()
             Assert.AreEqual(0, februaryComparer.Compare(month, month));
         }
 
+        public static TestCaseData[] PeriodMaxAndMinValues
```
jskeet, line 943: This can be simplified with an expression-bodied member, and it can be private:

```csharp
private static TestCaseData[] PeriodMaxAndMinValues =>
[
    new TestCaseData(Period.MaxValue, int.MaxValue, long.MaxValue).SetName("All members should be MaxValue"), 
    new TestCaseData(Period.MinValue, int.MinValue, long.MinValue).SetName("All members should be MinValue")
];
```

src/NodaTime.Test/PeriodTest.cs
jskeet: The reason for using reflection would be to avoid baking in assumptions about the type. If `Days` were actually a `long` rather than an `int`, the test would still pass. So I was planning on using all the properties, and testing the actual value against the expected value for that property's type.

I honestly don't know whether that's overkill - having explained my reasoning a bit more, what do you think?

The code would be something like this (which may not compile quite - I'm just typing it directly):

```csharp
foreach (var property in typeof(Period).GetProperties())
{
    var actualValue= property.GetValue(period);
    var expectedValue = actualValue is int ? expectedIntValue
        : actualValue is long ? expectedLongValue
        : throw new InvalidOperationException($"Property {property.Name} has unexpected type");
    Assert.AreEqual(expectedValue, actualValue);
}
```

jskeet (approved):

jskeet (commented):
(Sorry, hadn't meant to approve yet, as I'm interested in the testing discussion. But I'm happy that we'll get there.)

src/NodaTime.Test/PeriodTest.cs
azan-n: > The reason for using reflection would be to avoid baking in assumptions about the type

This is exactly what I was thinking and would try to do with every language. C# is a bit new to me, so I steer clear of reflection runtime sorcery.

Let me know if the new test works better.

src/NodaTime.Test/PeriodTest.cs
```diff
@@ -931,6 +941,37 @@ public void Comparer_NonDurationablePeriods()
...
+        /// </summary>
+        [Test]
+        [TestCaseSource(nameof(PeriodMaxAndMinValues))]
+        public void Period_MaxAndMinValues_AllMembers(Period period, int expectedIntValue, long expectedLongValue)
+        {
+            foreach (var property in typeof(Period).GetProperties(BindingFlags.Instance | BindingFlags.Public))
+            {
+                var actualValue = property.GetValue(period);
+                
+                // HasTimeComponent is a public method that will get included in this loop.
```
jskeet, line 961: Nit: method => property

jskeet (approved):
Looks good :) One more nit, then I'll squash and merge.

src/NodaTime.Test/PeriodTest.cs
azan-n: Done.

jskeet:
Great, thanks! Will squash and merge on green.

PR #1810, opened 2024-08-11
Improve perf of LocalDate.WithCalendar if calendar is unchanged

brantburnett:
If the LocalDate.WithCalendar is called with the same calendar the cost of the DaysSinceEpoch math can be avoided.

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3880/23H2/2023Update/SunValley3) 12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores .NET SDK 8.0.303
  [Host]     : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2

| Method                          | Mean     | Error     | StdDev    | Ratio |
|-------------------------------- |---------:|----------:|----------:|------:|
| WithCalendar_Unchanged_New      | 1.045 ns | 0.0150 ns | 0.0133 ns |  0.15 |
| WithCalendar_Unchanged_Previous | 7.043 ns | 0.0941 ns | 0.0880 ns |  1.00 |

### DIFF
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -627,7 +627,9 @@ namespace NodaTime
         public LocalDate WithCalendar(CalendarSystem calendar)
         {
             Preconditions.CheckNotNull(calendar, nameof(calendar));
-            return new LocalDate(DaysSinceEpoch, calendar);
+            return ReferenceEquals(calendar, Calendar)
+                ? this
+                : new LocalDate(DaysSinceEpoch, calendar);
         }
 
         /// <summary>
--- a/src/NodaTime/LocalDate.cs
+++ b/src/NodaTime/LocalDate.cs
@@ -627,7 +627,7 @@ namespace NodaTime
         public LocalDate WithCalendar(CalendarSystem calendar)
         {
             Preconditions.CheckNotNull(calendar, nameof(calendar));
-            return ReferenceEquals(calendar, Calendar)
+            return calendar.Ordinal == CalendarOrdinal
                 ? this
                 : new LocalDate(DaysSinceEpoch, calendar);
         }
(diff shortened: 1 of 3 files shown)


src/NodaTime/LocalDate.cs
```diff
@@ -627,7 +627,9 @@ public ZonedDateTime AtStartOfDayInZone(DateTimeZone zone)
         public LocalDate WithCalendar(CalendarSystem calendar)
         {
             Preconditions.CheckNotNull(calendar, nameof(calendar));
-            return new LocalDate(DaysSinceEpoch, calendar);
+            return ReferenceEquals(calendar, Calendar)
```
jskeet, line 630: Rather than use the Calendar property, which has to look up the CalendarSystem via its Ordinal, I'd just compare the ordinals: `calendar.Ordinal == CalendarOrdinal`.

jskeet (commented):
Love the change in general - just a suggestion for a more efficient approach to getting comparing the calendars.

src/NodaTime/LocalDate.cs
brantburnett: Ahh, i missed that subtlety, I just assumed the calendar was stored as a pointer. That's for catching that, this makes the performance improvement even more significant.

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3880/23H2/2023Update/SunValley3)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.303
  [Host]     : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2


| Method                          | Mean      | Error     | StdDev    | Ratio |
|-------------------------------- |----------:|----------:|----------:|------:|
| WithCalendar_Unchanged_New      | 0.0836 ns | 0.0096 ns | 0.0085 ns |  0.01 |
| WithCalendar_Unchanged_Previous | 7.1869 ns | 0.1479 ns | 0.1383 ns |  1.00 |

PR #1815, opened 2024-09-28
Tweaks to generic operators

jskeet:
(Just trivial tests and moving a member really.)


jskeet:
(Not sure why coverage reports aren't showing up, but never mind...)

Issue #1751, opened 2023-07-06
Plans for TimeProvider in .NET 8


jennings:
[.NET 8 introduces a new abstraction of time called TimeProvider](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#time-abstraction), and I'm wondering if NodaTime already has plans to integrate with it.

# What is TimeProvider?

TimeProvider has the following API:

```csharp
public abstract class TimeProvider
{
    // Clock methods
    public virtual DateTimeOffset GetUtcNow();
    public virtual TimeZoneInfo LocalTimeZone { get; }

    // Stopwatch methods
    public virtual long TimestampFrequency { get; }
    public virtual long GetTimestamp();

    // Timers
    public virtual ITimer CreateTimer(TimerCallback callback, object? state, TimeSpan dueTime, TimeSpan period);

    // Other non-virtual methods (implemented via the virtual methods)
    public DateTimeOffset GetLocalNow();
    public TimeSpan GetElapsedTime();
    public TimeSpan GetElapsedTime(long startingTimestamp);
}
```

This abstraction seems to combine the following functionality:

- Get the current time, optionally in a preconfigured time zone (similar to `ZonedClock`)
- Get a timestamp suitable for measurement (similar to `Stopwatch`)
- Create timers

Since this has clock-y behavior, it seems like NodaTime should probably do _something_ about this type.

# What to do about it

n.b.: I only learned of `TimeProvider` the other day, so this suggestion isn't fully thought through.

NodaTime _could_ write an implementation of `TimeProvider` backed by an `IClock`, but:

- NodaTime doesn't have anything to say about the timestamp methods
- The `LocalTimeZone` property is required but I don't think NodaTime can convert between `DateTimeZone` and `TimeZoneInfo` (right?)
- `IClock` doesn't have an API to power `CreateTimer`, so either `IClock` would need to be expanded or NodaTime would need separate implementations backed by a SystemClock and a FakeClock.

As I write this issue, I'm convincing myself that NodaTime should instead make it convenient to use a `TimeProvider` _in place_ of an `IClock` since `TimeProvider` _is_ a clock. It would be nice if NodaTime provided am extension method to use a `TimeProvider` in place of an `IClock`:

```csharp
public static class NodaTimeProviderExtensions
{
    public static Instant GetCurrentInstant(this TimeProvider timeProvider)
        => Instant.FromDateTimeOffset(timeProvider.GetUtcNow());
}
```

[...]

PR #1816, opened 2024-09-28
Use TimeProvider in .NET 8+ to implement IClock

jskeet:
First part of #1751.

Using static methods/properties to return IClock instances feels like it's more discoverable than "knowing" about SystemClock etc (or having a separate Clocks class). It's not supported in .NET Framework though, hence the conditionality. Given that TimeProvider is only available in .NET 8 though, there's no need to expose the TimeProviderClock class at all, hence the private implementation.

The ZonedClock factory methods are just for convenience.

We *could* do all of this from a TimeProviderExtensions class instead:

- `IClock ToClock(this TimeProvider provider)`
- `ZonedClock ToZonedClock(this TimeProvider provider)`
- `ZonedClock ToZonedClock(this TimeProvider provider, CalendarSystem calendar)`

Or we could do that as well... thoughts welcome.

(No tests written yet until we've got the production API discussed.)

### DIFF
--- a/src/NodaTime/IClock.cs
+++ b/src/NodaTime/IClock.cs
@@ -2,6 +2,11 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+#if NET8_0_OR_GREATER
+using NodaTime.Annotations;
+using NodaTime.Extensions;
+using NodaTime.Utility;
+#endif
 using System;
 
 namespace NodaTime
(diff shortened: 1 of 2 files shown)


jskeet:
Note: this currently fails to build because it's adding interface members to `IClock`. I *suspect* this is a bug in ValidatePackage, as I'd expect it to be valid to add *static* members to an interface (just not instance members). I'll follow up on that.

jskeet:
(It's forbidden in https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#members, but that doesn't talk about static members. I'll need to work out who to follow up with on this.)

patrikwlund:
#### API
I like this approach because it provides a single source of truth for both API's, in a very canonical manner.

In tests the developer may simply work with a single `FakeTimeProvider`, which integrates with _everything_. In business code you may choose to work with either `IClock` _or_ `TimeProvider`, whichever one suits the moment. They have the same source of truth.

#### Instant from TimeProvider
The issue (#1751) suggests allowing you to get an `Instant` directly from a `TimeProvider` (`timeProvider.GetCurrentInstant()`). I think that's fair enough, though it doesn't seem all that necessary. Using the current conversion methods is simple already, and a shortcut can be added in a separate PR.

#### Extension methods
I would love to see extension methods being included. Discoverability of extension methods has gotten quite good, and it's a natural and fluent way to go from A to B with `a.ToB()` (or `a.AsB()` depending on your taste and the semantics).

Simply:
`timeProvider.AsClock()`

Or in a dependency injection example:
```
services.AddSingleton<IClock>(s => s.GetRequiredService<TimeProvider>().AsClock());
```

jskeet:
Thanks @patrikwlund. The "adding members to the IClock interface" part will be blocked on a decision about whether this is actually a breaking change or not, so it might make sense to add the extension methods immediately. (And I can easily add the extension method for GetCurrentInstant at the same time, to be honest.) I'd still quite like to add the static members in IClock afterwards though.

Unless there are dissenting opinions, I'll do that over the weekend (possibly as a separate PR, keeping this one for *just* the additional static members in IClock as I've referred to it in an issue).

jennings:
I don't think I would put clock factories as static methods on IClock, but that's just my preference. I think `SystemClock.Instance` and a hypothetical `TimeProviderClock.Create(timeProvider)` would be perfectly discoverable. But there's probably nothing wrong with putting them on IClock too.

I still think an built-in extension method `Instant GetCurrentInstant(this TimeProvider timeProvider)` would be welcome because:

- Most future code is going to need a `TimeProvider`, since that's the built-in abstraction.
- I expect to gradually replace `IClock` with `TimeProvider` in my applications, since it can do everything `IClock` can plus more, and it's silly to depend on two clock abstractions.
- The extension method would make getting an Instant feel like a "first-class operation". Having to call `.GetUtcNow().ToInstant()` every time feels like it would encourage me to work directly with `DateTimeOffset` and avoid NodaTime altogether.

That said, this is all just my own preference. There's no fundamental problem here, and I can always write my own extension method if it is that important to me.

Thanks for this, and all the work you've put into NodaTime! I cannot begin to describe how many date/time bugs it has prevented for us.

jskeet:
> I don't think I would put clock factories as static methods on IClock, but that's just my preference. I think `SystemClock.Instance` and a hypothetical `TimeProviderClock.Create(timeProvider)` would be perfectly discoverable. But there's probably nothing wrong with putting them on IClock too.

I'll probably *start* with just extension methods. The benefit of that over `TimeProviderClock.Create` is that then I don't need to expose the `TimeProviderClock` class at all (yet).

I'm hoping to create a second PR tomorrow - and yes, I'm convinced about TimeProvider.GetCurrentInstant() as an extension method :)

PR #1820, opened 2024-10-05
Add TimeProvider extension methods

jskeet:
This adds:

- ToClock
- ToZonedClock (with and without calendar system)
- GetCurrentInstant

First steps towards #1751 (and may be all we ever do).

/cc @jennings @patrikwlund

### DIFF
--- a/src/NodaTime.Test/Annotations/ReflectionTest.cs
+++ b/src/NodaTime.Test/Annotations/ReflectionTest.cs
@@ -3,6 +3,7 @@
 // as found in the LICENSE.txt file.
 using NodaTime.Annotations;
 using NUnit.Framework;
+using System;
 using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
@@ -10,7 +11,7 @@ using System.Text.RegularExpressions;
 namespace NodaTime.Test.Annotations
 {
     /// <summary>
-    /// Not necessary annotations, but a general purpose "anything using reflection" test.
+    /// Not necessarily annotations, but a general purpose "anything using reflection" test.
     /// </summary>
     public class ReflectionTest
     {
@@ -26,9 +27,17 @@ namespace NodaTime.Test.Annotations
                 .SelectMany(m => m.DeclaredMethods)
                 .Where(m => m.IsPublic)
                 .Where(m => conversionName.IsMatch(m.Name))
-                .Where(m => !m.Name.StartsWith($"To{m.ReturnType.Name}"));
+                .Where(NameIsInconsistentWithReturnType);
 
             TestHelper.AssertNoFailures(badMethods, m => $"{m.DeclaringType?.Name}.{m.Name}", TestExemptionCategory.ConversionName);
+
+            bool NameIsInconsistentWithReturnType(MethodInfo method)
+            {
+                var type = method.ReturnType;
+                // When converting to an interface, the I prefix stutters somewhat - so we have "ToClock" instead of "ToIClock" for example.
+                var expectedTypeName = type.IsInterface ? type.Name.Substring(1) : type.Name;
+                return !method.Name.StartsWith($"To{expectedTypeName}", StringComparison.Ordinal);
+            }
         }
     }
 }
--- a/src/NodaTime.Test/Extensions/TimeProviderExtensionsTest.cs
+++ b/src/NodaTime.Test/Extensions/TimeProviderExtensionsTest.cs
@@ -1,0 +1,86 @@
+﻿// Copyright 2024 The Noda Time Authors. All rights reserved.
+// Use of this source code is governed by the Apache License 2.0,
+// as found in the LICENSE.txt file.
+
+#if NET8_0_OR_GREATER
+using Microsoft.Extensions.Time.Testing;
+using NodaTime.Extensions;
+using NUnit.Framework;
+using System;
+
+namespace NodaTime.Test.Extensions;
+
+public class TimeProviderOnlyExtensionsTest
+{
+    private static readonly DateTimeOffset SampleStartDto = new DateTimeOffset(2024, 10, 5, 11, 29, 0, TimeSpan.FromHours(1));
+    private static readonly Instant SampleStartInstant = Instant.FromUtc(2024, 10, 5, 10, 29);
+
+    [Test]
+    public void ToClock()
+    {
+        var provider = new FakeTimeProvider(SampleStartDto);
+        var clock = provider.ToClock();
+        Assert.AreEqual(SampleStartInstant, clock.GetCurrentInstant());
+
+        provider.Advance(TimeSpan.FromSeconds(1));
+        Assert.AreEqual(SampleStartInstant + Duration.FromSeconds(1), clock.GetCurrentInstant());
+    }
+
+    [Test]
+    public void GetCurrentInstant()
+    {
+        var provider = new FakeTimeProvider(SampleStartDto);
+        Assert.AreEqual(SampleStartInstant, provider.GetCurrentInstant());
+    }
+
+    [Test]
+    public void ToZonedClock_Iso()
+    {
+        var provider = new FakeTimeProvider(SampleStartDto);
+        // In .NET 8, IANA time zones are always available.
+        provider.SetLocalTimeZone(TimeZoneInfo.FindSystemTimeZoneById("Europe/London"));
+
+        var zonedClock = provider.ToZonedClock();
+        Assert.AreEqual(new LocalDateTime(2024, 10, 5, 11, 29), zonedClock.GetCurrentLocalDateTime());
+        Assert.AreEqual(Offset.FromHours(1), zonedClock.GetCurrentOffsetDateTime().Offset);
+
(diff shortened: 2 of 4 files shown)


src/NodaTime.Test/Extensions/TimeProviderExtensionsTest.cs
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright 2022 The Noda Time Authors. All rights reserved.
```
malcolmr, line 1: 2024 :)

src/NodaTime/Extensions/TimeProviderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+    private class TimeProviderClock : IClock
+    {
+        private readonly TimeProvider timeProvider;
+
+        internal TimeProviderClock(TimeProvider timeProvider)
+        {
+            this.timeProvider = Preconditions.CheckNotNull(timeProvider, nameof(timeProvider));
+        }
+
+        public Instant GetCurrentInstant() => timeProvider.GetUtcNow().ToInstant();
```
malcolmr, line 80: Since you're already relying on an extension method here, any reason not to use the one above to write

```csharp
public Instant GetCurrentInstant() => timeProvider.GetCurrentInstant();
```

?

src/NodaTime/Extensions/TimeProviderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// Returns the current instant in time provided by <paramref name="timeProvider"/>,
+    /// as an <see cref="Instant"/>.
+    /// </summary>
+    /// <remarks>
+    /// This is equivalent to calling <see cref="TimeProvider.GetUtcNow"/>, and then calling
+    /// <see cref="DateTimeOffsetExtensions.ToInstant(DateTimeOffset)"/> on the result.
+    /// </remarks>
+    /// <param name="timeProvider">The time provider to obtain the current instant in time from. Must not be null.</param>
+    /// <returns>The current instant in time as returned by the given time provider, expressed as an <see cref="Instant"/>.</returns>
+    public static Instant GetCurrentInstant(this TimeProvider timeProvider) =>
```
malcolmr, line 40: I did wonder whether this should try to match the naming that exists on `TimeProvider` (and e.g. be something like `GetInstant()`) or whether using the name we already have on `IClock` is better.

(I think what you have is _probably_ best?)

src/NodaTime/Extensions/TimeProviderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+        timeProvider.GetUtcNow().ToInstant();
+
+    /// <summary>
+    /// Returns a <see cref="ZonedClock"/> which obtains the current instant in time and the local time zone
+    /// from <paramref name="timeProvider"/>, and uses the ISO calendar.
+    /// </summary>
+    /// <remarks>
+    /// The local time zone is obtained from the time provider at the time of this call. If the time provider
```
malcolmr, line 48: perhaps `s/is obtained/is captured/`?

(and same below, if you decide to change it.)

src/NodaTime/Extensions/TimeProviderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// changes which time zone is reported by <see cref="TimeProvider.LocalTimeZone"/> after this call has returned,
+    /// that change will not be reflected in the returned <see cref="ZonedClock"/>.
+    /// </remarks>
+    /// <param name="timeProvider">The time provider to delegate to. Must not be null.</param>
+    /// <returns>A <see cref="ZonedClock"/> backed by the given time provider.</returns>
+    public static ZonedClock ToZonedClock(this TimeProvider timeProvider) => ToZonedClock(timeProvider, CalendarSystem.Iso);
```
malcolmr, line 54: I'm a little hazy on reasons, but I assume it would be a bad idea to consider having just the method below with an optional `calendar` parameter?

(If nothing else, it's probably a bad idea for consistency, since we don't use optional parameters elsewhere in the API, but I'm wondering whether there's a more-concrete reason.)

src/NodaTime.Test/Annotations/ReflectionTest.cs
```diff
@@ -26,9 +27,17 @@ public void ConversionNamesMatchTargetType()
                 .SelectMany(m => m.DeclaredMethods)
                 .Where(m => m.IsPublic)
                 .Where(m => conversionName.IsMatch(m.Name))
-                .Where(m => !m.Name.StartsWith($"To{m.ReturnType.Name}"));
+                .Where(NameIsInvalid);
 
             TestHelper.AssertNoFailures(badMethods, m => $"{m.DeclaringType?.Name}.{m.Name}", TestExemptionCategory.ConversionName);
+
+            bool NameIsInvalid(MethodInfo method)
```
malcolmr, line 34: Maybe `IsUnexpected` or `IsInconsistentWithReturnType`? If this test fails, the method name is definitely valid, it's just not what we wanted :)

src/NodaTime.Test/Extensions/TimeProviderExtensionsTest.cs
```diff
@@ -0,0 +1,72 @@
...
+        // 2024-10-05 ISO is 2024-09-22 Julian
+        Assert.AreEqual(new LocalDateTime(2024, 9, 22, 11, 29, calendar), zonedClock.GetCurrentLocalDateTime());
+        Assert.AreEqual(Offset.FromHours(1), zonedClock.GetCurrentOffsetDateTime().Offset);
+
+        // The time zone is unaffected by which calendar system we use.
+        provider.Advance(TimeSpan.FromDays(22));
+        Assert.AreEqual(new LocalDateTime(2024, 10, 14, 10, 29, calendar), zonedClock.GetCurrentLocalDateTime());
+        Assert.AreEqual(Offset.Zero, zonedClock.GetCurrentOffsetDateTime().Offset);
+    }
+}
```
malcolmr, line 71: Maybe add a test that demonstrates that `toZonedClock` captures the zone from `TimeProvider` at the point of construction? While admittedly not the most important property, it is something that we document.

src/NodaTime.Test/Extensions/TimeProviderExtensionsTest.cs
```diff
@@ -0,0 +1,72 @@
...
+        var provider = new FakeTimeProvider(SampleStartDto);
+        // In .NET 8, IANA time zones are always available.
+        provider.SetLocalTimeZone(TimeZoneInfo.FindSystemTimeZoneById("Europe/London"));
+
+        var zonedClock = provider.ToZonedClock(CalendarSystem.Julian);
+        // 2024-10-05 ISO is 2024-09-22 Julian
+        Assert.AreEqual(new LocalDateTime(2024, 9, 22, 11, 29, calendar), zonedClock.GetCurrentLocalDateTime());
+        Assert.AreEqual(Offset.FromHours(1), zonedClock.GetCurrentOffsetDateTime().Offset);
+
+        // The time zone is unaffected by which calendar system we use.
```
malcolmr, line 66: I'm not sure what this comment is trying to say? (Just that the time zone's offset will work the same way as in the ISO calendar?)

malcolmr (approved):
LGTM, optional suggestions below.

src/NodaTime.Test/Extensions/TimeProviderExtensionsTest.cs
jskeet: Good call - done.

src/NodaTime/Extensions/TimeProviderExtensions.cs
jskeet: My aim was for existing users of IClock to be able to effectively treat a TimeProvider *as* an IClock. (So yes, worth considering, but I'm opting to keep it as-is.)

src/NodaTime/Extensions/TimeProviderExtensions.cs
jskeet: Yes, good plan. Done.

src/NodaTime/Extensions/TimeProviderExtensions.cs
jskeet: At this point, it's just consistency really.
(There are reasons why we couldn't start with a parameterless method and then change it to just have an optional parameter, but they wouldn't apply here. There's *some* guidance around "consider languages which don't support optional parameters" but I think it's fair to ignore that these days.)

src/NodaTime/Extensions/TimeProviderExtensions.cs
jskeet: Yes, a very good reason: I hadn't considered `TimeProvider.GetCurrentInstant()` at all when I originally implemented this :)

Changed now.

src/NodaTime.Test/Annotations/ReflectionTest.cs
jskeet: The latter is nice. Done.

jskeet (commented):
Thanks - will force push the small updates, then merge on green - then do a release!

PR #1836, opened 2024-11-16
Add NanosecondsBetween method to Period

### LINKED ISSUE #1835: Non-allocative way of having TotalSeconds between two LocalTime inputs
Hello!
For some reason the project I work on have called something 

```
public double GetTotalSeconds(LocalTime time1, LocalTime time2)
{
    return (time1 - time2).ToDuration().TotalSeconds;
}
```
which results in 80B allocation per call.
We call this on 3 different time fields and this has to be repeated for a large number of times which results in large amount of allocation.

Im not sure how useful it is for others but if possible I would like to see if we can have something like https://github.com/nodatime/nodatime/issues/1450 and implement this without having a Period object created. ie. Get the same thing done by Period.Between return directly:
`(double)(EndTime.NanosecondOfDay - StartTime.NanosecondOfDay) / 1_000_000_000` 

This should yield a benchmark like below (shortrun job):
| Method                 | Mean       | Error      | StdDev    | Median     | Gen0   | Allocated |
|----------------------- |-----------:|-----------:|----------:|-----------:|-------:|----------:|
| TotalSecondsLocalTimePeriod  | 23.1932 ns | 12.0196 ns | 0.6588 ns | 22.8489 ns | 0.0064 |      80 B |
| TotalSecondsLocalTime |  0.0349 ns |  0.8089 ns | 0.0443 ns |  0.0200 ns |      - |         - |

If you are happy on this proposal, I can help to create a PR for it (first time :)).
Thanks in advance.
(...and question I have in mind..:
- I saw there is an uncheck on Period.Between()'s implementation - but curious under what case that would be required since it is always doing a subtraction?
- Why not implement #1450 as extension on LocalTime itself but in the Period class?)

nickkfwong:
Fixes #1835

### DIFF
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -427,9 +427,9 @@ namespace NodaTime
         /// Returns the difference between start and end time in nanoseconds.
         /// </summary>
         /// <remarks>
-        /// If <paramref name="end"/> is before <paramref name="start" />, returned value will be negative.
+        /// If <paramref name="end"/> is before <paramref name="start" />, the returned value will be negative.
         /// </remarks>
-        /// <param name="start"> Start time</param>
+        /// <param name="start">Start time</param>
         /// <param name="end">End time</param>
         /// <returns>The nanoseconds between the given times</returns>
         public static long NanosecondsBetween(LocalTime start, LocalTime end)
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -424,14 +424,14 @@ namespace NodaTime
         }
 
         /// <summary>
-        /// Returns the difference between start and end time in nanoseconds.
+        /// Returns the number of nanoseconds between two <see cref="LocalTime"/> objects.
         /// </summary>
         /// <remarks>
         /// If <paramref name="end"/> is before <paramref name="start" />, the returned value will be negative.
         /// </remarks>
         /// <param name="start">Start time</param>
         /// <param name="end">End time</param>
-        /// <returns>The nanoseconds between the given times</returns>
+        /// <returns>The number of nanoseconds between the given times.</returns>
         public static long NanosecondsBetween(LocalTime start, LocalTime end)
         {
             return unchecked(end.NanosecondOfDay - start.NanosecondOfDay);
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -424,7 +424,7 @@ namespace NodaTime
         }
 
         /// <summary>
-        /// Returns the number of nanoseconds between two <see cref="LocalTime"/> objects.
+        /// Returns the number of nanoseconds between two <see cref="LocalTime"/> values.
         /// </summary>
         /// <remarks>
         /// If <paramref name="end"/> is before <paramref name="start" />, the returned value will be negative.
--- a/src/NodaTime/Period.cs
+++ b/src/NodaTime/Period.cs
@@ -424,6 +424,20 @@ namespace NodaTime
         }
 
         /// <summary>
+        /// Returns the difference between start and end time in nanoseconds.
+        /// </summary>
+        /// <remarks>
+        /// If <paramref name="end"/> is before <paramref name="start" />, returned value will be negative.
+        /// </remarks>
+        /// <param name="start"> Start time</param>
+        /// <param name="end">End time</param>
+        /// <returns>The nanoseconds between the given times</returns>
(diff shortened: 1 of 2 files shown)


src/NodaTime/Period.cs
```diff
@@ -423,6 +423,20 @@ public static int DaysBetween(LocalDate start, LocalDate end)
             return InternalDaysBetween(start, end);
         }
 
+        /// <summary>
+        /// Returns the difference between start and end time in nanoseconds.
+        /// </summary>
+        /// <remarks>
+        /// If <paramref name="end"/> is before <paramref name="start" />, returned value will be negative.
```
jskeet, line 430:
```suggestion
        /// If <paramref name="end"/> is before <paramref name="start" />, the returned value will be negative.
```

src/NodaTime/Period.cs
```diff
@@ -423,6 +423,20 @@ public static int DaysBetween(LocalDate start, LocalDate end)
...
+        /// </remarks>
+        /// <param name="start"> Start time</param>
```
jskeet, line 432:
```suggestion
        /// <param name="start">Start time</param>
```

jskeet (commented):
Just a couple of tiny nits, then we can get this merged.

nickkfwong:
> Just a couple of tiny nits, then we can get this merged.

Updated the suggestions, great thanks for review!

src/NodaTime/Period.cs
```diff
@@ -423,6 +423,20 @@ public static int DaysBetween(LocalDate start, LocalDate end)
             return InternalDaysBetween(start, end);
         }
 
+        /// <summary>
+        /// Returns the difference between start and end time in nanoseconds.
```
malcolmr, line 427: Perhaps `Returns the number of nanoseconds between two <see cref="LocalTime"/> objects.` ? (following `DaysBetween()` above)

nickkfwong: Actually I have been thinking if there could be better description as well. 

A few quote from other methods:
1. Between(LocalTime start, LocalTime end)
`Returns the exact difference between two times.`

2. Between(LocalTime start, LocalTime end, PeriodUnits units)
`Returns the period between a start and an end time, using only the given units.`

3. Between(LocalDate start, LocalDate end)
`Returns the exact difference between two dates.`

4. DaysBetween(LocalDate start, LocalDate end)
`Returns the number of days between two <see cref="LocalDate"/> objects.`

Style of 4 looks absolutely fine to me, if you all agree I can submit a change for it.

jskeet: Yup, the style of 4 looks right to me.

src/NodaTime/Period.cs
```diff
@@ -424,14 +424,14 @@ public static int DaysBetween(LocalDate start, LocalDate end)
         }
 
         /// <summary>
-        /// Returns the difference between start and end time in nanoseconds.
+        /// Returns the number of nanoseconds between two <see cref="LocalTime"/> objects.
```
jskeet, line 427:
```suggestion
        /// Returns the number of nanoseconds between two <see cref="LocalTime"/> values.
```

(Sorry, didn't clock the word "objects" before, which isn't appropriate when LocalTime is a struct...)

jskeet (approved):
Looks good, thanks. Will squash and merge when the tests have all run.

src/NodaTime/Period.cs
nickkfwong: Updated both summary and return section to align with the DaysBetween()'s description

jskeet:
Just to set expectations: when this has been merged, it may still be quite a long time before we next do a minor release. I'd suggest adding an method in your own code (which can have the exact same implementation) until that point.

nickkfwong:
> Just to set expectations: when this has been merged, it may still be quite a long time before we next do a minor release. I'd suggest adding an method in your own code (which can have the exact same implementation) until that point.

No problem, that is expected! Thanks @jskeet & @malcolmr.

Btw, do you want to update the description on DaysBetween as well on this change? (objects vs values)

jskeet:
> Btw, do you want to update the description on DaysBetween as well on this change? (objects vs values)

I'll do that myself in a separate PR, I think - probably updating the style of the method bodies to use expression-bodied members at the same time :)

nickkfwong:
> > Btw, do you want to update the description on DaysBetween as well on this change? (objects vs values)
> 
> I'll do that myself in a separate PR, I think - probably updating the style of the method bodies to use expression-bodied members at the same time :)

sure :D

PR #1846, opened 2025-03-27
Migrate from VSTest to Microsoft.Testing.Platform

Youssef1313:


jskeet:
I'm not going to merge a pretty major PR which has only a single line of description - no explanation, no pros and cons, risk evaluation etc. I suggest you open an issue first to discuss whether we *should* do this, any concerns etc. (That's generally my recommendation for any change that's not just a minor fix.)

Youssef1313:
@jskeet Sorry I was doing a quick experiment yesterday and had to move to do something else before adding info in PR (I should have opened it as a draft, verry sorry). I opened an issue now #1847. Converting to draft for now.

jskeet:
As noted in the issue: I'll look at this and test it locally with the way I debug things when I get a chance, but it may be open for quite some time.

PR #1876, opened 2026-05-16
Remove SharpCompress in NodaTime tools

0xced:
Use the built-in System.IO.Compression.GZipStream + System.Formats.Tar.TarReader (available since .NET 7) instead.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,7 +13,6 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.3.0" />
     <PackageVersion Include="XmlSchemaClassGenerator-beta" Version="3.0.1270" />
-    <PackageVersion Include="SharpCompress" Version="0.48.0" />
     <PackageVersion Include="SauceControl.InheritDoc" Version="2.0.2" />
     <!-- Packages for tools -->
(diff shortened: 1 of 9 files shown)


jskeet:
Will try to look at this later. I have memories of having to add BZip support to SharpCompress for the sake of NodaTime, but that may be obsolete by now...

jskeet:
Not BZip, LZip. Maybe it'll just work - but if not, we can probably still remove SharpCompress from *some* projects, and add a test to demonstrate what would fail in this PR.

jskeet:
It looks like tzdata is still distributed with tar.gz - it's only the complete distribution (code + data) which is in .tar.lz. (I still think that's an odd choice, but hey...)

Merging.

jskeet:
(And thanks, by the way. The fewer dependencies we can get away with, the better...)

0xced:
I saw the comment about _the lzip file_ but did not understand it. Now it makes sense!

So I may have broken extracting the data from a **.tar.lz** file such as https://data.iana.org/time-zones/releases/tzdb-2026b.tar.lz (Now it throws System.IO.InvalidDataException: Unable to parse number). But I guess it doesn't really matter as long as the data only distribution is used, which is currently what the script does.

https://github.com/nodatime/nodatime/blob/769ee7b1a2895b693245263f7824b1048edb768d/build/tzdbupdate/update-main.sh#L29

PR #1877, opened 2026-05-17
Use the CLDR version for the WindowsZones Version property

### LINKED ISSUE #1559: Mapping revision not correctly injected in tzdb2020a.nzd
The revision number seem to have been badly injected in this file
This is visible in the VersionId property of an IDateTimeZoneProvider loaded from that file which is "TZDB: 2020a (mapping: $Revision$)" while a number would be expected instead of that $Revision$
This is also visible on that page: [https://nodatime.org/TimeZones?version=2020a#](https://nodatime.org/TimeZones?version=2020a#)

It' not a big deal but it's just visually annoying for me.

0xced:
For many years, the Version property has been `$Revision$` which is not useful at all. There's no resolution in sight: https://unicode-org.atlassian.net/browse/CLDR-13931

This changes the Version property to the CLDR version of the windowsZones-*.xml file so it's at least somewhat useful.

Fixes #1559

### DIFF
--- a/src/NodaTime.TzdbCompiler.Test/Tzdb/CldrWindowsZonesParserTest.cs
+++ b/src/NodaTime.TzdbCompiler.Test/Tzdb/CldrWindowsZonesParserTest.cs
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the Apache License 2.0,
 // as found in the LICENSE.txt file.
 
+using System.IO;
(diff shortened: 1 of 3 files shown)


jskeet (approved):
Thanks, I definitely agree that it seems unlikely that we'll get back to useful revision numbers in the CLDR data, at least any time soon...

PR #1882, opened 2026-06-30
Accept unambiguous, case-insensitive prefixes in TZDB month/day parsing

### LINKED ISSUE #1813: Make TZDB parsing more robust
Currently we parse full and abbreviated month names - but in theory it should be "anything unambiguous".

jskeet:
Fixes #1813


jskeet:
Note: this is the first use of AI when creating code for Noda Time.

Non-trivial prompts required:

- Highlight case-insensitivity, and use of this for day-of-week names as well (neither are mentioned in issue 1813)
- Originally left both LongMonths and ShortMonths, not spotting the commonality

PR #1892, opened 2026-09-15
Reject hour 24 in OffsetTime patterns

youdie006:
`OffsetTimePattern` accepts hour 24 and hands back a value outside `LocalTime`'s own range:

```
OffsetTimePattern.Rfc3339.Parse("24:30:15Z")   ->  success
  Hour = 24, NanosecondOfDay = 88215000000000   (NanosecondsPerDay = 86400000000000)
```

`OffsetTimePatternParser.cs:28` registers `'H'` with `maxValue` 24. That bound is sound for `LocalDateTime`/`OffsetDateTime`/`ZonedDateTime`, because `LocalDateTimeParseBucket.CombineBuckets` (`:136-163`) zeroes the hour, rolls the date forward, and rejects anything that is not exactly midnight. `OffsetTime` has no date, so `OffsetTimeParseBucket.CalculateValue` (`:93-102`) never runs that step and 24 reaches `LocalTime.FromHourMinuteSecondNanosecondTrusted` unchanged. `LocalTimePatternParser.cs:32` - the other time-only type - already bounds the same field at 23, and `LocalTimePattern` rejects the identical text.

`24:30:15` is not end-of-day either, so no reading of ISO-8601 rescues it, and RFC 3339 does not allow hour 24 at all.

**This changes observable behaviour**, so it is your call rather than mine: text that parsed now fails. My argument is that the value it used to return was already outside the type's documented invariant, so this repairs a defect rather than removing a feature - but if you would rather hold it for a major version, or prefer keeping the bound at 24 and returning `InvalidHour24` from the bucket instead (the shape the date-bearing types use), say so and I will redo it or close this.

<details>
<summary>Verification</summary>

`dotnet test -c Release -f net10.0 src/NodaTime.Test` on `mcr.microsoft.com/dotnet/sdk:10.0` (SDK pinned by `global.json`):

| | Failed | Passed | Total |
|---|---|---|---|
| this branch | 0 | 22702 | 22708 |
| parser reverted, tests kept | 3 | 22699 | 22708 |

The three are only the new rows. Nothing else moves.

Reverting the bound to 24 fails the three `Hour24IsRejected` rows; narrowing it to 22 instead fails the `23:59:59+00` row, which is in the diff for exactly that reason.

`OffsetTimePatternTest` had no hour-24 row at all, which is why this was not caught - `InvalidHour24` appears ten times in the suite, all in `LocalDateTimePatternTest` and `OffsetDateTimePatternTest`.

I did not run the rest of `build/ci.sh` - the Demo, TzdbCompiler, AOT publish and pack steps - only the test project that covers the change.

</details>

Prepared with AI assistance (Claude); I reproduced the out-of-range value and verified the red-green and both mutation directions myself.


jskeet:
I'll have a look when I can - it may be a week or two though. Will need to carefully check what we do for each other type and why.

jskeet:
Initial reading of this:

- We definitely shouldn't be allowing 24:30:15
- We *might* want to think about allowing 24:00:00, but I'm not convinced (it should be allowed in OffsetDateTime, but that would already move the date by a day and change the time to 00:00:00)

Assuming that's all true, I'm happy to consider this a bug to be fixed in a minor version. (I might end up looking at how we're violating the invariants too... I wonder if we've got other similar bugs elsewhere.)

src/NodaTime.Test/Text/OffsetTimePatternTest.cs
```diff
@@ -130,6 +133,20 @@ public void WithTemplateValue()
             Assert.AreEqual(Offset.FromHours(2), parsed.Offset);
         }
 
+        [Test]
+        [TestCase("24:00:00+00")]
```
jskeet: Assuming we do take this, I think this should probably just go in ParseFailureData. (That will validate that we're using the right message, too.)

jskeet:
Okay, I've had a look, and I think this is correct and that we don't have the same problem elsewhere. The other places that use an inclusive upper bound of 24 for `PatternFields.Hours24` are using `LocalDateTimeParseBucket.CombineBuckets` which checks that the time is midnight and increments the date (when the value is 24).

Tests all look good now - will merge on green, thanks.

I'm not planning on backporting this to the 3.3.x line, but could do if it's important to you to do so. Otherwise, I'll wait until 3.4.0 is released - I don't have a timing plan for that at the moment. Basically, let me know if this is actively causing you problems or whether it's just an annoyance.

